PR #1, opened 2020-10-26
Add trophy case

pabloest:


dlukeomalley:
Please open against develop :)

pabloest:
Np, changed (I think).

PR #4, opened 2020-10-27
make the left navbar thinner

ievans:
the CSS changes seem a bit brittle so I don't know if we actually want this, but it gives us a lot more space for content when I have my browser taking up 50% of my screen, which a good number of people might when reading our docs.

before then after:
Screen Shot 2020-10-27 at 9 02 21 AM
Screen Shot 2020-10-27 at 9 02 11 AM

### DIFF
--- a/docs/extra.css
+++ b/docs/extra.css
@@ -280,3 +280,21 @@ div[role="navigation"] hr {
 .lang-container {
   margin-top: 50px;
 }
+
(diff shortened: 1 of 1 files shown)


pabloest:
Thanks for this. LGTM, IMO getting the extra room for content is worth the slight squeezing of the nav.

PR #13, opened 2020-10-28
Dlukeomalley/overview tweaks

dlukeomalley:


docs/cli-usage.md
```diff
@@ -0,0 +1,17 @@
+# CLI Usage
```
dlukeomalley, line 1: @pabloest Note that I don't include this in the navigation tree

pabloest: I am fine with excluding it but how do I get to this doc in general?

docs/cli-usage.md
```diff
@@ -0,0 +1,17 @@
...
+
+The following usage documentation is for the [Semgrep CLI](https://github.com/returntocorp/semgrep).
+
+[TOC]
```
pabloest, line 5: TOC is probably not needed here? This doc is super short.

docs/cli-usage.md
dlukeomalley: It linked from the Semgrep README and appears in search results

docs/cli-usage.md
dlukeomalley: I've gotten into the practice of adding them. TOC provides anchor links too, should that be nice. Happy to remove though.

pabloest: Ah yes, probably useful for the anchors alone. 👍

PR #35, opened 2020-11-06
Update integrations.md

### LINKED ISSUE #34: Discoverability of inline PR comment feature in docs
Right now all the information around inline pr comments is just in a comment in an example yaml file. It would be easier to find if it had its own header.

(After making these changes, verify that searching the docs for inline pr comments will return a relevant result, which it doesn't do currently)

sabihb:
Fixes #34

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -67,9 +67,6 @@ jobs:
       - uses: returntocorp/semgrep-action@v1
 
         # Set GITHUB_TOKEN to leave inline comments on your pull requests.
-        # Note that this feature is experimental; please reach out to support@r2c.dev
-        # to report any issues.
-
         #env:
         #  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
@@ -87,6 +84,11 @@ jobs:
       #    sarif_file: semgrep.sarif
       #  if: always()
 ```
+
+#### Inline PR Comments (beta)
+Set `GITHUB_TOKEN` (commented out in the code above) to leave inline comments on your pull requests.
+Comments are left when Semgrep CI finds a result that blocks CI.
+Note that this feature is experimental; please reach out to support@r2c.dev to report any issues.
 
 </p>
 </details>
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -86,7 +86,11 @@ jobs:
 ```
 
 #### Inline PR Comments (beta)
-Set `GITHUB_TOKEN` (commented out in the code above) to leave inline comments on your pull requests.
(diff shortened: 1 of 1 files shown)


docs/integrations.md
```diff
@@ -88,6 +85,11 @@ jobs:
       #  if: always()
 ```
 
+#### Inline PR Comments (beta)
+Set `GITHUB_TOKEN` (commented out in the code above) to leave inline comments on your pull requests.
```
dlukeomalley, line 89: @sabihb Could we add a `!!! info` notice that this feature is currently for GitHub users only?

When you say "commented out in the code above", what code above has it commented out? Do all of the providers have that code?

docs/integrations.md
```diff
@@ -88,6 +85,11 @@ jobs:
...
+Comments are left when Semgrep CI finds a result that blocks CI.
```
dlukeomalley, line 90: This surfaces a good product question! Should inline comments only be triggered by "block CI" results? I'd love to chat this week, I've gotten feature requests for notify-only WITH inline pr comments :)

dlukeomalley (changes requested):

docs/integrations.md
sabihb: That is a very good question - I wonder which is more intuitive for the user. I think the reasoning for implementing it firing on "Block CI" is that an inline PR comment does "block" a PR.

Also, what exactly is "block CI" for in general? Findings that should cause CI to fail? (In which case, are rules/rulesets specified to not block CI effectively disabled?)

docs/integrations.md
sabihb: 👍  Will add the notice! I'll clarify the sentence about the commented out code. I was attempting to refer to the yaml file example above this paragraph (where the two lines to add inline pr comments are commented out).

docs/integrations.md
dlukeomalley: Does an inline PR comment have to block a PR or can it be informational (i.e. done outside of a PR-review)? Rules / rulesets specified not to block CI are effectively informational.

sabihb: Just looking around, it seems you can "Add single comment", which doesn't require a review. The comments right now do start a review (which is not something I remember setting, so it must just be the endpoint we're using), but I'm sure there exists an API endpoint that allows you to just leave an inline comment.

docs/integrations.md
sabihb: Updated! Let me know what you think.

dlukeomalley (approved):

PR #29, opened 2020-10-29
Clarify that Circle and Buildkite are unsupported

nbrahms:

PR #39, opened 2020-11-10
add generic CI instructions

chmccreery:
Screen Shot 2020-11-09 at 4 37 51 PM

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -11,6 +11,20 @@ The following instructions use [Semgrep CI](https://github.com/returntocorp/semg
 !!! danger
     `SEMGREP_APP_TOKEN` is a secret value: DO NOT HARDCODE IT AND LEAK CREDENTIALS. Use your CI provider's secret or environment variable management feature to store it. 
 
+<details><summary>Generic CI Provider</summary>
+<p>
+
+```sh
+# Set additional environment variables
+$ SEMGREP_JOB_URL=https://example.com/me/myjob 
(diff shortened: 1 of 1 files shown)


dlukeomalley:
@chmccreery Mind taking a look at https://github.com/returntocorp/semgrep-docs/pull/29 and seeing if there is anything to pull over from that? I believe we could make similar "supported vs. unsupported" sections here, but update Nathan's language to mention that `semgrep-agent` can now download policy and send notifications assuming the right env variables are set.

docs/integrations.md
```diff
@@ -11,6 +11,20 @@ The following instructions use [Semgrep CI](https://github.com/returntocorp/semg
 !!! danger
     `SEMGREP_APP_TOKEN` is a secret value: DO NOT HARDCODE IT AND LEAK CREDENTIALS. Use your CI provider's secret or environment variable management feature to store it. 
 
+<details><summary>Generic CI Provider</summary>
```
dlukeomalley, line 14: From a search perspective, do you think people will discover this section when they're looking for instructions for "non-supported" CI providers?

dlukeomalley (changes requested):

docs/integrations.md
pabloest: @dlukeomalley How about adding something like, “For Buildkite, CircleCI, [ insert other popular CI provider names], add the following to your workflow:” ?

That way when people search for their particular CI provider by name, this will appear in the results.

dlukeomalley (approved):

PR #42, opened 2020-11-10
Add note about required space before nosem

mschwager:
For https://github.com/returntocorp/semgrep/issues/1991

### DIFF
--- a/docs/running-rules.md
+++ b/docs/running-rules.md
@@ -53,3 +53,6 @@ For example, in Python:
 ```python
 bad_func()  # nosem: rule-id-1
 ```
+
+!!! note
+    The space (` `) before `nosem` is required for Semgrep to detect this annotation.


pabloest (approved):

dlukeomalley:
@mschwager Thank you for the doc updates. What's our thinking on requiring the space versus changing Semgrep CLI to work with and without it?

mschwager:
> @mschwager Thank you for the doc updates. What's our thinking on requiring the space versus changing Semgrep CLI to work with and without it?

Here are the trade-offs on requiring a space: https://github.com/returntocorp/semgrep/issues/1991#issuecomment-724885170

Another nicety is that this implementation Just Works with the new spacegrep functionality. Maintaining `nosem` functionality with comment syntax may be difficult/impossible with spacegrep because it's dealing with arbitrary text.

PR #46, opened 2020-11-12
add documentation on ignores

ievans:
@dlukeomalley -- after this lands we can move a lot of the semgrep action readme to just point here

@pabloest -- do you think "Ignoring Findings" should be its own, top-level section? I think that's a good idea


docs/running-rules.md
```diff
@@ -31,7 +31,9 @@ $ semgrep -e '$X == $X' --lang=py path/to/src
 $ semgrep --config path/to/yaml
 ```
 
-# Managing findings
+# Ignoring findings
```
dlukeomalley, line 34: n.b. - we'll need to add redirects for the docs before merging title changes (because they are used in existing links)

docs/running-rules.md
```diff
@@ -56,3 +59,24 @@ bad_func()  # nosem: rule-id-1
...
 !!! note
     The space (` `) before `nosem` is required for Semgrep to detect this annotation.
+
+
+## Ignoring findings via .gitignore or .semgrepignore
+
+!!! danger
+    Note that `.semgrepignore` is only used by Semgrep CI, and will not be honored when running the regular `semgrep` on the command-line.
+
+Semgrep CLI will respect the precense of a .gitignore in the project and will not scan any paths present in the git ignore. This behavior can be disabled by using the flag `--no-git-ignore`.
```
underyx, line 69:
```suggestion
Semgrep CLI will respect the presense of a .gitignore in the project and will not scan any paths present in the git ignore. This behavior can be disabled by using the flag `--no-git-ignore`. 
```

underyx: I'd also add that it's a known issue that this cannot be disabled in Semgrep CI at the moment: https://github.com/returntocorp/semgrep-action/issues/86

pabloest:
> @dlukeomalley -- after this lands we can move a lot of the semgrep action readme to just point here
> 
> @pabloest -- do you think "Ignoring Findings" should be its own, top-level section? I think that's a good idea

I think we should keep plenty of info in the semgrep action readme because it is auto-shown on the GitHub marketplace page: https://github.com/marketplace/actions/semgrep-action

I agree on the ignoring findings thing.

docs/running-rules.md
ievans: @underyx good point but it's left unstated in the docs, so I'd prefer to just let that issue naturally get resolved

docs/running-rules.md
ievans: Sadly we cannot use redirects here, because any content after the # is not passed to the server. Thoughts on how to handle? IMO this is an argument for having more pages rather than fewer pages with lots of local anchor # links.

pabloest: A crude way would just be to add an HTML element to catch that link and place it right next to the updated header. Something like: `<div id="managing-findings"></div>`

ievans: Good point @pabloest . I ended up just putting the header back in but linking ignore findings.

pabloest (approved):

docs/index.md
```diff
@@ -13,7 +13,7 @@ Explore the Semgrep docs and join an amazing community of engineering and securi
 - [Running rules](running-rules.md)
 - [Managing CI policy](managing-policy.md)
 - [Integrations](integrations.md)
-- [Experiments](experiments.md)
+- [Experiments](experiments)
```
dlukeomalley, line 16: What led to this change? Did the link not work before? 😬

docs/ignoring-findings.md
```diff
@@ -0,0 +1,50 @@
...
+
+If there are files that you want to track in Git but have Semgrep ignore, Semgrep CI (but _not_ the CLI) will respect a `.semgrepignore` file that follows a similar convention to the `.gitignore`. Eg:
+
+```
+.git/
+:include .gitignore
+semgrep-core/tests/
+```
+
+For a complete example, see [the .semgrepignore on the Semgrep source code itself](https://github.com/returntocorp/semgrep/blob/develop/.semgrepignore).
```
dlukeomalley, line 47: nit: use ` around .semgrepignore for consistency with rest of text.

PR #47, opened 2020-11-12
enable strict mode and add a deadlink checker GH action

ievans:


pabloest:
This is super cool!

docs/index.md
```diff
@@ -13,7 +13,7 @@ Explore the Semgrep docs and join an amazing community of engineering and securi
 - [Running rules](running-rules.md)
 - [Managing CI policy](managing-policy.md)
 - [Integrations](integrations.md)
-- [Experiments](experiments.md)
+- [Experiments](experiments)
```
pabloest, line 16: Note: this morning the path to the experiments doc changed to `experiments/overview.md`. I’m not sure what will happen upon merging this PR — probably it won’t fail until merge and then will need another fix for this path?

pabloest (approved):

docs/index.md
ievans: I did pull that in I think @pabloest . This will take the user to "semgrep.dev/docs/experiments" which will then render the overview page, so I think we're ok.

PR #50, opened 2020-11-13
Suggest to search the Semgrep Registry when searching docs returns no results

pabloest:
When one searches the docs and there are no results found, there may still be results available in the Semgrep Registry. This overrides the default “no results found” message with a message that includes a suggestion to go search in the Registry.


pabloest:
Screen Shot 2020-11-20 at 11 03 30 AM

pabloest:
Screen Shot 2020-11-20 at 11 23 01 AM

ievans (approved):

PR #55, opened 2020-11-23
Upload to subfolder in s3 bucket

brendongo:


DrewDennison (approved):

.circleci/config.yml
```diff
@@ -41,6 +41,7 @@ jobs:
             sudo ./aws/install
       - run: aws --version
       - run: aws s3 sync site/ s3://semgrep-docs && aws cloudfront create-invalidation --distribution-id E1PHOFPGTV8DCC --paths '/docs/*'
```
DrewDennison, line 43:
```suggestion
      - run: aws s3 sync site/ s3://semgrep-docs/docs && aws cloudfront create-invalidation --distribution-id E1PHOFPGTV8DCC --paths '/docs/*'
```

brendongo: I added the line below to sync to /docs cause I wasn't sure if something else was relying on it being in root?

DrewDennison: ah makes sense. thanks

PR #56, opened 2020-12-02
Mixture of info from two customers

chmccreery:


dlukeomalley:
@chmccreery Up for reconciling this PR with #84 ?

chmccreery:
Oops, yes. It looks like we should just accept the PR from Moggggg, though, do you agree?

dlukeomalley:
Works for me!

PR #61, opened 2020-12-07
document statements being searched for as expressions

ievans:


docs/writing-rules/pattern-syntax.md
```diff
@@ -421,6 +424,20 @@ It will not match the following:
 import foo
 ```
 
+### Statements as expressions
+
+`foo()` is an expression
+
+`foo();` is a statement.
+
+If youre search pattern is a statment, Semgrep will automatically try to search for it as _both_ an expression and a statement.
```
pabloest, line 433:
```suggestion
If your search pattern is a statement, Semgrep will automatically try to search for it as _both_ an expression and a statement. 
```

docs/writing-rules/pattern-syntax.md
```diff
@@ -421,6 +424,20 @@ It will not match the following:
...
+
+`foo()` is an expression
+
+`foo();` is a statement.
+
+If youre search pattern is a statment, Semgrep will automatically try to search for it as _both_ an expression and a statement. 
+
+When you write the expression `foo()` in a pattern, Semgrep will visit every expression and sub-expression in your program and try to find a match. 
+
+What about a statement like `foo();`? (Note that `foo()` is confusingly called an expression statement, but it is first a statement). Many programmers don't really see the difference between `foo()` and `foo();`. This is why when one looks for `foo()`; Semgrep thinks the user wants to match statements like `a = foo();`, or `print(foo());`.
```
pabloest, line 437: >What about a statement like `foo();`? (Note that `foo()` is confusingly called an expression statement, but it is first a statement). Many programmers don't really see the difference between `foo()` and `foo();`. This is why when one looks for `foo()`; Semgrep thinks the user wants to match statements like `a = foo();`, or `print(foo());`.

Is the semicolon in the last sentence (just after `foo()`) needed? I think it's throwing me off here.

docs/writing-rules/pattern-syntax.md
```diff
@@ -407,6 +407,9 @@ The deep expression operator works in:
 
 ## Statements types
 
+Many programming languages differentiate between expressions and statements. Those are 2 different things. Expressions can appear inside if conditions, in function call arguments, etc. Statements can not appear everywhere; they are sequence of operations (with ; as a separator/terminator) or special control flow constructs (if, while, etc.).
+
+
 Semgrep handles statements differently from other expressions like imports. For example, the pattern `foo` will match these statements:
```
aryx, line 410: Actually import are not really expressions

aryx: maybe "from other expressions like additions or function calls." ?

aryx: actually the whole paragraph confuses statements and expressions I think, partly because in Python you don't need the semicolon.
The 3 things below are really expressions, not statements. 
They are statements when they are used in a sequence of operations, but bar + foo is weird to use in a sequence (unless you put a 'return' before). cc @mschwager who might have written that.

docs/writing-rules/pattern-syntax.md
```diff
@@ -407,6 +407,9 @@ The deep expression operator works in:
 
 ## Statements types
 
+Many programming languages differentiate between expressions and statements. Those are 2 different things. Expressions can appear inside if conditions, in function call arguments, etc. Statements can not appear everywhere; they are sequence of operations (with ; as a separator/terminator) or special control flow constructs (if, while, etc.).
```
aryx, line 410: I would put this chapter below, next to "Statements as expressions"

docs/writing-rules/pattern-syntax.md
```diff
@@ -421,6 +424,20 @@ It will not match the following:
 import foo
 ```
 
+### Statements as expressions
+
```
aryx, line 428: I would put the little paragraph on expr vs statements here.
Then you could add:
"Note that in some programming languages such as Python, which does not use semicolons as a separator or terminator, things are even more confusing, but indentations in Python matters, and a newline after foo() is really the same than foo(); in other programming languages such as C."

docs/writing-rules/pattern-syntax.md
aryx: In some languages such as C yes, you need the semicolon. In python no, but you need a newline, which is treated internally like a semicolon, as a way to terminate a statement.

aryx (changes requested):
thx for doing that.
I know I should be the one doing that ...  I just procrastinate so much when it comes
to writing documentation (or writing anything in general).

dlukeomalley:
@ievans Is this still relevant, and if so, up for resolving conflicts and comments?

ievans:
nope closing in favor of https://github.com/returntocorp/semgrep-docs/pull/83

ievans:
Addressed changes. @aryx @pabloest @dlukeomalley want to stamp it?

pabloest (approved):

PR #74, opened 2021-01-06
Fix a copypasto (the metavariable is $F, not $FILE) in `Rule syntax', `pattern-not-inside' paragraph

iamleot:


minusworld (approved):
Whoops! Thanks! 🙏

brendongo (approved):

PR #75, opened 2021-01-14
Add explanation for each exit code

brendongo:


dvyakimov:
I had 7 error as I didn't insert the language into the rules

brendongo:
Oh looks like it is also used for configs with missing fields. Will update. Thanks!

pabloest (approved):

PR #77, opened 2021-01-17
Add 'community' tier to support expectations

minusworld:
In prep for adding these to the list:
* R
* Rust
* Lua
* C#

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -44,6 +44,9 @@
 
 # Support expectations
 
+* **Community**: this language is supported by open-source contributions.
+    * Support is given by community maintainers.
+    * r2c plans to give these mainline support in the future.
 * **Alpha**: experimental support with many known bugs.
     * Looking for dedicated users to help us improve these languages.
     * Expect limited support responses, as these languages will be lowest priority.


mschwager:
I think we need to iron this definition out more. Questions that come to mind:

- What level of maturity does "community" entail? IMO this should be >= alpha. A language is mostly unusable for regular users until it's hit alpha maturity.
- Who actually owns this? How many community maintainers do we require for a language, 1, 2, other? A [`CODEOWNERS`](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners) file probably makes sense here. IMO we need Github users to commit to owning a language and its continued support.
- What are our support expectations for community languages? A week, a month, anything? Can owners commit to these expectations? What happens if owners are unresponsive?
- How do we remove community supported languages? What if owners can't undertake the maintenance? What if we can't get a hold of owners?

We don't necessarily need to answer _all_ these questions, but we should take them into consideration. My concern is that we'll take on a lot of sub-alpha-level language support, which incurs a large support cost by us, and community language owners may never be heard from again.

mschwager (changes requested):
Aside from the general questions above, if we're going to include the "community" tier we either need to include it in the [maturity definitions](https://semgrep.dev/docs/status/#maturity-definitions), or set community at a maturity level like >= alpha.

minusworld:
Good questions.



> What level of maturity does "community" entail? IMO this should be >= alpha. A language is mostly unusable for regular users until it's hit alpha maturity.

In my mind, the Community tier is something altogether different than the Alpha->Beta->GA path.  Community means r2c doesn't officially work on it yet, and that means that support expectations are different than languages that we own. Community also means that we -will- take it under our wing at some point.

> Who actually owns this? How many community maintainers do we require for a language, 1, 2, other? A CODEOWNERS file probably makes sense here. IMO we need Github users to commit to owning a language and its continued support.

I'd love to name the original contributors who added the language as owners - I'm not actually sure the best way to do that. CODEOWNERS sounds like it would work 👍 

> What are our support expectations for community languages? A week, a month, anything? Can owners commit to these expectations? What happens if owners are unresponsive?

Part of the reason for calling out a Community Tier is because we just added four languages in the last month, and our support burden is already pretty high. Community Tier is basically our way of saying "we'll work on these languages in the future, but for now you'll need to go to the community maintainer."

> How do we remove community supported languages? What if owners can't undertake the maintenance? What if we can't get a hold of owners?

We do actually want to bring them to mainline support in the future, but for now I think there's not an expectation that we fix things right away.

pabloest:
I wonder if there’s a different name we can use to label these languages? My concern is that Semgrep already has a Community tier (for the app) and by placing languages in a Community “state” there may be some confusion due to perceived coupling between these languages and that tier. 

I don’t have any amazing ideas at the moment but was wondering if y’all have thought about it already. The best I’ve come up with so far is “contributed” but it also has some downsides.

mschwager:
Taking a step back here, the goal of these docs is to provide:

1. A consistent set of expectations and information that users and customers can rely on to receive high-quality support from us.
2. Expectations that we place on to users and customers so that we can achieve (1).

I'm not sure that we've thought about this enough such that we can provide (1) and (2). Perhaps it would be better to take a step back and come up with a "community" tier proposal for review. I know we have a lot of exciting new languages coming down the pike, but perhaps it's best to re-group and take a more formalized stab at this after launch. Especially considering the launch doesn't include any of this work. Thoughts?

> In my mind, the Community tier is something altogether different than the Alpha->Beta->GA path.

That makes sense to me, it sounds like we're using the community tier to shift support to community members, which can be beneficial for both parties. However, I think we still want to ensure _some_ bar has been passed before this happens. E.g. if we claim "Lua community support" and patterns don't work in Lua, do we really have any support? This wouldn't be a great user experience IMO. Also, passing a user off to a potentially unresponsive third-party doesn't feel great either.

> Community also means that we -will- take it under our wing at some point.

> Community Tier is basically our way of saying "we'll work on these languages in the future, but for now you'll need to go to the community maintainer."

> We do actually want to bring them to mainline support in the future, but for now I think there's not an expectation that we fix things right away.

These sound like open questions that we'll likely want answered before announcing this tier.

My main concern is that we provide _clear and concise_ information to our external users and customers, and less that we've hashed it all out here. As someone who's done a lot of OSS work, these issues will crop up. We really don't want to end up in a spot where we have 10 sub-alpha languages with unresponsive maintainers, and we have to A.) remove all claims of support for the languages (a big let down for excited users), or B.) have to take on a potentially unbearable maintenance burden. IMO a little forethought here can provide a much better experience for our users down the line.

ievans:
great discussion here @mschwager @minusworld, just chiming in to +1 to Pablo's comment, I'd suggest phrasing the support as "experimental" < alpha < beta < GA

minusworld:
Based on discussions here and elsewhere, I'm closing this out in favor of different veriage.

PR #78, opened 2021-01-20
Recommend checkout@v2 in GHA example

underyx:


dlukeomalley (approved):

dlukeomalley:
@underyx Approved. Anywhere else we need to change this on Semgrep.dev, Actions repos, or other?

PR #85, opened 2021-02-02
Un-nest sidebar links to simplify nav menu, fixes #68

### LINKED ISSUE #68: Sidebar always thinks I'm in the "Writing rules" section
Screen Shot 2020-12-14 at 15 08 36

After clicking on another sidebar item, it flashes correctly for a second and then resets to highlighting this.

pabloest:
Preview: Screen Shot 2021-02-01 at 4 09 18 PM

Also note [this comment](https://github.com/returntocorp/semgrep-docs/issues/68#issuecomment-771259333) from @chmccreery: 

> “I like it! The sections don't all make sense to me immediately (where would I go for Semgrep Action documentation, for example?), but otherwise I am not bothered by the lower amount of nesting.

cc @underyx and @dlukeomalley

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -2,23 +2,19 @@ site_name: Semgrep Docs
 site_url: https://semgrep.dev/docs/
 strict: true
 nav:
-  - Semgrep:
-      - Home: index.md
-      - getting-started.md
-      - Writing rules:
-          - writing-rules/overview.md
-          - Syntax:
-              - writing-rules/pattern-syntax.md
-              - writing-rules/rule-syntax.md
-          - Examples:
-              - writing-rules/pattern-examples.md
-              - writing-rules/rule-ideas.md
-          - writing-rules/testing-rules.md
-      - running-rules.md
-      - ignoring-findings.md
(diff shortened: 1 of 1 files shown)


dlukeomalley:
Tentatively lgtm, though I'd like to simplify the sections if possible. Perhaps Rule Syntax and Pattern Syntax can become "Syntax", Rule Examples and Pattern Examples can become Examples, etc.

pabloest:
Thanks, @dlukeomalley. I am mildly in favor of the simplification, my main concern being overall length (they’re already pretty long). 

In the interest of making progress and not complicating things too much, can we merge this and open a separate issue for the section simplification? A simplification change will require redirects and possibly updating links/labels outside the docs.

chmccreery (approved):
Agree with @dlukeomalley on simplifying or rethinking which categories go on the left hand toolbar, but also in favor of getting this PR in since it is fixing a bug and that is sort of a separate problem

pabloest:
Moving discussion over to a related PR: https://github.com/returntocorp/semgrep-docs/pull/91

PR #88, opened 2021-02-05
Add instructions to fail open

brendongo:


docs/managing-policy.md
```diff
@@ -42,3 +42,8 @@ To receive notifications via third-party services, like Slack or email:
 
 When Semgrep CI next runs and finds a result, the configured services will receive the finding.
 
+# Fail Open/Close
+
+By default, when semgrep-action fails to scan code for some reason (i.e. an invalid config), it will exit with a return code greater than 1. If you do not want this kind of failure to block CI from passing, you can set semgrep-action to fail open on semgrep.dev. Go to https://semgrep.dev/manage/projects and on the desired repository edit, select, and save the checkbox under the Fail Open column.
+
+When semgrep-action is set to fail-open it will notify available notification channels of the failure to perform a scan and then exits with return code 0.
```
pabloest, line 49:
```suggestion
When semgrep-action is set to fail open it will notify available notification channels of the failure to perform a scan and then exits with return code 0.
```

docs/managing-policy.md
```diff
@@ -42,3 +42,8 @@ To receive notifications via third-party services, like Slack or email:
 
 When Semgrep CI next runs and finds a result, the configured services will receive the finding.
 
+# Fail Open/Close
+
+By default, when semgrep-action fails to scan code for some reason (i.e. an invalid config), it will exit with a return code greater than 1. If you do not want this kind of failure to block CI from passing, you can set semgrep-action to fail open on semgrep.dev. Go to https://semgrep.dev/manage/projects and on the desired repository edit, select, and save the checkbox under the Fail Open column.
```
pabloest, line 47:
```suggestion
By default, when semgrep-action fails to scan code for some reason (e.g., an invalid config), it will exit with a return code greater than 1. If you do not want this kind of failure to block CI from passing, you can set semgrep-action to fail open. Go to [https://semgrep.dev/manage/projects](https://semgrep.dev/manage/projects) and on the desired repository edit, select, and save the checkbox under the Fail Open column.
```

docs/managing-policy.md
```diff
@@ -42,3 +42,8 @@ To receive notifications via third-party services, like Slack or email:
 
 When Semgrep CI next runs and finds a result, the configured services will receive the finding.
 
+# Fail Open/Close
```
pabloest, line 45:
```suggestion
# Fail open/close
```

Also: I’m not sure about the terminology here but I would have thought to use "Fail closed" instead of Fail close"?

pabloest (approved):

docs/managing-policy.md
```diff
@@ -42,3 +42,8 @@ To receive notifications via third-party services, like Slack or email:
 
 When Semgrep CI next runs and finds a result, the configured services will receive the finding.
 
+# Fail open/close
+
+By default, when semgrep-action fails to scan code for some reason (e.g., an invalid config), it will exit with a return code greater than 1. If you do not want this kind of failure to block CI from passing, you can set semgrep-action to fail open. Go to [https://semgrep.dev/manage/projects](https://semgrep.dev/manage/projects) and on the desired repository edit, select, and save the checkbox under the Fail Open column.
```
dlukeomalley, line 47: @pabloest Should we standardize on "Semgrep CI" rather than `semgrep-action` here?

@brendongo Is there a reference for what "fail open" means? I believe it's a term of art that may be helpful to explain or link to.

Rather than "fails to scan code", what do you think of language like "internal Semgrep error"? It feels more concise/accurate than than "for some reason" ;)

dlukeomalley, line 47:
```suggestion
By default, Semgrep CI returns a non-zero exit code when it has an internal error (e.g. invalid configuration, parse error). This causes a build failure and will prevent merging.

To prevent Semgrep CI from interfering with the development flow due to internal errors, enable "fail open" on [https://semgrep.dev/manage/projects](https://semgrep.dev/manage/projects) for your desired repository. Click edit, select, and save the checkbox under the Fail Open column.
```

Played with some copy-edits here. How do we make it clearer what problem "fail open" solves for users? Top of mind for me is that it makes Semgrep easier to adopt, reduces downside, and is only appropriate in cases where it's acceptable to merge unscanned code.

docs/managing-policy.md
```diff
@@ -42,3 +42,8 @@ To receive notifications via third-party services, like Slack or email:
...
+
+When semgrep-action is set to fail open it will notify available notification channels of the failure to perform a scan and then exit with return code 0.
```
dlukeomalley, line 49:
```suggestion
When fail open is enabled a notification will be sent to your configured notification channels and Semgrep CI will continue with a 0 return code.
```

Maybe something like ^?

What happens if I don't have notifications enabled?

docs/managing-policy.md
pabloest: @dlukeomalley I think yes, and also think we should probably do that more consistently? Perhaps even rename the action’s repo? (something for another day though)

PR #92, opened 2021-02-08
Replace nosem with nosemgrep

pabloest:
and some copy edits

### DIFF
--- a/docs/ignoring-findings.md
+++ b/docs/ignoring-findings.md
@@ -4,7 +4,7 @@
 
 # Ignoring findings via inline comments
 
-Semgrep allows for ignoring findings in code by specifying a `nosemgrep` comment on the first line of a finding. Comments take the form of `nosemgrep` or `nosemgrep: <rule-id>`. This functionality works across languages.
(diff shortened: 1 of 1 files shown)


docs/ignoring-findings.md
```diff
@@ -4,47 +4,47 @@
 
 # Ignoring findings via inline comments
 
-Semgrep allows for ignoring, or whitelisting, findings in code by specifying a `nosem` comment on the first line of a finding. Comments take the form of `nosem` or `nosem: <rule-id>`. This functionality works across languages.
+Semgrep allows for ignoring findings in code by specifying a `nosemgrep` comment on the first line of a finding. Comments take the form of `nosemgrep` or `nosemgrep: <rule-id>`. This functionality works across languages.
```
dlukeomalley, line 7: Thoughts on including a reference to `nosem` as well since it does work and historical findings may already have the comment applied?

dlukeomalley (approved):
lgtm, with one comment to still mention `nosem` as a no-longer idiomatic way of ignoring results (but still works).

docs/ignoring-findings.md
pabloest: Yeah good point. I’ll try to keep it short.

PR #93, opened 2021-02-08
Make supported languages table taller so viewers don’t have to scroll

pabloest:


dlukeomalley:
@pabloest @DrewDennison Is there an approach that dynamically scales the iframe height to match the contents? Next time a language is added we'll have to update this again.

pabloest:
AFAIK yes but it involves writing JavaScript that sends messages between iframe and parent windows, then dynamically adjusts the DOM element to reflect the size communicated in the message. 
🍝

dlukeomalley:
This change seems moot with changes I see on /docs:

pabloest:
@dlukeomalley That’s on a different page. One is the index page (overview), the other is the language status page. This PR is for the latter.

PR #104, opened 2021-02-10
Rearrange files and add semgrep CI section

brendongo:


brendongo:
Screen Shot 2021-02-09 at 4 48 34 PM
Rearranged some sections

brendongo:
In particular I am reserving the word integrations to mean the downstream notification channels and using the word provider to reference where semgrep is run (github, gitlab, circleci, travisci, etc.)

brendongo:
Outside scope of this PR: adding more copy to the Notifications/Integrations page

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
+# Semgrep CI
```
brendongo, line 1: This is the only new copy @pabloest the other files move things around

docs/providers.md
```diff
@@ -0,0 +1,213 @@
+# Continuous Integration (CI) Providers
```
brendongo, line 1: Nothing on this page is new. Copy review out of scope of this PR

docs/semgrep-ci.md
pabloest: Thanks for pointing this out :)

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
...
+
+[Semgrep CI](https://github.com/returntocorp/semgrep-action) is a wrapper around [Semgrep](https://github.com/returntocorp/semgrep) for running as a GitHub Action, in Gitlab, and in other CI providers and interfacing with [https://semgrep.dev](https://semgrep.dev). See [CI Providers](providers.md) to see how to setup Semgrep CI in the provider you already use.
```
pabloest, line 3:
```suggestion
[Semgrep CI](https://github.com/returntocorp/semgrep-action) is a wrapper around [Semgrep](https://github.com/returntocorp/semgrep) for running as a GitHub Action, in GitLab, and in other CI providers and interfacing with [Semgrep App](https://semgrep.dev). See [CI Providers](providers.md) to see how to setup Semgrep CI in the provider you already use.
```

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
...
+
+Semgrep CI adds a layer of git-awareness on top of Semgrep. It reviews only the changed files in pull requests with Semgrep whenever a new commit is added to them, and reports only issues that are newly introduced in that pull request.
```
pabloest, line 5:
```suggestion
Semgrep CI adds a layer of git-awareness on top of Semgrep. Whenever a new commit is added to a pull request, Semgrep CI reviews only the changed files and reports only issues that are newly introduced in that pull request.
```

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Technical details
+
+Semgrep CI scans files in the current directory with [semgrep](https://github.com/returntocorp/semgrep), and exits with a return code of 1 if blocking issues are found.
+
+Findings are blocking by default. They can be set to non-blocking by changing the action in semgrep.dev/manage/policy.
```
pabloest, line 11:
```suggestion
Findings are blocking by default. They can be set to non-blocking by changing the action in Semgrep App [Manage → Policy](https://semgrep.dev/manage/policy).
```

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
...
+
+Semgrep-action has the option to report only new issues, added since a specific commit.
```
pabloest, line 13: I am slightly confused here (and in a few places below) by Semgrep-action vs. Semgrep CI.

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
...
+When run in a continuous integration (CI) pipeline, semgrep-action determines the base commit from [environment variables](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/meta.py), as set by GitHub, GitLab, Travis or CircleCI. The base commit can also be passed on the command line using the option --baseline-ref.
```
pabloest, line 14:
```suggestion
When run in a continuous integration (CI) pipeline, semgrep-action determines the base commit from [environment variables](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/meta.py), as set by GitHub, GitLab, Travis CI or CircleCI. The base commit can also be passed on the command line using the option `--baseline-ref`.
```

docs/semgrep-ci.md
```diff
@@ -0,0 +1,60 @@
...
+
+Semgrep-action determines new issues by only [scanning modified files](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/targets.py), and scanning twice. It scans the current commit, checks out the base commit and scans that, and removes previously existing findings from the scan result. [Findings are compared](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/findings.py) on identifier, file path, code and count. If the identifier of a rule is modified in the semgrep configuration, or if the file containing the issues is renamed, all findings are considered new. Changing code that is matched by a rule will thus result in a new finding, even though the finding was previously present and the change did not introduce it.
```
pabloest, line 16: This is a super paragraph! I think I never really understood how the Action works until now.

Matching the structure you used above which I like ([verb] only):

```suggestion
Semgrep-action determines new issues by [scanning only modified files](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/targets.py), and scanning twice. It scans the current commit, checks out the base commit and scans that, and removes previously existing findings from the scan result. [Findings are compared](https://github.com/returntocorp/semgrep-action/blob/develop/src/semgrep_agent/findings.py) on identifier, file path, code and count. If the identifier of a rule is modified in the Semgrep configuration, or if the file containing the issues is renamed, all findings are considered new. Changing code that is matched by a rule will thus result in a new finding, even though the finding was previously present and the change did not introduce it.
```

brendongo: Just copied a lot from the semgrep-action git repo :D

docs/semgrep-ci.md
brendongo: What if I remove this paragraph and below the paragraph below have something like 
```
When running in Github or GitLab, Semgrep CI is able to determine the base commit using environment variables set but the CI provider. For other providers, Semgrep CI will need to be explicitly told the base commit. See setup instructions for other providers [Link here] for more info.
```

pabloest: Thanks @brendongo, I think that’s helpful.

brendongo: Made this change

pabloest (approved):

PR #105, opened 2021-02-10
Add semgrep cli help text for SEO

brendongo:


DrewDennison (approved):

docs/help-text.md
```diff
@@ -0,0 +1,151 @@
+# Other arguments
```
pabloest, line 1: How about naming this title and the file something like “CLI reference”?

pabloest (commented):
Love this idea, @brendongo! Thank you. I left two comments.

docs/help-text.md
brendongo: Changed!

brendongo:
Made suggested changes

pabloest:
Thanks, @brendongo !

pabloest:
@brendongo I just realized we already have this page: https://semgrep.dev/docs/cli-usage/

How about merging the content? I would propose moving that content (exit codes, primarily) into the file you created (`cli-reference`).

PR #106, opened 2021-02-10
Increase languages table iframe width

mschwager:
Before:

image

After:

image

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -5,7 +5,7 @@
 # Language maturity
 
 <div class="languages-table">
-  <iframe width="550" height="1000" border=0 frameBorder=0 src="https://dashboard.semgrep.dev/languages/table"></iframe>
+  <iframe width="100%" height="1000" border=0 frameBorder=0 src="https://dashboard.semgrep.dev/languages/table"></iframe>
 </div>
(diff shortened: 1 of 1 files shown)


docs/status.md
```diff
@@ -5,7 +5,7 @@
 # Language maturity
 
 <div class="languages-table">
-  <iframe width="550" height="1000" border=0 frameBorder=0 src="https://dashboard.semgrep.dev/languages/table"></iframe>
+  <iframe width="100%" height="1000" border=0 frameBorder=0 src="https://dashboard.semgrep.dev/languages/table"></iframe>
```
pabloest, line 8: iframe widths must be specified in pixels: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe

I suppose most browsers just render this correctly anyway?

pabloest (approved):

docs/status.md
mschwager: Huh, interesting. I checked the view without specifying `width` at all, and it looks like it defaults to 300 pxs, which is way to narrow. I guess we can roll with this for now until becomes problematic :man_shrugging:

PR #116, opened 2021-02-19
adding specifications for metavariable-regex

colleend:
Trying to specify that a metavariable matching a string includes the quotes as part of the fix for https://github.com/returntocorp/semgrep/issues/2578


chmccreery (approved):
I love it! Very clear, nice work!

mschwager:
IMO, I don't find it particular helpful to list out things that _don't_ work. The list of things that don't work is infinite, so I find it best to stick to working examples. Also, since the code example and rules are fairly large it's not easy to tell what the difference between the correct and incorrect rules are. Can we simply include the "Note" here instead of re-working the examples? If we need to provide an example we could link out to something like: https://semgrep.dev/s/mschwager:include-quotes

colleend:
good point that there's an infinite amount of non-working examples. I'll remove the part that lists out the example that doesn't work and include a note with another example.

PR #118, opened 2021-02-22
Promote 'metavariable-comparison' from experimental

mschwager:
`metavariable-comparison` has been hanging out in experimental for 4 months. Since then we've added it to 15 rules:

```
$ rg 'metavariable-comparison' | wc -l
15
```

My concern is that people won't find it if it's hiding under experiments, and may resort to `pattern-where-python`. Lets bring this thing out for primetime :+1:

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -41,6 +41,7 @@ The below optional fields must reside underneath a `patterns` or `pattern-either
 | Field                                                           | Type     | Description                                                                                                              |
 | :-------------------------------------------------------------- | :------- | :----------------------------------------------------------------------------------------------------------------------- |
 | [`metavariable-regex`](#metavariable-regex)     | `map`    | Search metavariables for [Python `re`](https://docs.python.org/3/library/re.html#re.match) compatible expressions |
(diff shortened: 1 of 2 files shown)


docs/writing-rules/rule-syntax.md
```diff
@@ -191,6 +192,55 @@ rules:
...
+
+```
+- pattern: to_integer($ARG)
+- metavariable-comparison:
+    metavariable: '$ARG'
+    comparison: '$ARG > 2147483647'
+    strip: true
+```
+
+This will remove quotes (`'`, `"`, and `` ` ``) from both ends of the metavariable content. So `"2147483648"` will be detected but `"2147483646"` will not. This is useful when you expect strings to contain integer or float data.
```
pabloest, line 242: I see the numerical difference between `"2147483648"` and `"2147483646"`,  but the quotes are still included in both - am I missing something? It’s just that the comparison is against 2147483647? If that's the case, it might be more straightforward to just include one the `"2147483648"` example.

pabloest (approved):

docs/writing-rules/rule-syntax.md
mschwager: Yeah, it's to highlight that these are strings in code, but are treated as integers in the comparison. This option isn't frequently used, but we do have a few places in `semgrep-rules` making this type of comparison. I'll roll with this for now so it's a simple copy and paste from experiments :+1:

PR #129, opened 2021-03-11
Add an explanation about metavariable matching behavior in pattern-no…

minusworld:
…t\* that addresses https://github.com/returntocorp/semgrep/issues/2305

Closes https://github.com/returntocorp/semgrep/issues/2305

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -284,6 +284,16 @@ The above rule looks for use of Django’s [`FloatField`](https://docs.djangopro
 
 # Metavariable matching
 
+Semgrep will enforce equivalence between pattern clauses with metavariables of the same name. For example, the following patterns will detect when a function is called inside itself recursively.
+
+```yaml
+patterns:
+- pattern-inside: |
+    def $FUNC(...):
+      ...
+- pattern: $FUNC(...)
+```
+
 Metavariable matching operates differently for logical AND (`patterns`) and logical OR (`pattern-either`) parent operators. Behavior is consistent across all child operators: `pattern`, `pattern-not`, `pattern-regex`, `pattern-inside`, `pattern-not-inside`.
 
 ## Metavariables in logical ANDs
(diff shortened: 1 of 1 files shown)


mschwager:
Hmm, there seems to be all kinds of corner cases with this. E.g.

```yaml
rules:
- id: matching-negation
  message: Matching negation
  severity: ERROR
  languages: [python]
  patterns:
    - pattern: $VAR1 = ...
    - pattern-not-inside: |
        def $FUNC(...):
          ...
          $VAR2 = $VAR1
```

```python
a = 5

def func():
    b = a
    return b
```

```
$ python -m semgrep --config /tmp/test.yaml /tmp/test.py
running 1 rules...
/tmp/test.py
severity:error rule:tmp.matching-negation: Matching negation
1:a = 5
--------------------------------------------------------------------------------
4:    b = a
ran 1 rules on 1 files: 2 findings
```

It looks like it's somehow detecting `b` as `$VAR1` in the second finding:

```json
                    "$VAR1": {
                        "start": {
                            "line": 4,
                            "col": 5,
                            "offset": 23
                        },
                        "end": {
                            "line": 4,
                            "col": 6,
                            "offset": 24
                        },
                        "abstract_content": "b",
                        "unique_id": {
                            "type": "id",
                            "value": "b",
                            "kind": "Local",
                            "sid": 2
                        }
                    }
```

Would it make more sense to disable metavariable matching on negation operators? It's a weird operation because `pattern-not-inside` will be true simply if the metavariables don't match, even if the corresponding `pattern` _is_ inside.

aryx:
I don't think we want to disable metavariable matching on negation. We have examples relying on that.

aryx:
and in your example a = 5 is normal, because it's outside the func (so clearly will satisfy pattern-not-inside).
For b = a, b is never found on the right hand side, so the pattern-not-inside returns a match, but with VAR1 bound to a, never bound to b, hence the match

aryx (approved):
looks good to me.

PR #130, opened 2021-03-15
Remove stats-specific template file and use docs-wide template

pabloest:
This page initially used its own template, but there’s no need for that. No other page uses that template, either.

Before:
Screen Shot 2021-03-15 at 1 43 07 PM

After:
Screen Shot 2021-03-15 at 1 44 29 PM


ievans (approved):
lgtm!

PR #141, opened 2021-03-26
Update 404 page and add some redirects

pabloest:


pabloest:
404 before this change:
Screen Shot 2021-03-26 at 9 06 10 AM

404 after this change (the footer is not visible but it is at the bottom of the screen and matches the rest of the docs pages):
Screen Shot 2021-03-26 at 9 06 12 AM

DrewDennison (approved):

PR #156, opened 2021-04-22
Convert rule operator examples to embedded editor widgets

mschwager:


pabloest:
I think it’d be great to move many of the docs code examples to use the embedded editor. I couldn't see this change locally since the iframes won't load, but we already know what this looks like from previous PRs so I think it's fine.

Btw we should check to see if there are any SEO implications of moving code examples to use the embedded editor, but we can do that after this is merged.

pabloest (approved):

PR #159, opened 2021-04-23
fix typo (Ruby, not Python)

gilesbowkett:
Hi — this is a very cool project! Very trivial PR here, hope you don't mind. I showed my team the Rails XSS use case specifically, and in doing so noticed (and fixed) a typo.

XSS_Prevention_for_Ruby_on_Rails_-_Semgrep_Docs


minusworld:
I think those Ruby links are actually dead and aren't representative of an actual problem. I'll remove them in a future PR.

minusworld:
Thanks for the fix @gilesbowkett ! 🙏

PR #173, opened 2021-05-18
Promote C# to alpha

mschwager:


mschwager:
This will be happening in the next release of Semgrep: https://github.com/returntocorp/semgrep/blob/develop/CHANGELOG.md#added

docs/index.md
```diff
@@ -37,14 +37,14 @@ Semgrep is an evolution of [pfff](https://github.com/returntocorp/pfff/), which
...
-| Go         | C                                | C#           |
-| Java       | OCaml                            | Kotlin       |
-| JavaScript | PHP                              | Lua          |
-| JSON       | YAML                             | R            |
-| JSX        | Generic (ERB, Jinja, etc.)       | Rust         |
-| Python     |                                  |              |
-| Ruby       |                                  |              |
-| TypeScript |                                  |              |
-| TSX        |                                  |              |
+| GA         | Alpha                      | Experimental |
```
mschwager, line 40: Can we remove this table in favor of linking to the supported languages doc, or similarly iframing? Maintaining many different lists of the same information seems like a pain.

mschwager: Also, here we call languages "Experimental" and on the support languages doc we say "develop" :confused:

pabloest: The intent of the table is to have a simplified overview of supported languages that doesn’t include extensions and tags which would be pulled in when using an iframe. Additionally, the iframe would have to be periodically updated so that its height accounts for all the content inside and so the user doesn’t have to scroll. 

IMO the table is a bit cleaner and since there’s already multiple places that need to be updated (supported languages on dasboard.semgrep.dev, live editor, Semgrep readme, Semgrep App front page, r2c.dev), I didn’t think it was a big deal to keep this as a table.

@DrewDennison: Matt has a good point about develop vs. experimental. Should we update these docs and the Semgrep readme to use develop instead of experimental, or do you think we should update dashboard.semgrep.dev to use experimental instead of develop?

pabloest: The intent of the table is to have a simplified overview of supported languages that doesn’t include extensions and tags which would be pulled in when using an iframe. Additionally, the iframe would have to be periodically updated so that its height accounts for all the content inside and so the user doesn’t have to scroll. 

IMO the table is a bit cleaner and since there’s already multiple places that need to be updated (supported languages on dasboard.semgrep.dev, live editor, Semgrep readme, Semgrep App front page, r2c.dev), I didn’t think it was a big deal to keep this as a table.

@DrewDennison: Matt has a good point about develop vs. experimental. Should we update these docs and the Semgrep readme to use develop instead of experimental, or do you think we should update dashboard.semgrep.dev to use experimental instead of develop?

pabloest (approved):

mschwager:
Sounds good, we can worry about the other concerns in a separate PR :+1:

PR #178, opened 2021-05-21
Fix two broken links

pabloest:


pabloest:
I'm not sure why the Markdown link checker is failing. I’ve tried 3 ways to link, all of which work locally, but they fail due to getting http code 400 during the check markdown link step.

daghan (approved):

PR #183, opened 2021-05-27
Use embedditor in experiments page

mschwager:
A snippet says a 1000 words :)

### DIFF
--- a/docs/experiments/overview.md
+++ b/docs/experiments/overview.md
@@ -77,7 +77,7 @@ rules:
 
 # Equivalences
 
-Equivalences allow for defining equivalent code patterns (i.e. a commutative property: `$X + $Y <==> $Y + $X`). Equivalence rules use the `equivalences` top-level key and one `equivalence` key for each equivalence.
+Equivalences enable defining equivalent code patterns (i.e. a commutative property: `$X + $Y <==> $Y + $X`). Equivalence rules use the `equivalences` top-level key and one `equivalence` key for each equivalence.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


pabloest:
It does say a 1,000 words!

One downside of this is that we lose some SEO benefits. For example, related to the snippet about Django open redirects, today this docs page is shown on a [Google search for "django open redirect"](https://www.google.com/search?q=django+open+redirect+site%3Asemgrep.dev&oq=django+open+redirect+site%3Asemgrep.dev) as shown by the last result in this screenshot:

Screen Shot 2021-05-27 at 10 02 58 AM

But moving to the Embedditor would cause that last entry to drop out of the search results due to this issue: https://github.com/returntocorp/semgrep-app/issues/2341

Thoughts?

mschwager:
> But moving to the Embedditor would cause that last entry to drop out of the search results due to this issue: returntocorp/semgrep-app#2341
> 
> Thoughts?

I definitely agree with the spirit of #2341, but I don't think it applies in this particular situation. The "django open redirect" has nothing to do with `equivalences`, per se, so the fact that that query lead to the experiments page seems to be a happy accident.

The new example actually doesn't mention Django at all: https://semgrep.dev/s/jNnn. I think this example is much more succinct than the original (https://semgrep.dev/s/AEL), and highlights the `equivalences` functionality much more directly. For someone just trying to understand `equivalences`, the original example is much more complex than it needs to be.

pabloest:
@mschwager perfect, let’s go with this change then :)

pabloest (approved):

PR #184, opened 2021-06-01
Fix broken markdown in example CI configs

Gby56:
It seems to be broken in the current documentation online, there are 4 backticks to close instead of 3, which makes the Jenkins block spill into the others
You can take a look here https://semgrep.dev/docs/sample-ci-configs/#jenkins


pabloest:
This looks great. Thanks so much for finding and fixing this, @Gby56! I’ll approve and merge this.

pabloest (approved):

PR #187, opened 2021-06-04
Add dependabot permissions workaround

chmccreery:
From a user 🎉


zabbal (approved):

dlukeomalley (approved):
lgtm.

@underyx @chmccreery Should this be a line used in our default config in semgrep-app, our Semgrep marketplace listing, and any other lingering config samples?

underyx:
Note that there is some movement to fix this on the dependabot/github side:

- https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544
- https://github.community/t/opt-in-to-allow-secrets-on-dependabot-workflows/183040
- https://github.community/t/use-dependabot-secrets-instead-of-actions-secrets-in-dependabot-workflows/183050

zabbal:
Even if that's fixed on dependabot side, is there really that much sense for the user to run Semgrep scan on 1-line change like "1.2.0" -> "1.2.3"?

chmccreery:
> @underyx @chmccreery Should this be a line used in our default config in semgrep-app, our Semgrep marketplace listing, and any other lingering config samples?

Yes, I think so. Wanted to get discussion about this change in one spot, but now that this is in, will update elsewhere as well

PR #197, opened 2021-06-09
Add line about optimizations to the docs

emjin:


docs/running-rules.md
```diff
@@ -35,6 +35,8 @@ $ semgrep -e '$X == $X' --lang=py path/to/src
 $ semgrep --config path/to/yaml
 ```
 
+We are working on optimizations to improve Semgrep's performance, which may have unexpected behavior. If you encounter a bug, you can run with `--optimizations none` to use the original code path.
```
pabloest, line 38: @emjin Could this be more specific about the changes and the bugs that people may encounter, or how those bugs may present to the user?

One suggestion:
“Changes to the way Semgrep pre-processes rules introduced in Semgrep v0.55.0 include optimizations that improve speed but may sometimes lead to unexpected behavior. If you encounter {X,Y,Z} issues, try running Semgrep with `--optimizations none`.

pabloest:
@emjin Can you also add something to the CLI reference? https://semgrep.dev/docs/cli-usage/

docs/cli-usage.md
```diff
@@ -93,6 +93,8 @@ config:
                         performed.
   --strict              Only invoke semgrep if configuration files(s) are
                         valid.
+  --optimizations [OPTIMIZATIONS]
+                        Turn on/off optimizations. Default = 'none'. Use 'all' to                         turn all optimizations on.
```
pabloest, line 97:
```suggestion
                        Turn on/off optimizations. Default = 'none'. Use 'all' to turn all optimizations on.
```

docs/running-rules.md
```diff
@@ -35,7 +35,9 @@ $ semgrep -e '$X == $X' --lang=py path/to/src
 $ semgrep --config path/to/yaml
 ```
 
-We are working on optimizations to improve Semgrep's performance, which may have unexpected behavior. If you encounter a bug, you can run with `--optimizations none` to use the original code path. 
+We are working on optimizations to improve Semgrep's performance, which necessitates changing how rules are processed. If you are using v0.55.0 or later and you encounter an unexpected metavariable binding or missing result, try running with `--optimizations none` to use the original code path. 
+
+You may find that some files that were previously parsed are now skipped; this will happen if Semgrep can confirm the rule will not match the file without parsing it. You can similarly run with `--optimizations none` to avoid this.
```
pabloest, line 40: This is great!

pabloest (approved):
Left one suggestion to remove seemingly unneeded spacing in the CLI help.

PR #202, opened 2021-06-14
Add meta descriptions to more pages

underyx:


custom_theme/cheat-sheet-base.html
```diff
@@ -6,6 +6,7 @@
   <meta charset="utf-8">
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <meta name="description" content="{{ page.meta.summary }}">
```
pabloest, line 9: Does this conflict with the line that immediately follows? Or is it ok to have two meta descriptions? (I just realized the base theme may have the same issue btw)

underyx: Oh, I'm not sure, didn't even see that! but `page.is_homepage` will only be true on the `/docs/` page, right? So the next line can never be active on a cheat sheet page I think?

pabloest: Yeah I think so. So that `is_homepage` line is probably useless here.

pabloest (approved):

PR #200, opened 2021-06-09
Remove universal analytics and add GA4 analytics

pabloest:
According to [this note from mkdocs](https://www.mkdocs.org/about/release-notes/#backward-incompatible-changes-in-12):

> So long as the old "UA-" ID and new "G-" ID are properly linked in your Google account, and you are using the "G-" ID, the data will be made available in both the old and new formats by Google Analytics.

Note I’m not able to get this working locally yet. I don’t see the GA tag loading on the local page and when running in verbose mode, I see what seems to be mkdocs not picking up the new gtag configuration:
```
Loaded theme configuration for 'readthedocs' from '/usr/local/lib/python3.9/site-packages/mkdocs/themes/readthedocs/mkdocs_theme.yml': {'static_templates': ['404.html'], 'locale': 'en',
            'include_search_page': True, 'search_index_only': False, 'highlightjs': True, 'hljs_languages': [], 'analytics': {'gtag': None}, 'include_homepage_in_sidebar': True,
            'prev_next_buttons_location': 'bottom', 'navigation_depth': 4, 'titles_only': False, 'sticky_navigation': True, 'collapse_navigation': True}
```

@underyx perhaps it’s something wrong with my yaml? I’m intending to follow what’s [described here](https://www.mkdocs.org/user-guide/choosing-your-theme/#readthedocs).

PR #204, opened 2021-06-16
Fix mkdocs version used for deploying

underyx:
this syncs to the version intended to be used since #200


pabloest:
Thanks for this, @underyx

PR #217, opened 2021-06-30
Updated metavariable-comparison docs to include example of complex expression

jtimmons:
Following up on the discussion from returntocorp/semgrep#3424.

Added an example of using arithmetic and boolean operators in `metavariable-comparison` to highlight the fact that it can perform more complex expressions than just simple comparisons.

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -140,6 +140,12 @@ The `metavariable-comparison` operator is a mapping which requires the `metavari
 
 This will catch code like `set_port(80)` or `set_port(443)`, but not `set_port(8080)`.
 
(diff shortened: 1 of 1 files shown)


pabloest (approved):
Thanks so much for this contribution, @jtimmons!

pabloest:
Thanks for this PR, @jtimmons !

To express our thanks, we’d love to send you some swag. If that’s ok, could you fill out some details on [this form](https://r2c.dev/swag) so we can send it to you?

jtimmons:
@pabloest thanks for the offer! I filled one out as part of [this change](https://github.com/returntocorp/semgrep/pull/3424#issuecomment-869732739) already so I think I'm all set!

PR #219, opened 2021-07-01
writing-rules: New features in 0.57.0

IagoAbal:


docs/writing-rules/rule-syntax.md
```diff
@@ -304,6 +317,16 @@ def foo(something):
...
+# `options`
+
+Enable/disable the following matching features:
+
+| Option                 | Default | Description                                                            |
+| :--------------------- | :------ | :--------------------------------------------------------------------- |
+| `constant_propagation` | `true`  | [Constant propagation](./pattern-syntax.md#constants), including [intra-procedural flow-sensitive constant propagation](../experiments/overview.md#constant-propagation). |
+| `ac_matching`          | `true`  | [Matching modulo associativity and commutativity](./pattern-syntax.md#associative-and-commutative-operators), we treat boolean AND/OR as associative, and bitwise AND/OR/XOR as both associative and commutative. |
+| `commutative_boolop`   | `false` | Treat boolean AND/OR as commutative even if not semantically accurate. |
+
```
aryx, line 329: I would have added a link to the Config_semgrep.atd if people want the full list

aryx (approved):

docs/writing-rules/rule-syntax.md
IagoAbal: I added the link, but warned that these other options are experimental and may change or disappear without notice (maybe the warning is a bit exaggerated... but just in case).  I can also document the other options the same way, but I was unsure about the long-term plans there since we have several comments pointing out that things may change in the future.

aryx: sounds good

aryx (approved):

PR #220, opened 2021-07-02
Adjusting pattern-syntax.md with recent features

aryx:
This diff mostly reorganize and fix what was there already.
Next commit will document some new features like /$XXX/.

### DIFF
--- a/docs/writing-rules/pattern-syntax.md
+++ b/docs/writing-rules/pattern-syntax.md
@@ -138,6 +138,15 @@ crypto.set_secret_key("HARDCODED SECRET")
 
 This also works with [constant propagation](#constants).
 
+In languages where regular expressions use a special syntax
+(e.g., Javascript), the pattern `/.../` will match
+any regular expression construct:
+
+```javascript
+re1 = /foo|bar/;
+re2 = /a.*b/;
+```
+
 ## Binary operations
 
 The ellipsis operator can match any number of arguments to binary operations. The pattern `$X = 1 + 2 + ...` matches:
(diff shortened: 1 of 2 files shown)


docs/writing-rules/pattern-syntax.md
```diff
@@ -473,16 +493,43 @@ Many programmers don't really see the difference between `foo()` and `foo();`. T
...
 ```
 
-```text
-pattern: if $CONDITION:
+A complete expression is needed (like `1 + $X`)
+
+## Partial statements
+
+Partial statements are partially supported. For example,
+you can just match the header of a conditional with `if ($E)`,
```
IagoAbal, line 509: I am not fully aware about where this is particularly useful. When is this preferred to `if ($E) ...` ?

IagoAbal: Is this only because you want to match that fragment of the code and nothing else?

aryx: Can be useful when used in a pattern-inside, to restrict the context in which to search code

aryx: I can probaly add that as a comment

docs/writing-rules/pattern-syntax.md
```diff
@@ -473,16 +493,43 @@ Many programmers don't really see the difference between `foo()` and `foo();`. T
...
+function `foo`. In the same way, you can just match a class header
+(e.g., with `class $A`).
+
+# Deprecated features
+
+## String matching
+
+!!! warning
+    String matching has been deprecated. You should use 
+	[metavariable regexp](rule-syntax.md#metavariable-regex) instead.
```
IagoAbal, line 525:
```
- [metavariable regexp](rule-syntax.md#metavariable-regex)
+ [`metavariable-regex`](rule-syntax.md#metavariable-regex)
```

IagoAbal (approved):
I am missing the `function ...` feature that matches both function definitions (stamtents) and lambda expressions. But maybe that goes into a different PR?

ievans (approved):
🙏 ty for updating the docs!

PR #238, opened 2021-08-02
Add links to rules in the registry

inkz:
turn rule ids into links to registry


pabloest (approved):
Oh you did it, so very cool! Thanks @inkz !

PR #239, opened 2021-08-03
Fix snippet for taint example

emjin:
Test plan: run snippet


pabloest (approved):

IagoAbal (commented):
Thanks for updating the example!

PR #244, opened 2021-08-06
Flag deprecated features and deprecated experiments

pabloest:
In this case: equivalences, taint-mode, and pattern-where-python

### DIFF
--- a/docs/experiments/overview.md
+++ b/docs/experiments/overview.md
@@ -73,14 +73,6 @@ rules:
   <img src="https://web-assets.r2c.dev/inline-autofix-regex.gif" width="100%" alt="Apply Semgrep autofix direclty to a file"/>
 </p>
 
-## Equivalences
-
-Equivalences enable defining equivalent code patterns (i.e. a commutative property: `$X + $Y <==> $Y + $X`). Equivalence rules use the `equivalences` top-level key and one `equivalence` key for each equivalence.
-
-For example:
-
-<iframe src="https://semgrep.dev/embed/editor?snippet=jNnn" border="0" frameBorder="0" width="100%" height="435"></iframe>
-
 ## Data-flow analysis
 
 Semgrep can perform intra-procedural flow-sensitive analyses. The data-flow engine still has several limitations, therefore expect both false positives and false negatives. False positives could be removed by using [pattern-not](../writing-rules/rule-syntax.md#pattern-not).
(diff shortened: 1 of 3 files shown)


docs/experiments/overview.md
```diff
@@ -117,3 +97,33 @@ For example:
 ## Generic pattern matching
 
 See [generic pattern matching](../generic-pattern-matching/).
+
+## Deprecated experiments
+
+### Equivalences
+
+:::note
+This feature was deprecated in Semgrep v0.61.0.
```
dlukeomalley, line 106: Is there a reason we can give for this deprecation?

docs/experiments/overview.md
```diff
@@ -117,3 +97,33 @@ For example:
...
+Equivalences enable defining equivalent code patterns (i.e. a commutative property: `$X + $Y <==> $Y + $X`). Equivalence rules use the `equivalences` top-level key and one `equivalence` key for each equivalence.
+
+For example:
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=jNnn" border="0" frameBorder="0" width="100%" height="435"></iframe>
+
+### Taint tracking
+
+:::note
+This feature was deprecated in Semgrep v0.61.0.
```
dlukeomalley, line 118: Same question as above, is there an alternative to now use?

dlukeomalley (approved):

docs/experiments/overview.md
pabloest: I’m not aware. @brendongo any ideas?

docs/experiments/overview.md
pabloest: Our newest taint improvements but they’re not yet documented.

IagoAbal: For the PA team taint-tracking not only is not a deprecated feature but, in fact, it is a priority to push taint-tracking from experimental to beta. Am I parsing this PR the wrong way?

docs/experiments/overview.md
IagoAbal: https://r2c-community.slack.com/archives/CK86BJ5DW/p1628577588061300

I believe there is an issue or something somewhere in Semgrep's GitHub project. But to my knowledge the feature was not being used that much, and the really important equivalences were being made built-in, so it was proposed to deprecate it in order to focus on other things. Besides, now with the `options:` key is much easier to add equivalences because the users have the power to enable/disable them if needed, previously you could not so we had to be more careful with what equivalences we added.

PR #248, opened 2021-08-11
Add docs for join mode

minusworld:


docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
+---
+slug: join-mode
+append_help_link: true
+description: "Join mode runs several Semgrep rules at once and only returns results if certain conditions on the results are met."
+---
+
+# Join Mode
```
pabloest, line 7:
```suggestion
# Join mode
```
(following caps convention of other pages)

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+
+Join mode runs several Semgrep rules at once and only returns results if certain conditions on the results are met. Semgrep is brilliant for finding code patterns with an easy syntax, but its search is typically limited to single files. Join mode is an experimental mode that lets you cross file boundaries, allowing you to write rules for whole code bases instead of individual files. As the name implies, this was inspired by join clauses in SQL queries.
+
+You can think of join mode like this: distinct Semgrep rules are used to gather information about a code base. Then, conditions you define are used to select specific results from these rules, and the selected results are reported by Semgrep in the normal fashion. You can join results on metavariable contents or on the result's file path.
+
+## Example
+
+Here is an example join mode rule that detects a cross-site scripting (XSS) vulnerability with high precision.
```
pabloest, line 15:
```suggestion
Here’s an example join mode rule that detects a cross-site scripting (XSS) vulnerability with high precision.
```

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+    - 'unescaped-extensions.$PATH > template-vars.path'
+  message: |
+    Detected a XSS vulnerability: '$VAR' is rendered
+    unsafely in '$PATH'. 
+  severity: ERROR
+```
+
+Let's explore how this works. First, some background on the vulnerability. Second, we'll walk through the join mode rule.
+
+**Vulnerability Background**
```
pabloest, line 44:
```suggestion
**Vulnerability background**
```

Also, any particular reason to use bold here (and below) instead of a smaller heading? The docs’ search indexing works in part on grabbing hierarchy from the headers.

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+
+In Flask, templates are only HTML-escaped if the [template file ends with the `.html` extension](https://flask.palletsprojects.com/en/2.0.x/templating/#jinja-setup). Therefore, detecting these two conditions present in a Flask application is a high indicator of 
+
+1. User input directly enters a template without the `.html` extension
+2. The user input is directly rendered in the template
+
+**Join Mode Rule Explanation**
```
pabloest, line 51:
```suggestion
**Join mode rule explanation**
```

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+```yaml
+rules:
+- id: any-template-var
+  languages: [generic]
+  severity: INFO
+  message: '$...EXPR'
+  pattern: '{{ $...EXPR }}'
+```
+
+Finally, we want to "join" the results from these together. Below are the join conditions, in English.
```
pabloest, line 94:
```suggestion
Finally, we want to "join" the results from these together. Below are the join conditions, in plain language.
```
?

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+
+We can translate this roughly into the following condition statements.
+
+```
+- 'user-input.$VAR == unescaped-extensions.$VALUE'
+- 'unescaped-extensions.$VAR == template-vars.$VAR'
+- 'unescaped-extensions.$PATH > template-vars.path'
+```
+
+Putting togther the three code pattern Semgrep rules and the three conditions gives us the join rule at the top of this section. This rule will match the code shown here.
```
pabloest, line 108:
```suggestion
Combining the three code pattern Semgrep rules and the three conditions gives us the join rule at the top of this section. This rule will match the code shown here.
```

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+  pattern: '{{ $...EXPR }}'
+```
+
+Finally, we want to "join" the results from these together. Below are the join conditions, in English.
+
+1. The variable `$VAR` from `flask-user-input` has the same content as the value `$VALUE` from `unescaped-template-extension`
+2. The keyword argument `$VAR` from `unescaped-template-extension` has the same content as `$...EXPR` from `any-template-var`
+3. The template file name `$PATH` from `unescaped-template-extension` is a substring of the file path of a result from `any-template-var`
+
+We can translate this roughly into the following condition statements.
```
pabloest, line 100:
```suggestion
We can translate these roughly into the following condition statements.
```

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+running 3 rules...
+ran 3 rules on 16 files: 14 findings
+matching...
+matching done.
+./templates/launch.htm.j2
+severity:error rule:flask-likely-xss: Detected a XSS vulnerability: '$VAR' is rendered unsafely in '$PATH'.
+9:	<li>person_name_full is <b>{{ person_name_full }}</b></li>
+```
+
+**Other Stuff**
```
pabloest, line 126:
```suggestion
**Helpers**
```
?

docs/experiments/join-mode.md
```diff
@@ -0,0 +1,197 @@
...
+
+## Limitations
+
+Join mode **is not taint mode**! While it can look on the surface like join mode is "connecting" things together, it is actually just creating sets for each Semgrep rule and returning all the results that meet the conditions. This means some false positives will occur if unrelated metavariable contents happen to have the same value.
+
+Right now, external rules are required for join mode. (This is why the rules are called `refs` - they are references to other locations.) To use join mode, you must define your individual Semgrep rules in independent locations. This can be anything that works with `semgrep --config <here>`, such as a file, a URL, or a Semgrep registry pointer like `r/java.lang.security.some.rule.id`. 
+
+Currently, join mode will only report the code location of the **last finding that matches the conditions**. Join mode will parse the conditions from top-to-bottom, left-to-right. This means that findings from the "bottom-right" condition will be the reported code location.
+
+## More Ideas
```
pabloest, line 191:
```suggestion
## More ideas
```

pabloest (approved):

docs/experiments/join-mode.md
minusworld: I didn't think it merited its own section because it didn't quite match the "tree structure" I had in my head. If you think it's more beneficial to break this out, though, I can do that!

pabloest: The search will index all the text regardless, it just might rank things differently because of the headings. Since this is v1 of the docs for this feature, it’s probably a-ok as-is.

PR #296, opened 2021-09-28
minor fix

raghavjain3:
### Security

- [ ] Change has security implications (if so, ping the security team)


chmccreery (approved):
So fast!

PR #308, opened 2021-10-08
Change p/ci to auto where applicable, add metrics notice

adamboros:
- Replace p/ci with config=auto. 
- Add metrics notice where we suggest using the registry
- Add p/python example next to auto at /running-rules
- Fix a line on confirm installation which triggered the horizontal scrollbar

### Security


adamboros:
For standard screens this line is too long now, 'recommended Registry rules' can fit instead. 
image

PR #322, opened 2021-10-20
Simplify language maturity stages and add Terraform as beta

pabloest:
### Security


pabloest:
@ievans Thanks!

Yeah I agree about the extensions and tags. Will break them out.

I think I can add some of the green aspect you refer to, not sure how effective it will be. I’ll try it.

For parse rates, maybe just link to the dashboard instead of displaying them all here? Then those who are interested can just click through if they want.

pabloest:
I moved the extensions and tags into an expandable section on the rule syntax page, since `languages` is a required field and that’s where other required fields are described:

Screen Shot 2021-10-20 at 7 13 16 PM

I also added a bit of decoration to the header row of the supported languages table:
Screen Shot 2021-10-20 at 6 18 58 PM

I tried adding more, like this, but IMO it’s unnecessary:

Screen Shot 2021-10-20 at 6 20 05 PM copy

PR #329, opened 2021-10-27
Fix broken md

inkydragon:
image
image

### Security


pabloest:
Thank you so much for this PR, @inkydragon !

pabloest (approved):

pabloest:
I think there are a few other cases of the same issue you found here in other places in this file. See for example line 27, “The ReDoS occurs due to…”

inkydragon:
Looks like I've fixed them.

image
image

PR #334, opened 2021-11-03
Add documentation about embeditor use

minusworld:
Added a section in the FAQ about how to use the embeditor.

@pabloest If there's a better section for this, let me know! We had someone ask about how to do this during a workshop and realized we didn't have docs for this.


minusworld:
Messed up git

PR #341, opened 2021-11-07
Add release notes to docs

adamkvitek:
### Security

- [ ] Change has no security implications (otherwise, ping the security team)


adamkvitek:
@pabloest hi there! This is still just WIP, I need to add the code snippets as they were in the original release notes.

We need to decide about the layout. Maybe we can even add months similarly to VS code release notes. What we definitely need to do is a specific styling where we will not show headers below level 2 in the right panel (I need to research how to do it, if you know, let me know please).

pabloest:
Great, thanks for starting this.

OK, I can look at the header thing you mention.

btw you can also mark this PR as being in Draft mode, that way it’s not even available to approve or merge because GitHub knows it is WIP: https://github.blog/changelog/2020-04-08-convert-pull-request-to-draft/#:~:text=To%20convert%20a%20pull%20request,you%20to%20confirm%20the%20change.

pabloest:
@adamkvitek For hiding particular heading levels, I think it’s straightforward. You need to configure the parameter `toc_max_heading_level` in the YAML frontmatter of the release notes file. For example:

```
toc_max_heading_level: 2
```

See here for the full reference: https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs

pabloest:
Separately, I’m not sure if the release notes should live in the root of the directory structure or nested within the Semgrep folder. I’m open to discussion about it.

adamkvitek:
Hello @pabloest!

Thanks for the reference with CSS settings. ❤️ 

Should I as well figure out months in which we have released particular versions, make these months and years level 2 headers, and downgrade the version headers (currently level 2) to level 3 (and all other headers one level below as well)?

I would keep release notes in the root directory as some information in release notes seems to be related to the whole functionality of Semgrep basically? Do our release notes potentially encompass Semgrep APP and CI as well?

pabloest:
I suggest only looking at the months for 2021, and leaving releases before that to a single 2020 bucket. Yeah it makes sense to “downgrade” the headers in the way you describe.

Currently the release notes only include Semgrep CLI. While this could change in the future, I see your point about the Semgrep (CLI) releases applying to App and CI as well. So I agree, let’s keep them in the root directory for now.

adamkvitek:
@pabloest I included release notes as early as July, should format and include everything from this year? Will be no major problem with that definitely. The only problem in older release notes are code snippets - since I am using the release notes you posted on public Slack, there are no ` ` ` to show code, and I would need to go through all of them manually and find it.

I do not know if we have styling for level 5 headers as well... this is starting to be a real pyramid of headers, so I will probably have to put some filling text in between them stating the obvious ("here are the release notes"). Even level 5 should be slightly bigger than level 4, all of them should be always bigger than the normal paragraph. Maybe @s-santillan, please, can you help me here?

pabloest:
@adamkvitek I wouldn’t worry about going through the 6 months * 4 weeks/month * 1 release per week = 24 release notes to sort through. For now I’d put those into a Jan - July 2021 bucket. I know it’s not super clean but it would work.

adamkvitek:
Thanks! Will do so! :-) The formatting, should I check the Slack and manually reintroduce all the code snippets as they were? I will not be able to copy them and automatically. Or we can just... not have them there. @pabloest

pabloest:
I tried exporting the release-announcements channel messages but Slack gives a large json blob so you’d have to write your own processor to convert from that into Markdown. Not worth it, so I say just continue without them.

adamkvitek:
@pabloest please check the document - I am not sure how to include everything until the beginning of the year. It means that I have to manually copy-paste the release notes from Slack to Google Docs, edit them at least a bit, export them to .docx, convert to markdown and then paste them here. It requires a bit of repetitive work.

pabloest:
Thanks, @adamkvitek. I agree it’s not likely that people would go so far back in time and still not be well served by the GitHub release notes. I checkout out your branch and saw the last header. It looks good to me, let’s ship it like this!

pabloest (approved):

PR #343, opened 2021-11-10
style: Dark mode bg fix in Firefox/Safari

adamboros:
Changing an rgba color to hex so Firefox/Safari properly shows the dark mode background. Also moved this color from the property to the responsible variable and changed the background color to true black for the benefit of all those OLED screens. 

After-before 
image

### Security


adamboros:
closed in favor of #344 where CircleCI build error is fixed

PR #348, opened 2021-11-16
Make diff scan behavior more clear

pabloest:
### Security

### DIFF
--- a/docs/semgrep-ci/overview.md
+++ b/docs/semgrep-ci/overview.md
@@ -62,7 +62,7 @@ For full project scans:
 docker run -v $(pwd):/src --workdir /src returntocorp/semgrep-agent:v1 semgrep-agent --config auto --config <other rule or rulesets>
 ```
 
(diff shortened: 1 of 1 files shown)


raghavjain3 (commented):
Should we just replace "Semgrep will determine the files that have been modified since this reference point and return only newly introduced issues." with "Semgrep will only scan the files modified in your branch and output the difference in findings between the baseline branch and the new branch."?

pabloest:
@raghavjain3 I see what you mean. I just pushed a copy edit, let me know what you think.

raghavjain3:
@pabloest I'd expect info about --baseline-ref to come first and then explain what it does. Also instead of "Semgrep can scan" I think it should be "Semgrep will scan"

pabloest:
Ah yep I see. I like that. Pushing copy edits now.

raghavjain3 (approved):

PR #357, opened 2021-11-30
Recommend diff scans and full scans on separate branches

chmccreery:
Make it clearer what the recommended setup is in terms of when diff-scans are used and when full scans are used.

We have had multiple users configure Semgrep incorrectly, and have poor experiences with the app as a result.

### Security

### DIFF
--- a/docs/semgrep-ci/configuration-reference.md
+++ b/docs/semgrep-ci/configuration-reference.md
@@ -31,7 +31,7 @@ since branching off from your `main` branch, set
 SEMGREP_BASELINE_REF=main
 ```
 
-It is not recommended to do diff-aware scans on any branches that are also doing full-project scans. This is because Semgrep App assumes that all previous findings have been fixed if the branch they are on is re-scanned and they do not show up.
+NOTE: performing a diff scan after a full-project scan of the same branch will close the full-project scan's findings. This is because all findings are marked as fixed if the branch they are on is re-scanned and they are no longer reported. Therefore, it is best to perform diff scans on branches other than your `main` branch.
 
 ## Connect to Semgrep App (`SEMGREP_APP_TOKEN`)
 
--- a/docs/semgrep-ci/configuration-reference.md
+++ b/docs/semgrep-ci/configuration-reference.md
@@ -31,6 +31,8 @@ since branching off from your `main` branch, set
 SEMGREP_BASELINE_REF=main
 ```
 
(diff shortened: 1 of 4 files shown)


docs/semgrep-ci/overview.md
```diff
@@ -11,7 +11,7 @@ import MoreHelp from "/src/components/MoreHelp"
 
 [Semgrep CI](https://github.com/returntocorp/semgrep-action) (aka Semgrep Action or `semgrep-agent`) is a specialized Docker image for running Semgrep in CI environments. It can also optionally connect to [Semgrep App](https://semgrep.dev/manage) for centralized rule and findings management.
 
-- **Scan every commit**. Semgrep CI rapidly scans modified files on pull and merge requests, protecting developer productivity. Longer full project scans are configurable on merges to specific branches.
+- **Scan every commit**. Semgrep CI rapidly scans modified files on pull and merge requests, protecting developer productivity. In most successful configurations, full project scans are reserved for specific branches, such as trunk branches, and diff-aware scans are done on other branches before merging them into the trunk.
```
pabloest, line 14:
```suggestion
- **Scan every commit**. Semgrep CI rapidly scans modified files on pull and merge requests, protecting developer productivity. Usually full project scans are reserved for specific branches, such as trunk branches, and diff-aware scans are done on other branches before merging them into the trunk.
```

docs/semgrep-ci/configuration-reference.md
```diff
@@ -31,6 +31,8 @@ since branching off from your `main` branch, set
 SEMGREP_BASELINE_REF=main
 ```
 
+It is not recommended to do diff-aware scans on any branches that are also doing full-project scans. This is because Semgrep App assumes that all previous findings have been fixed if the branch they are on is re-scanned and they do not show up.
```
pabloest, line 34:
```suggestion
It is not recommended to perform diff-aware scans of branches on which Semgrep App also performs full-project scans. This is because all findings detected during the diff-aware scan will be marked as fixed if Semgrep App doesn’t detect them upon the full-project scan.
```

Trying some copy editing but not sure it’s much better. Do what you like with it!

docs/semgrep-ci/overview.md
chmccreery: I didn't love this either... tried a third option 😄

docs/semgrep-ci/configuration-reference.md
chmccreery: I didn't love this either... tried a third option 😄

s-santillan (approved):

PR #360, opened 2021-12-02
Update docs on .semgrepignore

mmcqd:
Once https://github.com/returntocorp/semgrep/pull/4350 is merged, the Semgrep CLI will support `.semgrepignore` files. This PR removes the warning on the Semgrep CI doc that says the CLI does not support this, and adds a section to the CLI doc that duplicates the explanation in the CI doc. Eventually we plan to remove this functionality from Semgrep CI, and when we do that we can remove the duplication in the docs.

### Security

### DIFF
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -190,6 +190,13 @@ Options:
   --help                          Show this message and exit.
 ```
 
+## Ignoring Files
+
+The Semgrep command line tool supports a `.semgrepignore` file that follows the `.gitignore` syntax and is used to skip files and directories during scanning. This is commonly used to avoid vendored and test related code. For a complete example, see the [.semgrepignore file on Semgrep’s source code](https://github.com/returntocorp/semgrep/blob/develop/.semgrepignore).
+
+If no `.semgrepignore` file is found, no default will be provided.
+
+
 ## Exit codes
 
 `semgrep` may exit with the following exit codes:
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -194,6 +194,19 @@ Options:
 
 The Semgrep command line tool supports a `.semgrepignore` file that follows the `.gitignore` syntax and is used to skip files and directories during scanning. This is commonly used to avoid vendored and test related code. For a complete example, see the [.semgrepignore file on Semgrep’s source code](https://github.com/returntocorp/semgrep/blob/develop/.semgrepignore).
 
+`.semgrepignore` syntax mirrors `.gitignore` syntax, with the following modifications:
+* "Include" patterns (lines starting with "!") are not supported.
+* "Character range" patterns (lines including a collection of characters inside brackets) are not supported.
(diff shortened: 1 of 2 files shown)


docs/semgrep-ci/overview.md
```diff
@@ -241,7 +241,7 @@ If no configuration is provided and no `.semgrep.yml` or `.semgrep/` directory e
 Semgrep CI supports a `.semgrepignore` file that follows the `.gitignore` syntax and is used to skip files and directories during scanning. This is commonly used to avoid vendored and test related code. For a complete example, see the [.semgrepignore file on Semgrep’s source code](https://github.com/returntocorp/semgrep/blob/develop/.semgrepignore).
 
 :::caution
-`.semgrepignore` is only used by Semgrep CI and is not honored by the Semgrep command line tool or by integrations like [GitLab's Semgrep SAST Analyzer](https://gitlab.com/gitlab-org/security-products/analyzers/semgrep).
+`.semgrepignore` is only used by Semgrep CI and the Semgrep command line tool. It is not honored by integrations like [GitLab's Semgrep SAST Analyzer](https://gitlab.com/gitlab-org/security-products/analyzers/semgrep).
 :::
 
 By default Semgrep CI skips files and directories such as `tests/`, `node_modules/`, and `vendor/`. The full list of ignored items is in [the `.semgrepignore` template file](https://github.com/returntocorp/semgrep-action/blob/v1/src/semgrep_agent/templates/.semgrepignore), which is used by Semgrep CI when no explicit `.semgrepignore` file is found in the root of your project.
```
emjin, line 247:
```suggestion
By default Semgrep CI skips files and directories such as `tests/`, `node_modules/`, and `vendor/`. (Note: this is NOT the same as Semgrep CLI, which by default skips nothing). The full list of ignored items is in [the `.semgrepignore` template file](https://github.com/returntocorp/semgrep-action/blob/v1/src/semgrep_agent/templates/.semgrepignore), which is used by Semgrep CI when no explicit `.semgrepignore` file is found in the root of your project.
```

docs/cli-usage.md
```diff
@@ -190,6 +190,13 @@ Options:
   --help                          Show this message and exit.
 ```
 
+## Ignoring Files
+
+The Semgrep command line tool supports a `.semgrepignore` file that follows the `.gitignore` syntax and is used to skip files and directories during scanning. This is commonly used to avoid vendored and test related code. For a complete example, see the [.semgrepignore file on Semgrep’s source code](https://github.com/returntocorp/semgrep/blob/develop/.semgrepignore).
+
+If no `.semgrepignore` file is found, no default will be provided.
+
+
```
emjin, line 199: If we don't have this already, we should add details of what `.gitignore` syntax means.

emjin (approved):

docs/cli-usage.md
mmcqd: Added a section on this

mmcqd: You can resolve this if it looks good to you

PR #361, opened 2021-12-02
Sara/mkt 193 document rbac feature

s-santillan:
### Security


docs/semgrep-app/rbac.md
```diff
@@ -0,0 +1,69 @@
...
+append_help_link: true
+title: Role-based access control
+description: "This page informs users of role-based access control in Semgrep App and teaches users about roles and role management."
+---
+
+import MoreHelp from "/src/components/MoreHelp"
+
+# Controlling access through roles
+
+![Screenshot of x ](../img/rbac-overview.png)<br />
```
pabloest, line 12:
```suggestion
![Screenshot of role-based access control ](../img/rbac-overview.png)<br />
```

docs/semgrep-app/rbac.md
```diff
@@ -0,0 +1,69 @@
...
+
+Access control in Semgrep App determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on Team/Enterprise tiers.
+
+## Setting up RBAC
+
+Semgrep App divides users into two roles:
```
pabloest, line 18: I just thought that “users” may be confused with “developer users” (meaning how we charge for the number of seats in Semgrep App). So maybe rephrase this to something like “Semgrep App divides users who log in into two roles:”

@raghavjain3 Do you think there’s potential for confusion with “developer users” here?

raghavjain3: I think the current wording is fine and implies that it's app users but also open to change to your suggestion of "Semgrep App divides users who log in into two roles" if that is more clear!

raghavjain3 (approved):
Approving but will wait for the +1 from @pabloest as well

pabloest (approved):

PR #362, opened 2021-12-03
migrate from quip resulted in styles not correctly formatted — now fixed

s-santillan:
### Security


pabloest (approved):
Thank you, @s-santillan !

brendongo (approved):

PR #369, opened 2021-12-08
removing extra.css and matching semgrep.dev font-family

adamboros:
- `extra.css` was a leftover from mkdocs 
- Changed the custom defined h5 font-weight to match (rest is defined by docusaurus via Infima (ifm))
- Removed the backup fonts from font-family as we are not using those anymore

### Security


pabloest:
This looks good to me but I know @s-santillan has recently been looking at header weights, so I wanted to make her aware of it.

s-santillan:
The lighter weight LGTM too! Here's a screenshot of them in action (best to use release notes to get an idea).
!All weights](https://user-images.githubusercontent.com/20766840/145456609-bb0ef038-2056-4093-b6dd-d453df5ed973.png)

s-santillan (approved):

adamboros:
I've always forget about the screenshots, sorry 😅

@s-santillan I'm glad you looked at it in more detail, I've just mindlessly corrected it to have the same weight as we do in the App. I've tried to make the [npm package](https://www.npmjs.com/package/@fontsource/inter) working, but couldn't, so in the docs it's uses Google fonts as before. It might handle italic well though cause it can import the variant font which handles that, but it's a blackbox and hard to know when it actually does it.

PR #370, opened 2021-12-10
Sara/mkt 217 clarify that regex matching is

s-santillan:
### Security


nbrahms (approved):
lgtm; thanks Sara!

chmccreery (approved):

PR #383, opened 2021-12-17
Document that we log the rules run by semgrep

emjin:
We now save the rules run by semgrep in the semgrep action logs. Update the docs to reflect this

### Security


s-santillan:
(I am assigning myself to review this. Hope that's okay.)

s-santillan (approved):

PR #400, opened 2022-01-12
Update information about default semgrepignore behavior

mmcqd:
The CLI now uses a default `.semgrepignore` file

### Security

### DIFF
--- a/copy-default-semgrepignore.py
+++ b/copy-default-semgrepignore.py
@@ -1,14 +1,14 @@
 #!/usr/bin/env python3
 
-import requests
+import subprocess
 
 FILE = "docs/cli-usage.md"
-DEFAULT = requests.get("https://raw.githubusercontent.com/returntocorp/semgrep/develop/semgrep/semgrep/templates/.semgrepignore")
+DEFAULT = subprocess.run(["curl","https://raw.githubusercontent.com/returntocorp/semgrep/develop/semgrep/semgrep/templates/.semgrepignore"],capture_output=True).stdout.decode("utf-8")
 
 with open(FILE) as file:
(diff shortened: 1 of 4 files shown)


emjin (approved):

ievans:
Is there a way to reduce the coupling, i.e. fetch it from https://github.com/returntocorp/semgrep when compiling the docs?

mmcqd:
> Is there a way to reduce the coupling, i.e. fetch it from https://github.com/returntocorp/semgrep when compiling the docs?

Not sure. @pabloest do you know if something like this can be done?

pabloest:
We could probably do something similar to what we do for the privacy / metrics text? [This is the PR](https://github.com/returntocorp/semgrep-docs/pull/317). But I’m not sure where the source for the default ignore file resides. Looks like the file we have in the root of the Semgrep project is customized for the project.

ievans:
I think https://github.com/returntocorp/semgrep/blob/develop/semgrep/semgrep/templates/.semgrepignore is the template, so this should be pretty straightforward

pabloest:
@mmcqd I got this error when try to run `yarn start`:

```
yarn start                                                                                                                                   
yarn run v1.22.17
$ python3 copy-default-semgrepignore.py && bash copy-metrics-md && docusaurus start
Traceback (most recent call last):
  File "/Users/pablo/sites/semgrep-docs/copy-default-semgrepignore.py", line 3, in <module>
    import requests
ModuleNotFoundError: No module named 'requests'
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

PR #408, opened 2022-01-18
Add release notes for Semgrep v0.78.0

pabloest:
### Security

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -10,6 +10,33 @@ toc_max_heading_level: 2
 
 Welcome to Semgrep release notes. This document provides an overview of the changes, additions, and fixes made in different versions.
 
+## January 2022
+
+### Version 0.78.0
+
+#### Additions
+
+##### Symbolic propagation
+
+Semgrep can now symbolically propagate simple definitions. For example, given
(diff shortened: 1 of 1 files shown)


s-santillan (commented):
This only occurred to me now: do we need an example for "metavariable-comparison now handles metavariables that bind to arbitrary constant expressions"? If this is domain knowledge such that programmers will know what it implies for rule writing, then I think it's okay not to have an example, but if it's is an operator with entirely Semgrep-made mechanics, it might be worth it to have an example? But this comment should not be treated as a blocker if it's a minor thing.

pabloest:
@s-santillan Hmm yeah, the PR cites this example rule: [https://semgrep.dev/s/By6x](https://semgrep.dev/s/By6x) 

@IagoAbal Do you think it would be helpful to link to that rule? And should we update a few words on the [metavariable-comparison docs](https://semgrep.dev/docs/writing-rules/data-flow/constant-propagation/#metavariable-comparison) text to reflect this change?

pabloest:
@IagoAbal Any thoughts on adding that rule?

@s-santillan Do you think we can merge this even without the additional context?

s-santillan:
@pabloest Let's merge it for now and then add the example later on? I'll approve it now.

s-santillan (approved):

pabloest:
Thanks, @s-santillan!

PR #412, opened 2022-01-21
Add script to dynamically replace text when building

mmcqd:
Several parts of the docs require reference to information that may sometimes (or regularly) change. For instance, the contents of the default `.semgrepignore` file, and the latest version of Semgrep, which appears in the recommended `pre-commit` configuration. 

This PR adds the file `replace-on-build.py`. In this file there is a list called `replacements`, which contains python dictionaries of the form `{"file":FILE, "find":FIND, "replace":REPLACE}`, which indicate that we want to find the text `FIND` in `FILE` and replace it with the text `REPLACE`. This script is run when the docs are built.

This PR also uses the above functionality to solve both previously mentioned issues.

### Security


s-santillan (approved):
My last issue was resolved on my end :) I'll stick to using `yarn start`. @mmcqd Please feel free to squash and merge, thank you :)

PR #438, opened 2022-02-11
Use most-recent privacy doc URL

minusworld:
I noticed the CLI usage was out-of-date after a conversation on Slack. I copied the most recent CLI help text here.

### DIFF
--- a/docs/cli-usage.md
+++ b/docs/cli-usage.md
@@ -17,177 +17,192 @@ See `semgrep --help` for command line options.
 ```sh
 Usage: semgrep [OPTIONS] [TARGET]...
 
-  Semgrep CLI. Searches TARGET paths for matches to rules or patterns. Defaults to
-  searching entire current working directory.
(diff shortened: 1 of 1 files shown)


s-santillan:
I am not sure what our convention is with hard/soft wrapping in pre/code blocks. The PR itself is an addition explaining what data we collect with config auto (thank you @minusworld) but it looks like the whole section has increased margins. I am attaching both screenshots for comparison.

Patch:
minusworld-patch

Main:
main-branch

pinging @pabloest for the solution.

pabloest:
I believe these increased margins aren’t margins in the CSS sense, but rather they are visible because the source contents (the CLI reference) simply has line breaks in different areas than before.

s-santillan:
Oh! Yeah, I meant that the hard line break from the patch is what did it. Should we stick to the old line break (greater width = less scrolling) or are we okay with Grayson's new ones? @pabloest

s-santillan (approved):

PR #445, opened 2022-02-18
Fix a few broken links and remove analytics

adamboros:
### Security


s-santillan:
@adamboros Thank you for all these catches! I don't know what implications there are upon removing analytics, could you explain it to me please? CC @pabloest in case you have an opinion, Pablo? Thank you both!

adamboros:
Oh sorry, missed this totally. We deleted the analytics page in favor of the new dashboard where these screenshots originate from and right now users can't access these information anywhere in the App.

adamkvitek (approved):

PR #107, opened 2021-02-10
Move cli-reference to cli-usage

brendongo:

PR #450, opened 2022-02-18
Rewrite minor issues in CLI docs

adamkvitek:
Linear: https://linear.app/r2c/issue/MKT-302/check-cli-docs-make-first-minor-edits-to-familiarize-with-the-current

### Security


adamkvitek:
This comment will include notes about the current state of CLI documentation:

- It seems that Getting started shows first some procedures how to actually run Semgrep, but then it diverges to a collection of links. If this article tells me something about rules and just lists what these rules may possibly help me with, it is not documentation but a marketing material. - Solution - consider deleting or moving the content which is not directly showing actions user should do. Make this getting started step by step - by including how to create first rules.

docs/getting-started.md
```diff
@@ -1,54 +1,49 @@
 ---
 slug: getting-started
 append_help_link: true
-description: "Start by running Semgrep locally to scan your code. It runs offline on uncompiled code: no code leaves your machine."
+description: "This article provides basic informatin on how to start using Semgrep CLI."
```
ievans, line 4:
```suggestion
description: "How to start using Semgrep CLI."
```

docs/cli-reference.md
```diff
@@ -1,5 +1,5 @@
 ---
-slug: cli-usage
+slug: cli-reference
```
pabloest, line 2: What’s the reasoning to change the slug? I didn’t see a discussion about that in the PR conversation.

Note there is also this history: https://github.com/returntocorp/semgrep-docs/pull/107#discussion_r575481763

docs/getting-started.md
```diff
@@ -1,54 +1,49 @@
 ---
 slug: getting-started
 append_help_link: true
-description: "Start by running Semgrep locally to scan your code. It runs offline on uncompiled code: no code leaves your machine."
+description: "How to start using Semgrep CLI."
```
pabloest, line 4: What’s the reasoning behind this change? My sense is the previous description added some more details beyond what’s directly in or implied by the URL and title themselves. This description seems mostly reflexive of the title.

docs/cli-reference.md
adamkvitek: I am just trying to be consistent - the file is called cli-usage, but the first header and left navigation entry is CLI reference. @pabloest I changed all the links leading to this file as well.

docs/getting-started.md
adamkvitek: I am not sure that the previous text is an actual description, it is the first sentence of the article actually... I can expand the description, I just do not like to see the first sentence of the article reused as a description. @pabloest

docs/cli-reference.md
adamkvitek: It just seems weird when the main title is CLI reference and the URL states cli-usage.

docs/getting-started.md
pabloest: Fair enough! Then I suggest something different. Otherwise it’s a swap of one duplicate thing (the first sentence of the article) with another duplicate thing (the page title).

docs/cli-reference.md
pabloest: Sounds great. Thanks!

docs/getting-started.md
adamkvitek: @pabloest ok, pushed this, check this description out:

`description: "This getting started guides you through the installation of Semgrep, shows you how to run Semgrep locally, and gives an overview of benefits which you may reap by writing Semgrep rules and by using Semgrep CI."`

docs/getting-started.md
```diff
@@ -1,7 +1,7 @@
 ---
 slug: getting-started
 append_help_link: true
-description: "This getting started guides you through the installation of Semgrep, shows you how to run Semgrep locally, and gives an overview of benefits which you may reap by writing Semgrep rules and by using Semgrep CI."
+description: "This getting started guides you through the installation of Semgrep, shows you how to run Semgrep locally, and provides an overview of benefits which you may reap by writing Semgrep rules and by using Semgrep CI."
```
pabloest, line 4: Thank you for this!

pabloest (approved):

PR #464, opened 2022-03-07
Add docs for metavariable-analysis experiment

kurt-r2c:

### Please ensure


ievans (approved):
Love it!

mjambon:
Thanks, @kurt-r2c, this is great. Hopefully, I'll get around to adding some details about ReDoS predictions (ideally, we would provide input that causes slow matching but it's tricky because not all regexp engines are vulnerable to the same patterns).

PR #467, opened 2022-03-08
Capitalize ScanMyCode

dlukeomalley:
I'm updating the capitalization of Scanmycode to ScanMyCode after receiving feedback that `scanmy` was mistaken for `scammy`. This doesn't match the capitalization used by the underlying project, but I think is a valuable change given that these are Semgrep-hosted docs for our user base.

@marcinguy, are you comfortable with this change based on concerns on how it reads?


marcinguy:
@dlukeomalley Yes, please change it. There was a lot of work spent on it and it would be an unfortunate association.

pabloest (approved):

PR #478, opened 2022-03-24
Delete project-depends-on-docs

mmcqd:
As per discussion on slack, this removes the public facing documentation of `project-depends-on`

### Please ensure


adamkvitek:
Are we deleting this because we have a new functionality for the project-depends-on-docs which needs to be documented? Just asking for the context to understand this. @pabloest @mmcqd

PR #480, opened 2022-03-28
Review recursive join 

adamkvitek:

I just noticed a great PR with the recursive join mode merged, checked, and had some stylistic nits, please check it out.

### Please ensure

### DIFF
--- a/docs/experiments/join-mode/recursive-joins.md
+++ b/docs/experiments/join-mode/recursive-joins.md
@@ -2,7 +2,7 @@
 
 Join mode is an extension of Semgrep that runs multiple rules at once and only returns results if certain conditions are met. Join mode is an experimental mode that enables you to cross file boundaries, allowing you to write rules for whole codebases instead of individual files. More information is available in [Join mode overview](/docs/experiments/join-mode/overview).
 
-Recursive join mode has a recursive operator, `-->`, which executes a recursive query on the given condition. This recursive operator allows you to write a Semgrep rule that effectively crawls the codebase on some condition, letting you build chains such as function call chains or class inheritance chains.
+Recursive join mode has a recursive operator, `-->`, which executes a recursive query on the given condition. This recursive operator allows you to write a Semgrep rule that effectively crawls the codebase on condition you specify, letting you build chains such as function call chains or class inheritance chains.
 
 ## How recursve join mode works
 
--- a/docs/experiments/join-mode/recursive-joins.md
+++ b/docs/experiments/join-mode/recursive-joins.md
@@ -2,7 +2,7 @@
 
 Join mode is an extension of Semgrep that runs multiple rules at once and only returns results if certain conditions are met. Join mode is an experimental mode that enables you to cross file boundaries, allowing you to write rules for whole codebases instead of individual files. More information is available in [Join mode overview](/docs/experiments/join-mode/overview).
 
-Recursive join mode has a recursive operator, `-->`, which executes a recursive query on the given condition. This recursive operator allows you to write a Semgrep rule that effectively walks the codebase on some condition, letting you build chains such as function call chains or class inheritance chains.
+Recursive join mode has a recursive operator, `-->`, which executes a recursive query on the given condition. This recursive operator allows you to write a Semgrep rule that effectively crawls the codebase on some condition, letting you build chains such as function call chains or class inheritance chains.
 
 ## How recursve join mode works
 
--- a/docs/experiments/join-mode/recursive-joins.md
+++ b/docs/experiments/join-mode/recursive-joins.md
@@ -50,7 +50,7 @@ rules:
       $CALLEE(...)
 ```
 
-A join condition such as the following: `python-callgraph.$CALLER --> python-callgraph.$CALLEE` Produces a table below. Notice how `function_1` appears with `function_4` and `function_5` as callees, even though it is not directly called.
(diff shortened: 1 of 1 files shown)


docs/experiments/join-mode/recursive-joins.md
```diff
@@ -1,12 +1,12 @@
 # Recursive joins
 
-Join mode is an extension of Semgrep that runs multiple rules at once and only returns results if certain conditions are met. Join mode is an experimental mode that lets you cross file boundaries, allowing you to write rules for whole code bases instead of individual files. More information is available in [these docs](/docs/experiments/join-mode/overview).
+Join mode is an extension of Semgrep that runs multiple rules at once and only returns results if certain conditions are met. Join mode is an experimental mode that enables you to cross file boundaries, allowing you to write rules for whole codebases instead of individual files. More information is available in [Join mode overview](/docs/experiments/join-mode/overview).
 
-Recursive join mode adds a new recursive operator, `-->`, which executes a recursive query on the given condition. This lets you write a Semgrep rule that effectively crawls the codebase on some condition, letting you build chains such as function call chains or class inheritance chains.
```
adamkvitek, line 5: By crawl the codebase - do we mean to walk the codebase (recursively)? https://docs.pyfilesystem.org/en/latest/walking.html#walking

minusworld: 👋  I used "effectively crawls" purposefully, since recursive join doesn't actually walk the codebase but it _simulates_ walking the codebase.

It uses a Semgrep pattern just like every other Semgrep rule, parsing individual files into ASTs and matching the given Semgrep patterns on that AST. Recursive join simulates recursively walking the codebase by saving those results into a database table and recursively joining the tables together until a terminal condition is reached.

Not sure if this makes sense.... 😅  Let me know if this is helpful, or if I've only served to create more questions 🤣

adamkvitek: @minusworld I am very much interested in how Semgrep works in the background with AST, maybe we can make a call for noobs like me and record it? I mean... I need to properly know what is AST, how it works... if we already have recordings like that it would be immensely helpful!

Just for my understanding, crawl the file system or codebase is a regular term? Crawling the codebase means that Semgrep goes through the codebase recursively, but does not really edit or change anything (is that what you mean by simulating walking the codebase) while walking the codebase would mean that we edit something in it? 
Screenshot 2022-03-29 at 13 54 44

docs/experiments/join-mode/recursive-joins.md
```diff
@@ -50,7 +50,7 @@ rules:
       $CALLEE(...)
 ```
 
-A join condition like this, `python-callgraph.$CALLER --> python-callgraph.$CALLEE`, will produce a table like this. Notice how `function_1` appears with `function_4` and `function_5` as callees, even though it isn't directly called.
+A join condition such as the following: `python-callgraph.$CALLER --> python-callgraph.$CALLEE` Produces a table below. Notice how `function_1` appears with `function_4` and `function_5` as callees, even though it is not directly called.
```
pabloest, line 53:
```suggestion
A join condition such as the following: `python-callgraph.$CALLER --> python-callgraph.$CALLEE` produces a table below. Notice how `function_1` appears with `function_4` and `function_5` as callees, even though it is not directly called.
```

pabloest (approved):

docs/experiments/join-mode/recursive-joins.md
minusworld: >  I am very much interested in how Semgrep works in the background with AST, maybe we can make a call for noobs like me and record it? I mean... I need to properly know what is AST, how it works... if we already have recordings like that it would be immensely helpful

@adamkvitek Oh, sure thing! Happy to talk over video. Just schedule a time on my calendar and I'll be there! :)

> Just for my understanding, crawl the file system or codebase is a regular term? Crawling the codebase means that Semgrep goes through the codebase recursively, but does not really edit or change anything (is that what you mean by simulating walking the codebase) while walking the codebase would mean that we edit something in it?

Walking and crawling share very similar meanings in this case. For illustration purposes, let's say there's a file system that looks like this:

```
root
|- Downloads
|-- GitHub
|--- code.py
|- Documents
|-- Videos
|--- cats.mp4
|-- Pictures
|--- cats.jpg
```

"Walking" the filesystem from `root` would mean retrieving all the directory, subdirectory, and filenames underneath `root`. Some Python code to walk the filesystem looks like this:

```
for root, dirs, files in os.walk("root"):
  for file in files:
    print(file)
```

The term "crawling" is borrowed from something called "web crawlers". Web crawlers, or spiders as they are sometimes called, are how search engines like Google find and archive content for their search engine. A web crawler will continuously follow hyperlinks, downloading and archiving the web page that the hyperlink points to.

Notably, both walking and crawling can be recursive algorithms, hence their similarity to recursive join mode.

Both walking and crawling, to me, convey a sense that something is continually moving on to the next thing. So, when I say "effectively crawl the codebase", I am trying to communicate that recursive joins will continually follow sections of the codebase kind of like web crawlers follow hyperlinks. You could also say "effectively walk the codebase" which would communicate that recursive joins will drill down further and further into the codebase, like when you "walk" the filesystem.

Hope this is clear as mud. :)

docs/experiments/join-mode/recursive-joins.md
adamkvitek: Oh thank you!

docs/experiments/join-mode/recursive-joins.md
adamkvitek: @minusworld thanks a lot! I am saving your explanation. I think I understood the main point. :-) Maybe we can discuss it more later on a short call. But thanks for this input!

PR #496, opened 2022-04-08
Sync CLI help with 0.87.0

underyx:

### Please ensure


adamkvitek (approved):

adamkvitek:
@underyx just noticed a bit strange thing - when I try `semgrep --help` I see:
Screenshot 2022-04-08 at 11 19 10
 
I have to use `semgrep ci --help` to see more. But we should print the whole man page I guess. I hope that this is not expected behavior... :D

underyx:
Woop I indeed synced an old version. Updated now, it's actually intended that the basic help doesn't tell you all the scan flags.

PR #500, opened 2022-04-12
Add release notes for version 0.87.0

adamkvitek:

### Please ensure

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -12,19 +12,19 @@ Welcome to Semgrep release notes. This document provides an overview of the chan
 
 ## April 2022
 
-Version 0.87.0
+### Version 0.87.0
 
-### Additions
+#### Additions
 
-- A new `focus-metavariable` operator that enables you to focus (or zoom in) the match on the code region delimited by a metavariable. This operator is useful for narrowing down the code matched by a rule, to focus on what really matters. For more information, see [focus-metavariable documentation](https://semgrep.dev/docs/experiments/focus-metavariable/). ([#4453](https://github.com/returntocorp/semgrep/issues/4453))
+- A new `focus-metavariable` operator that enables you to focus (or zoom in) the match on the code region delimited by a metavariable. This operator is useful for narrowing down the code matched by a rule, to focus on what really matters. For more information, see [focus-metavariable documentation](/experiments/focus-metavariable/). ([#4453](https://github.com/returntocorp/semgrep/issues/4453))
 
 - Semgrep CI uses `GITHUB_SERVER_URL` to generate URLs if it is available.
 
 - You can now set `NO_COLOR=1` to force-disable colored output.
 
-### Changes
+#### Changes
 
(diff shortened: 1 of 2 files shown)


docs/release-notes.md
```diff
@@ -10,6 +10,28 @@ toc_max_heading_level: 2
...
+
+- The `returntocorp/semgrep` Docker image no longer sets `semgrep` as the entrypoint. This means that Semgrep is no longer prepended automatically to any command you run in the image. This makes it possible to use the image in CI executors that run provisioning commands within the image.
```
s-santillan, line 29: Is it possible that people who use the Docker image _now_ will experience breaking behavior?

docs/release-notes.md
```diff
@@ -10,6 +10,28 @@ toc_max_heading_level: 2
...
 
+## April 2022
+
+### Version 0.87.0
+
+#### Additions
+
+- A new `focus-metavariable` operator that enables you to focus (or zoom in) the match on the code region delimited by a metavariable. This operator is useful for narrowing down the code matched by a rule, to focus on what matters. For more information, see [focus-metavariable documentation](/experiments/focus-metavariable/). ([#4453](https://github.com/returntocorp/semgrep/issues/4453))
+
+- Semgrep CI uses `GITHUB_SERVER_URL` to generate URLs if it is available.
```
s-santillan, line 21: What is this used for?

adamkvitek: Good question. Please, @underyx, can you tell us more about this?

s-santillan:
@adamkvitek I'm not sure if I'm extra vigilant for this release note, but I just want to understand two changes. Thank you!

docs/release-notes.md
adamkvitek: @underyx or @brendongo, can this cause breaking behavior?

docs/release-notes.md
adamkvitek: CC @emjin

docs/release-notes.md
brendongo: Nope it wont immediately cause breaking behavior since we try to detect if they are using it the old way and keep that behavior. It does print a deprecation notice

docs/release-notes.md
brendongo: It is for github enterprise users

brendongo: since their base url is not github.com but usually something like github.company.com or something

docs/release-notes.md
s-santillan: We should mention the deprecation of the old way, I think. Such as "Affected users may receive a deprecation notice. Adjust scripts accordingly"? @adamkvitek

docs/release-notes.md
s-santillan: This is a passive buff (improvement) for Enterprise then. :) (For Adam: There's no action the user needs to undertake, they are describing a change in behavior.) I wonder how might we convey enterprise vs community updates. A question for another time.

adamkvitek: Maybe we can just state that this is for GitHub Enterprise users then? @s-santillan 

Suggesting:

```
- For GitHub Enterprise users: Semgrep CI uses `GITHUB_SERVER_URL` to generate URLs if it is available.
```

docs/release-notes.md
adamkvitek: I agree with @s-santillan.

```
- The `returntocorp/semgrep` Docker image no longer sets `semgrep` as the entry point. This means that Semgrep is no longer prepended automatically to any command you run in the image. This makes it possible to use the image in CI executors that run provisioning commands within the image. Affected users may receive a deprecation notice. Adjust scripts accordingly.
```

We should specify what the user needs to do. How should they change their configuration? Does the deprecation notice instruct them on what to do? @brendongo

docs/release-notes.md
s-santillan: Yeah, let's do that 👍

adamkvitek: Thanks! Resolving.

s-santillan (approved):
The additions look good to me 👍 Thank you.

docs/release-notes.md
brendongo: Yes it does. They just need to add a word to their invocation of the docker command

PR #510, opened 2022-04-18
Sara/mkt 143 clarify and simplify getting started

s-santillan:

### Please ensure


docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
...
+
+
+#### Permissions for GitLab
+
+Semgrep requires the following permissions (scopes) in order to function:
+
+
+<dl>
+    <dt>`api`</dt>
+    <dd>Legacy usage enables Semgrep App to retrieve repositories from GitLab.</dd>
```
s-santillan, line 96: This is the last item to make sure it's absolutely correct. I was informed that we don't use api scope for what we used to use it for (i.e. ListProjects API) but still request permissions? For that reason I used "legacy" but not sure as to the accuracy of this claim. @raghavjain3 what do you think? cc @zyannes

raghavjain3: Should we just remove this permission and exclude it from the docs then?

zyannes: I think we can remove this permission for now since we're not using it for Gitlab. However, we do have plans to implement this feature for Gitlab. That said, I think we could eventually support an opt-in permissions model, where customers who grant the `api` scope have access to the features requiring that permission, whereas customers who do not will have a limited feature set (more manual steps for which we can provide walkthroughs/documentation). What do you think?

s-santillan: Agree that we can remove the permission for now (and I'll remove it from the docs) . 
When we implement it in the future for GitLab, we can add it back.

I agree with the opt-in model as well, but will leave it to prod/eng to be the final decision owner. Cheers y'all. Thanks!

raghavjain3: @zyannes that sounds good! Would users be able to select individual permissions? I think with GitHub that isn't possible so wondering what it's like for Gitlab. 

Also, could we confirm with @dlukeomalley here as the PM for integrations?

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
+---
+slug: getting-started-with-semgrep-app
+append_help_link: true
+title: Getting started with Semgrep App
+description: "Get started with Semgrep App to scan for security vulnerabilities on cloud repositories hosted on GitHub and GitLab."
+---
+
+# Scanning cloud repositories with Semgrep App
+
+Semgrep App enables you to run scans on multiple repositories by integrating with your GitHub or GitLab-hosted account. Scans are powered by **rules**. A Semgrep rule encapsulates pattern-matching logic and data-flow analysis used to find code violations, security issues, outdated libraries, and other issues.
```
dlukeomalley, line 10: "powered" in this instance strikes me as interesting. I do think rules are the *fuel* for the Semgrep engine, but not necessary what "powers" a scan.

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
...
+
+# Scanning cloud repositories with Semgrep App
+
+Semgrep App enables you to run scans on multiple repositories by integrating with your GitHub or GitLab-hosted account. Scans are powered by **rules**. A Semgrep rule encapsulates pattern-matching logic and data-flow analysis used to find code violations, security issues, outdated libraries, and other issues.
+
+
+![Diagram of Semgrep App flow](../img/semgrep-app-diagram.png "Diagram of Semgrep App flow")
+
+
+Many rules are available from [Semgrep Registry](https://semgrep.dev/r), an open-source, community-driven repository of rules. Write your own rules to create general security checks through the Registry while customizing Semgrep to work for your team's specific secret-keeping and API practices.
```
dlukeomalley, line 16: The power of rules when individuals write them is that they aren't general, they're highly specific! I wonder how we can capture for folks something like: you can write your own rules (n.b. you don't do this through the Registry, as I perceive it) either for the entire community, or to tailor Semgrep for your team's specific practices (the ones that no "generic" rule written by someone else could ever catch). For the former, this is a huge selling point in the customer discussions I'm in.

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
...
+
+After performing a scan, use Semgrep App's Rule Board feature to prevent pull requests (PRs) or merge requests (MRs) from merging until findings are resolved. This helps prevent vulnerable code from shipping to widely-accessible environments, such as production or staging servers.
```
dlukeomalley, line 18: Is the Rule Board important _before_ a scan? Specifically, you interact with the Rule Board before a scan to configure the behavior of future scans. The language here implies to me you scan, then decide, but in practice the workflow is the opposite.

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
...
+
+Semgrep App enables you to scan whole repositories on a cloud. You can try Semgrep CLI for local scans as well. For more information, see [Getting started with Semgrep CLI](https://semgrep.dev/docs/getting-started/).
```
dlukeomalley, line 20: What does it mean to "scan whole repositories on a cloud"? I could take this to mean code is uploaded to Semgrep App, although the code never leaves a user's environment. Semgrep App enables you to deploy, configure, and manage Semgrep running in your CI environment. Would you be open to language more like this?

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
...
+
+
+## Signing into Semgrep App
+
+Signing into Semgrep App requires either a GitHub or hosted GitLab account. Semgrep App also supports Single Sign-On (SSO) on Team or Enterprise tiers. This guide focuses on GitHub and GitLab sign-ins. See [SSO Configuration](https://semgrep.dev/docs/semgrep-app/sso/) for information on single sign-on.
```
dlukeomalley, line 25: What is the signficance of stating "hosted GitLab account"?

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,284 @@
...
+3. Sign in with your credentials.
+4. Click **Authorize semgrep-app**. See the GitHub documentation about [Authorizing GitHub Apps](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/authorizing-github-apps) to understand the scope of permissions requested by Semgrep.
+5. You are redirected back to Semgrep App.
+6. Click **Accept** to accept Semgrep's Terms of Service.
+7. Optional: Fill out the survey and click Complete or click Skip to omit this step.
+
+You are now signed in to Semgrep App.
+
+
+#### Permissions for GitHub
```
dlukeomalley, line 48: As these permissions are subject to change over time, is it necessary to document them (and the sign-in) process? I wonder if there is feedback on the onboarding flow (CC @juliat @milanwilliams @adamboros) that if we addressed would deprecate the need for documentation here. Thoughts on how to balance the product being the documentation vs. having explicit external docs?

docs/semgrep-app/getting-started-with-semgrep-app.md
dlukeomalley: We know these permissions are overly permissive. As part of onboarding (CC @milanwilliams @TinMarkovic) we can consider reducing scope (it's a customer ask) and these docs could be updated at that time.

dlukeomalley (commented):
@s-santillan Did a quick pass on the onboarding portions because I was tagged. My main questions are around the scope of documentation and what to, and not to, document here vs. within the product itself. I'm worried about Q2 onboarding changes and how they could change these docs quite significantly. Happy to chat, perhaps you, @milanwilliams, and I jump into a room as helpful!

Thank you for writing these! Helpful to see it written out and grok how complex the flow is today.

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: Adam actually gave the same note! I will fix this.

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: (1) There was an external engineer that asked "why do we need api scope?"
(2) I referenced @raghavjain3 's document that was shared with some users.
(And those two were what prompted me to raise the priority on this task.)

I would prefer that the product be the documentation, but there was clearly a need.

Looking forward to more input from prod/design!

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: This is a nice catch. I should be referring to GitLab as GitLab SaaS at first mention then GitLab moving forward except where precision is absolutely necessary. It has been fixed. Thank you.

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: My assumptions about user flow went as follows:

(1) A complete Semgrep newbie wants to dive in and try out app (no prior experience with CLI)
(2) "I want to test Semgrep App" -> user adds a project
(3) Once project added, a scan **automatically** starts using the default ruleset (default behavior of the app.) A person entirely new to Semgrep and dipping their toes through app would not necessarily know Semgrep's "make rules then scan" flow.
(4) "I got good results, can I refine it further?" -> looks up rules (writing rules or rule board) OR
(5) "I didn't get great results", why? -> looks up rules (writing rules or rule board.)

I think my language is unclear here because I say "After performing a scan" - I should say, "After trying out Semgrep's default ruleset on some scans" or something similar.

What do you think?

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: Have added this in!

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: Resolved by removing "powered" and simplifying the wording in the interests of accessibility and clarity. In hindsight, using metaphors should be avoided. Thanks.

docs/semgrep-app/getting-started-with-semgrep-app.md
TinMarkovic: Yup. It's a thing we'd love to do... but don't have now, and I'm not sure we'll have in a relevant timeframe (for these docs). I think it's fine to leave it in writing or remove it entirely, but don't hold up the docs waiting for major changes.

s-santillan: Thank you @TinMarkovic! 

@dlukeomalley Just making sure I got this right:

(1) My action for this doc right now: Remove mention of api in these docs (done).
(2) @zyannes Will remove permission requirement(?)
(3) Tin/Milan/Zach will eventually return this permission with support for opt-in permissions. Docs will be updated to re-include this permission when it happens in the future.

docs/semgrep-app/getting-started-with-semgrep-app.md
adamboros: I think if we are changing the onboarding flow, it's going to be a bigger change and we need to update the docs anyways. Also I do find it useful to have this explained as it can give clarity to users who look at the docs first and sign in later.

s-santillan: (Resolving - no actionable item. Thank you.)

s-santillan:
@adamkvitek I committed all of your review suggestions into this branch, could you kindly check and approve (i.e. made Prerequisites a subheader, "sign in to", etc.)?

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,285 @@
...
+
+To change these scan parameters, either:
+
+* Edit the `semgrep.yml` file manually.
+
+* Remove the project and redo the steps described in Adding a project section.
+
+Additional scan parameters include:
+
+**Rule recommendation**
```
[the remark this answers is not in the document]
adamkvitek: This seems like a "dummy" description list. :D

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,285 @@
...
+    Select this toggle to receive rule recommendations in the Rule Board based on the framework and language of the repository.
```
adamkvitek, line 165:
```suggestion
    <dd>Select this toggle to receive rule recommendations in the Rule Board based on the framework and language of the repository.</dd>
```

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Autofix**
+    Select this toggle to enable autofix, which creates suggestions in addition to PR or MR comments. For example, a rule may suggest using a function such as `logging.debug()` instead of `print()`.
```
adamkvitek, line 168:
```suggestion
    <dd>Select this toggle to enable autofix, which creates suggestions in addition to PR or MR comments. For example, a rule may suggest using a function such as <code>logging.debug()</code> instead of <code>print()</code>.</dd>
```

adamkvitek: I hope that the `<code>something</code>` will be rendered the same as with markdown ``...

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,285 @@
...
+
+**Path ignores**
+	Paths and files specified here are not scanned by Semgrep App.
```
adamkvitek, line 171:
```suggestion
	<dd>Paths and files specified here are not scanned by Semgrep App.</dd>
```

docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -0,0 +1,285 @@
...
+Both the Dashboard and the Findings page display the results of a scan. These pages are accessible from the left sidebar in Semgrep App. The **[Dashboard](https://semgrep.dev/docs/semgrep-app/dashboard/)** is a report view to assist in evaluating security posture across repositories. It organizes findings into OWASP categories, enabling users to assess habits and trends within their team or organization.
+
+The **[Findings](https://semgrep.dev/docs/semgrep-app/findings/#managing-triage-states-bulk-triage)** page is used to triage findings. Triaging refers to prioritizing a finding based on criteria set by your team or organization. While severity is a factor in triage, your organization may define additional criteria based on coding standards, business, or product goals.
+
+To see the rule specifics that triggered the finding, click on the rule entry.
+
+
+### Automatically resolving findings
+
+
```
adamkvitek, line 236: @s-santillan we have a lot of additional empty lines in this whole document, if you want, you can delete them... just to make the source code cleaner.

Generally, I think that Markdown builders ignore more than one empty line, so there is no difference after build. It just does not look that clean, but it is the tinniest thing here.

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: (Marking as resolved as there are no definite action items. Willl discuss with onboarding team these "assumptions".)

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: @adamkvitek I'll test

docs/semgrep-app/getting-started-with-semgrep-app.md
s-santillan: I'll pare down some extra whitespace

s-santillan:
@adamkvitek Added all suggestions and resolved all other issues. Docs build and work. Thank you!

PR #524, opened 2022-04-29
Update Yarn packages

adamkvitek:

### Please ensure


adamkvitek:
@pabloest do you get any errors now when you try to build the docs after pulling from this branch? 🙏

pabloest:
@adamkvitek I don’t get any Yarn errors any more but I still get the error about the `cli-usage.md` file not found. It’s not related to Yarn but it’d be nice to fix both.

adamkvitek:
@pabloest I hope that removing `cli-usage.md` from `replace-on-build.py` should help here.

I hope that this might really help. I worry that I tried this upgrade as eye candy ("Woow, this UI now looks a bit nicer!"), but I'm unaware of possible consequences. I removed package-lock.json as instructed by yarn start as well.

Do you think we should install these dependencies as Install Deps in `ci/circleci: test` suggests?

```
warning "@docusaurus/core > react-dev-utils > fork-ts-checker-webpack-plugin@6.5.2" has unmet peer dependency "typescript@>= 2.7".
warning "@docusaurus/preset-classic > @docusaurus/theme-search-algolia > @docsearch/react@3.0.0" has unmet peer dependency "@types/react@>= 16.8.0 < 18.0.0".
warning "@docusaurus/preset-classic > @docusaurus/theme-search-algolia > @docsearch/react > @algolia/autocomplete-preset-algolia@1.5.2" has unmet peer dependency "@algolia/client-search@^4.9.1".
```

This might not be connected to this PR, but I am seeing this for months during build, not sure how to solve it:

```
[WARNING] Duplicate routes found!
- Attempting to create page at /docs/metrics/, but a page already exists at this route.
This could lead to non-deterministic routing behavior
```

We have some broken links during Yarn Build stage as well, but they seem to be resolved ok?

Just feels like a nice clean-up here, maybe we can prevent some issues here by making our pipelines run without warnings I hope.

pabloest:
@adamkvitek Thanks for working on this. 

For the reference to the file `cli-usage.md`, I believe [this commit](https://github.com/returntocorp/semgrep-docs/pull/450/commits/f4dafac08f298194569eab6e8fbf73913ce149eb) you made renamed the file from `cli-usage.md` to `cli-reference.md`. This then caused the replace-on-build step to fail, because it expects to find a file named `cli-usage.md`. The solution should be to update the replace-on-build step by updating the file name, not by entirely removing the thing to be replaced as done in [this commit](https://github.com/returntocorp/semgrep-docs/pull/524/commits/c5d93791b3b848ec88250c8904e65c76737d0e71). Let me know if that’s not clear.

I wouldn’t worry about the CircleCI warnings there.

The duplicate routes warning is known and expected, and it’s due to the way we update the contents of the metrics file (by replacing contents in a “placeholder” file).

adamkvitek:
@pabloest Oh, yes, thank you! Sorry for this mistake and thank you for your patience.
 So I should include the following line to `replace-on-build.py`?
```
{"file":"docs/cli-reference.md", "find":"DEFAULT_SEMGREPIGNORE_TEXT", "replace":DEFAULT_SEMGREPIGNORE},
```

Just asking to learn something:

I see that `DEFAULT_SEMGREPIGNORE` points to: https://raw.githubusercontent.com/returntocorp/semgrep/develop/semgrep/semgrep/templates/.semgrepignore

What was `DEFAULT_SEMGREPIGNORE_TEXT`?

pabloest:
@adamkvitek Yes, exactly.

`DEFAULT_SEMGREPIGNORE_TEXT` is just a placeholder that the replace-on-build step can easily find, and then replace with the contents that you linked to from that `.semgrepignore` file. The reason to do this is so that the contents of the defaults `.semgrepignore` file (which is distributed as part of Semgrep’s templates) is always accurately reflected here in the docs, and we then don’t have to worry about updating contents in two places when there is a change to the template (instead we can update in just one place).

adamkvitek:
This is an awesome solution actually... it would be great to do the same for our Semgrep help overview, but the source file is not so easily understandable. @pabloest

adamkvitek:
This is an awesome solution actually... it would be great to do the same for our Semgrep help overview, but the source file is not so easily understandable. @pabloest

adamkvitek:
@pabloest let me know if there are no other blockers for this PR. 🙏

pabloest:
@adamkvitek Yes, this looks good to me now!

pabloest (approved):

PR #544, opened 2022-05-18
fix wrong formatting of dl dt tags

s-santillan:

Fixes!


s-santillan:
Should look like this

:

pabloest (approved):

PR #547, opened 2022-05-20
bump up to latest version

s-santillan:

### Please ensure

Update to v20 docusaurus


pabloest (approved):
@s-santillan I tested this locally and it looks good to me. Thank you for the upgrade!

PR #559, opened 2022-06-02
added yarn script dev to prevent replacements script

s-santillan:

### Please ensure

### Usage
This PR creates a `yarn dev` command which is identical to yarn start without running the replacements script. Use when developing docs. Run `yarn start` if you want to see the final form of the docs (with replacements).


pabloest (approved):
What a simple solution!

Could you also update the README so that others know to run `yarn dev` instead of `yarn start`?

PR #565, opened 2022-06-07
Add fail open, fail close mention

adamkvitek:
Resolves: https://linear.app/r2c/issue/MKT-365/document-ci-handling-fail-open-and-fail-closestates

### Please ensure


adamkvitek:
@daghan is this a specific state which CI somehow displays (in that case it should be `fail-open` or `fail-closed`), or is this a general tech support terminology (in that case it should be just "fail open" and "fail closed")? I tried to search for fail open and fail closed for GitHub Actions, it seems more like terminology used outside of the system itself.

daghan:
@dlukeomalley can you chime in here? 

@adamkvitek We use the nomenclature a lot with our customers. The CI agent doesn't display a specific `fail-open` or `fail-closed` state so it is #2 (a part of general tech support methodology)

daghan (approved):
Looks great

s-santillan (approved):
lgtm!

PR #571, opened 2022-06-10
Sara/mkt 373 disclaimer for config auto sending

s-santillan:

### Please ensure

Quick PR here just to add some notices that we send the project URL when using config auto. In other places this is already mentioned, so I just covered the few places it isn't.

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -36,7 +36,10 @@ Start by running [Semgrep](https://github.com/returntocorp/semgrep/) locally to 
     semgrep -e '$X == $X' --lang=py path/to/src
     ```
 
-When you use the Semgrep Registry for any ruleset (like the auto ruleset above), [usage metrics](../metrics) are collected.
+:::note
+* `--config auto` sends your repository's git URL to [Semgrep Registry](https://semgrep.dev/r) to find rules configured for your repository and as a key for cached rule recommendations.
+* When Semgrep Registry is used, [usage metrics](../metrics) are collected.
+:::
 
 ### Next steps
 
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -37,8 +37,8 @@ Start by running [Semgrep](https://github.com/returntocorp/semgrep/) locally to 
     ```
 
 :::note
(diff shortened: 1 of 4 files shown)


docs/getting-started.md
```diff
@@ -36,7 +36,10 @@ Start by running [Semgrep](https://github.com/returntocorp/semgrep/) locally to
     semgrep -e '$X == $X' --lang=py path/to/src
     ```
 
-When you use the Semgrep Registry for any ruleset (like the auto ruleset above), [usage metrics](../metrics) are collected.
+:::note
+* `--config auto` sends your repository's git URL to [Semgrep Registry](https://semgrep.dev/r) to find rules configured for your repository and as a key for cached rule recommendations.
+* When Semgrep Registry is used, [usage metrics](../metrics) are collected.
```
pabloest, line 41: Should this be “…are collected by default” or otherwise indicating that it happens unless the user operates Semgrep in a way not to collect those metrics?

docs/running-rules.md
```diff
@@ -34,7 +34,10 @@ semgrep --config=<ruleset-id> path/to/src
 semgrep --config=<ruleset-id> --config=path/to/yml path/to/src
 ```
 
-When the Registry is used, [usage metrics](../metrics) are collected.
+:::note
+* `--config auto` sends your repository's git URL to [Semgrep Registry](https://semgrep.dev/r) to find rules configured for your repository and as a key for cached rule recommendations.
+* When Semgrep Registry is used, [usage metrics](../metrics) are collected.
```
pabloest, line 39: Should this be “…are collected by default” or otherwise indicating that it happens unless the user operates Semgrep in a way not to collect those metrics?

s-santillan:
@pabloest My understanding is that the URL will always be sent, while the other metrics can be turned off, I will confirm first!

s-santillan:
@pabloest Have addressed your note to add "by default" for usage metrics. Have also changed wording from git URL to project URL (the difference is that project URL is truncated, i.e. returntocorp/semgrep-docs is the project URL). I THINK I should update the usage metrics in semgrep's repository to capture the bit about the project repo behaving as a key rather than as a usage metric.

This PR is ready for another review :)

pabloest (approved):
Thanks, @s-santillan! LGTM.

Yes, any updates to the metrics text should go into the Semgrep repository and then those changes will propagate here automagically.

PR #572, opened 2022-06-13
Fix and revert markdown-link-check.yml

adamboros:
The mentioned issue was fixed upstream and the 'with:' was accidentally removed which causes run issues now. Fixed both and lifted the version pin (on @v1 no breaking changes should happen).

# Thanks for improving Semgrep Docs 

### Please ensure


s-santillan (approved):
Thank you so much @adamboros !

Issue #340, opened 2021-11-07
Broken links in "trophy case"


inkydragon:
Search [contrib.dlint.redos](https://semgrep.dev/r?q=contrib.dlint.redos&lang=go%2Cjavascript%2Cpython%2Cjson%2Cjava%2Cruby%2Cocaml%2Cc%2Cnone%2Cts) return nothing.

All links like `contrib.dlint.redos.*` are broken.

`dlint.redos` was removed in https://github.com/returntocorp/semgrep-rules/pull/1397

PR #578, opened 2022-06-16
Fix broken links

pabloest:

### Please ensure

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -67,7 +67,7 @@ Semgrep supports 20+ languages.
 
 </div>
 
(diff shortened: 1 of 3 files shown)


s-santillan (approved):
Thank you!

adamkvitek:
Thank you! Adding a link also...

adamkvitek:
@pabloest just thinking - somebody reported this issue https://github.com/returntocorp/semgrep-docs/issues/340. What do you think would be a suitable solution? May I delete all `contrib.dlint.redos.*` and push as a commit to this PR or do it in a separate PR?

pabloest:
@adamkvitek I’d make a separate PR. It’s not clear what would replace those links and if there’s nothing to replace them, what should be done with the content that’s associated with those links. It’s worth discussing.

PR #600, opened 2022-07-08
Document extract mode

kopecs:
Depends on https://github.com/returntocorp/semgrep/pull/5621

### Please ensure

- [ ] Redirects are added if the PR changes page URLs (n/a?)

### DIFF
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -66,7 +66,7 @@ else:
     print("Success")
 ```
 
-Likewise, if our query included a Dockerfile such as the example below Semgrep would match the contained Bash and report a match with the `curl-eval` rule.
+Likewise, if our query included a Dockerfile with an offending Bash command, such as the example below, Semgrep would match the contained Bash and report a match with the `curl-eval` rule.
 
 ```dockerfile
 FROM fedora
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -76,7 +76,7 @@ RUN eval `curl -s "http://www.very-secure-website.net"`
 
 ## Extract mode rule schema
 
-Extract mode rules **require** the following usual Semgrep rule keys:
+Extract mode rules **require** the following [usual Semgrep rule keys](../../writing-rules/rule-syntax/#required):
   - `id`
   - `languages`
   - One of `pattern`, `patterns`, `pattern-either`, or `pattern-regex`
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -1,24 +1,20 @@
 ---
 id: extract-mode
 append_help_link: true
-description: "Extract mode allows for easier handling of files containing more than one language."
+description: "This article explains the extract mode, which allows for easier handling of files containing more than one language."
 ---
 
 # Extract mode
 
-Extract mode enables the ability to run existing rules on subsections of files
-where the rule language is different than the language of the file. For
-example, running a Javascript rule on code contained inside of script tags in
-an HTML document.
+## Introduction
 
-## Example
+Extract mode enables you to run existing rules on subsections of files where the rule language is different than the language of the file. For example, running a Javascript rule on code contained inside of script tags in an HTML document.
 
-Without extract mode, writing rules to validate template, markdown or
-configuration files which contain code in another language can be burdensome
-and require significant rule duplication.
+## Example of extract mode
 
-Let's take the following bash rule as an example (a simplified version of [this
-rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
+Without extract mode, writing rules to validate template, Markdown or configuration files which contain code in another language can be burdensome and require significant rule duplication.
+
+Let's take the following Bash rule as an example (a simplified version of [our `curl-eval` rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
 
 ```yaml
 rules:
@@ -37,14 +33,9 @@ rules:
       - pattern: eval ...
 ```
 
-Naïvely this rule would only be run against bash files. However, our project
-might contain Dockerfiles or Python scripts which invoke bash
-commands&mdash;this rule won't currently be any bash contained in those files.
+Usually, this rule would only be run against Bash files. However, our project might contain Dockerfiles or Python scripts which invoke Bash commands&mdash;this rule won't currently be run against any Bash contained in those files.
 
-We can change that by using extract mode! With extract mode we can supplement
-our bash rule with instructions on how to extract any bash commands we might be
-using in a Docker `RUN` instruction or via Python's `os.system` standard
-library function.
+However, with extract mode we can supplement our Bash rule with instructions on how to extract any Bash commands we might be using in a Docker `RUN` instruction or via Python's `os.system` standard library function.
 
 ```yaml
 rules:
@@ -64,7 +55,7 @@ rules:
     dest-language: bash
 ```
 
-With the addition of these extract rules, a Python file like
+By adding the above extract mode rules, Semgrep matches the below Python file, reporting the contained Bash as matching with the `curl-eval` rule.
 
 ```python
 from os import system
@@ -75,7 +66,7 @@ else:
     print("Success")
 ```
 
-or a Dockerfile such as
+Likewise, if our query included a Dockerfile such as the example below Semgrep would match the contained Bash and report a match with the `curl-eval` rule.
 
 ```dockerfile
 FROM fedora
@@ -83,33 +74,30 @@ RUN dnf install -y unzip zip curl which
 RUN eval `curl -s "http://www.very-secure-website.net"`
 ```
 
-will produce a match with the `curl-eval` rule against the contained bash.
+## Extract mode rule schema
 
-## Rule Schema
+Extract mode rules **require** the following usual Semgrep rule keys:
+  - `id`
+  - `languages`
+  - One of `pattern`, `patterns`, `pattern-either`, or `pattern-regex`
 
-Like "normal" (i.e., search mode) rules, extract mode rules require the fields
-`id`, `languages`, and one of `pattern`, `patterns`, `pattern-either` or
-`pattern-regex`. Extract mode rules also have two additional required fields:
-`extract` and `dest-language`.
+Extract mode rules **also require** two additional fields:
+  - `extract`
+  - `dest-language`
+
+The extract mode specific fields are further explained in two sections below.
 
 ### `extract`
 
-The `extract` key is required when in extract mode. The value must be a
-metavariable appearing in your pattern(s), and the value bound to that
-metavariable will be extracted.
+The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and the value bound to that metavariable will be extracted.
 
 ### `dest-language`
 
-The `dest-language` key is required in extract mode. The value must be a
-[language tag](../../writing-rules/rule-syntax/#language-extensions-and-tags)
-other than `generic`.
+The `dest-language` key is required when in extract mode. The value must be a [language tag](../../writing-rules/rule-syntax/#language-extensions-and-tags) other than `generic`.
 
-## Limitations
 
-Currently, extract mode does not support additional processing for the extracted
-text, such as unescaping strings, which might be useful for some applications
-like Jupyter Notebooks.
+## Limitations of extract mode
 
-While extract mode can help enable rules which try and track taint across a
-language boundary within a file, taint rules cannot have a source and sink
-split across the original file and extracted text. 
+Currently, extract mode does not support additional processing for the extracted text, such as unescaping strings, which might be useful for some applications like Jupyter Notebooks.
+
+While extract mode can help to enable rules which try and track taint across a language boundary within a file, taint rules cannot have a source and sink split across the original file and extracted text.
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -1,0 +1,115 @@
+---
+id: extract-mode
+append_help_link: true
+description: "Extract mode allows for easier handling of files containing more than one language."
+---
+
+# Extract mode
+
+Extract mode enables the ability to run existing rules on subsections of files
+where the rule language is different than the language of the file. For
+example, running a Javascript rule on code contained inside of script tags in
+an HTML document.
+
+## Example
+
+Without extract mode, writing rules to validate template, markdown or
+configuration files which contain code in another language can be burdensome
+and require significant rule duplication.
+
+Let's take the following bash rule as an example (a simplified version of [this
+rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
+
+```yaml
+rules:
+  - id: curl-eval
+    severity: WARNING
+    languages:
+      - bash
+    message: Evaluating data from a `curl` command is unsafe.
+    mode: taint
+    pattern-sources:
+      - pattern: |
+          $(curl ...)
+      - pattern: |
+          `curl ...`
+    pattern-sinks:
+      - pattern: eval ...
+```
+
+Naïvely this rule would only be run against bash files. However, our project
+might contain Dockerfiles or Python scripts which invoke bash
+commands&mdash;this rule won't currently be any bash contained in those files.
+
+We can change that by using extract mode! With extract mode we can supplement
+our bash rule with instructions on how to extract any bash commands we might be
+using in a Docker `RUN` instruction or via Python's `os.system` standard
+library function.
+
+```yaml
+rules:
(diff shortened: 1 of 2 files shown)


docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,115 @@
...
+[language tag](../../writing-rules/rule-syntax/#language-extensions-and-tags)
+other than `generic`.
+
+## Limitations
+
+Currently, extract mode does not support additional processing for the extracted
+text, such as unescaping strings, which might be useful for some applications
+like Jupyter Notebooks.
+
+While extract mode can help enable rules which try and track taint across a
```
adamkvitek, line 113:
```suggestion
While extract mode can help to enable rules which try and track taint across a
```

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,115 @@
...
+metavariable appearing in your pattern(s), and the value bound to that
+metavariable will be extracted.
+
+### `dest-language`
+
+The `dest-language` key is required in extract mode. The value must be a
+[language tag](../../writing-rules/rule-syntax/#language-extensions-and-tags)
+other than `generic`.
+
+## Limitations
```
adamkvitek, line 107: Just to avoid issues when users will search for "limitations" to find some specific Semgrep limitations.
```suggestion
## Limitations of generic mode
```

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,115 @@
...
+
+Currently, extract mode does not support additional processing for the extracted
+text, such as unescaping strings, which might be useful for some applications
+like Jupyter Notebooks.
+
+While extract mode can help enable rules which try and track taint across a
+language boundary within a file, taint rules cannot have a source and sink
+split across the original file and extracted text.
```
adamkvitek, line 115: Please, make both of these last paragraphs soft-wrapped as well.

adamkvitek (commented):
Thank you @kopecs for this documentation! Few notes: line breaks can significantly limit editing in GitHub comments for Markdown texts, so a lot of these suggestions are for simple reformatting. I made some suggestions to headers, left one question, and generally tried to make some sentences flow a little better. Thank you 🙇. 

These are just nits, I find the structure and explanation style of this document very satisfactory. Let me know when you implement these suggestions, I will have a quick check again.

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,115 @@
...
+    pattern-sources:
+      - pattern: |
+          $(curl ...)
+      - pattern: |
+          `curl ...`
+    pattern-sinks:
+      - pattern: eval ...
+```
+
+Naïvely this rule would only be run against bash files. However, our project
```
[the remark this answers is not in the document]
kopecs: Ah that's definitely missing a word or two---should read:

this rule won't currently be [run against] any bash contained in those files.

docs/experiments/extract-mode.md
adamkvitek: Oh yeah 🙈 I messed up here. :D fixed to extract mode @kopecs

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+```dockerfile
+FROM fedora
+RUN dnf install -y unzip zip curl which
+RUN eval `curl -s "http://www.very-secure-website.net"`
+```
+
+## Extract mode rule schema
+
+Extract mode rules **require** the following usual Semgrep rule keys:
```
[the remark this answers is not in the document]
adamkvitek: Feel free to do so, thank you!

adamkvitek (approved):

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+  - `id`
+  - `languages`
+  - One of `pattern`, `patterns`, `pattern-either`, or `pattern-regex`
+
+Extract mode rules **also require** two additional fields:
+  - `extract`
+  - `dest-language`
+
```
adamkvitek, line 87: Do we have one rule which uses extract mode in our rules repository? It would be cool to put a link to it here, so users can learn about `extract` and `dest-language` also from reference.

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+```python
+from os import system
+
+if system('eval `curl -s "http://www.very-secure-website.net"`'):
+    print("Command failed!")
+else:
+    print("Success")
+```
+
+Likewise, if our query included a Dockerfile with an offending Bash command, such as the example below, Semgrep would match the contained Bash and report a match with the `curl-eval` rule.
```
adamkvitek, line 69: Not blocking, feel free to dismiss this comment: Maybe we can say: `problematic Bash command` instead of `offending Bash command`

docs/experiments/extract-mode.md
kopecs: No, we currently don't. I can submit another revision once we get to that point though.

adamkvitek: @kopecs that sounds great!

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+          $(curl ...)
+      - pattern: |
+          `curl ...`
+    pattern-sinks:
+      - pattern: eval ...
+```
+
+Usually, this rule would only be run against Bash files. However, our project might contain Dockerfiles or Python scripts which invoke Bash commands&mdash;this rule won't currently be run against any Bash contained in those files.
+
+However, with extract mode we can supplement our Bash rule with instructions on how to extract any Bash commands we might be using in a Docker `RUN` instruction or via Python's `os.system` standard library function.
```
adamkvitek, line 38:
```suggestion
However, with extract mode we can supplement our Bash rule with instructions on how to extract any Bash commands we might be using in a Docker `RUN` instruction or through Python's `os.system` standard library function.
```

docs/experiments/extract-mode.md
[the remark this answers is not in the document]
kopecs: Added in [1a2c6be](https://github.com/returntocorp/semgrep-docs/pull/600/commits/1a2c6bec2be42ccaae57246a21c3c5eda4a64b5e)

docs/experiments/extract-mode.md
adamkvitek: Avoid this abbreviation and use through more often. It can be a bit harder to understand by non-natives (see via in https://developers.google.com/style/word-list?hl=en).

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+    pattern-sources:
+      - pattern: |
+          $(curl ...)
+      - pattern: |
+          `curl ...`
+    pattern-sinks:
+      - pattern: eval ...
+```
+
+Usually, this rule would only be run against Bash files. However, our project might contain Dockerfiles or Python scripts which invoke Bash commands&mdash;this rule won't currently be run against any Bash contained in those files.
```
adamkvitek, line 36:
```suggestion
Usually, this rule would only be run against Bash files. However, our project might contain Dockerfiles or Python scripts which invoke Bash commands&mdash;without extract mode, Semgrep will **not** run this rule against any Bash contained in those files.
```

The active voice might be better here.

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+  - id: extract-python-os-system-to-bash
+    mode: extract
+    languages:
+      - python
+    pattern: os.system("$CMD")
+    extract: $CMD
+    dest-language: bash
+```
+
+By adding the above extract mode rules, Semgrep matches the below Python file, reporting the contained Bash as matching with the `curl-eval` rule.
```
adamkvitek, line 58:
```suggestion
By adding the extract mode rules as shown above, Semgrep can match the following Python file example. Semgrep can match the contained Bash as matching with the `curl-eval` rule.
```

docs/experiments/extract-mode.md
adamkvitek: Understand then. Feel free to ignore this comment then. Maybe "violates" the Bash rule would be too strong here.

adamkvitek: Understand. Feel free to ignore this comment then. Maybe "violates" the Bash rule would be too strong here.

kopecs:
Going to hold off on merging until the corresponding PR for Semgrep.

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+    pattern-sources:
+      - pattern: |
+          $(curl ...)
+      - pattern: |
+          `curl ...`
+    pattern-sinks:
+      - pattern: eval ...
+```
+
+Usually, this rule would only be run against Bash files. However, our project might contain Dockerfiles or Python scripts which invoke Bash commands&mdash;without an extract mode rule Semgrep won't know about any contained Bash, and therefore, will **not** run any Bash rules against those contained commands.
```
adamkvitek, line 36: Blocking:
- Avoid would, should, may - see https://developers.google.com/style/word-list?hl=en#would
  Screenshot 2022-07-12 at 23 03 21
- Avoid software anthropomorphism - Semgrep cannot know anything by itself. See [Avoiding software anthropomorphism](https://www.notion.so/r2cdev/r2c-technical-communication-style-guide-4d1104e78927460aa8c7624aedd6561f#1d51679a2882492abaf13f4ed2b822ea) in r2c technical communication style guide.
- Do not use "we", "our" and similar words unless really necessary. See: [Use of we](https://www.notion.so/r2cdev/r2c-technical-communication-style-guide-4d1104e78927460aa8c7624aedd6561f#39871bd6aca04e37bac6dc2e70dd2baf) in r2c technical communication style guide.
- Avoid [will](https://developers.google.com/style/word-list?hl=en#will) when possible. Documentation needs to be written mostly in the present simple tense: See will in the Google developer style guide.Avoid [will](https://developers.google.com/style/word-list?hl=en#will) when possible. Documentation needs to be written mostly in the present simple tense: See will in the Google developer style guide. 

```suggestion
Usually, Semgrep uses this rule only against Bash files. However, a project might contain Dockerfiles or Python scripts that invoke Bash commands&mdash;without an extract mode rule, Semgrep does **not** run any Bash rules against commands contained in files of different languages.
```

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+However, with extract mode we can provide Semgrep with instructions on how to extract any Bash commands we might be using in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
```
adamkvitek, line 38: It is also better to avoid words such as `might` sometimes. Expressing uncertainty is not the key here. Documentation needs to incline more towards simple English, be more directive, simple, to the point: https://simple.wikipedia.org/wiki/Simplified_English

```suggestion
With extract mode, we can provide Semgrep with instructions on how to extract any Bash commands we are using in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
```

docs/experiments/extract-mode.md
adamkvitek: I probably should write rules for this...

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+  - id: extract-python-os-system-to-bash
+    mode: extract
+    languages:
+      - python
+    pattern: os.system("$CMD")
+    extract: $CMD
+    dest-language: bash
+```
+
+By adding the extract mode rules show above, Semgrep can match the Bash contained in the below Python file, and will report the contained Bash as  matching against the `curl-eval` rule.
```
adamkvitek, line 58: Blocking:
- Do not indicate directions in this manner in one sentence. In this particular case, code snippets below the paragraph are the following code snippets, code snippets that are above the paragraph can be indicated as: in the previous code snippet.
-  Avoid will when possible. Documentation needs to be written mostly in the present simple tense: See [will](https://developers.google.com/style/word-list?hl=en#will) in the Google developer style guide. 

```suggestion
By adding the extract mode rules as shown in the previous code snippet, Semgrep matches the Bash contained in the following Python file and reports the contained Bash as matching against the `curl-eval` rule.
```

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+```python
+from os import system
+
+if system('eval `curl -s "http://www.very-secure-website.net"`'):
+    print("Command failed!")
+else:
+    print("Success")
+```
+
+Likewise, if our query included a Dockerfile with an equivalent Bash command, such as the example below, Semgrep would report the contained Bash as matching against the `curl-eval` rule.
```
adamkvitek, line 69: Blocking:
- Avoid using would: Screenshot 2022-07-12 at 23 25 50 (See [would](https://developers.google.com/style/word-list?hl=en#would) in Google developer style guide)
- Do not use "we", "our", and similar words unless necessary. See: [Use of we](https://www.notion.so/r2cdev/r2c-technical-communication-style-guide-4d1104e78927460aa8c7624aedd6561f#39871bd6aca04e37bac6dc2e70dd2baf) in r2c technical communication style guide.

```suggestion
Likewise, if a query included Dockerfile with an equivalent Bash command, Semgrep reports the contained Bash as matching against the `curl-eval` rule. See the following Dockerfile example that contains a Bash command:
```

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+Extract mode rules **also require** two additional fields:
+  - `extract`
+  - `dest-language`
+
+The extract mode specific fields are further explained in two sections below.
+
+### `extract`
+
+The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and the value bound to that metavariable will be extracted.
```
adamkvitek, line 92: Blocking: 
- Use active voice - is extracted - Semgrep extracts - what do we mean by extracting it actually? Detected maybe?
- Avoid [will](https://developers.google.com/style/word-list?hl=en#will) when possible. Documentation needs to be written mostly in the present simple tense: See will in the Google developer style guide.

```suggestion
The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and Semgrep extracts the value bound to that metavariable.
```

kopecs:
Argh, @adamkvitek I apparently have the worst possible timing. Do you want me to revert the merge and open another PR, or just open a fresh one to address those comments?

adamkvitek:
@kopecs oh no problem here. Just open a new one 🙏 - I can do it for you if you have other pressing tasks and you can just check my suggestions. I included some more explanations as you may see above, so you can understand some of my reasoning here. I should really create rules for our docs repository from them... these are repeated stylistic mistakes.

docs/experiments/extract-mode.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Introduction
+
+Extract mode enables you to run existing rules on subsections of files where the rule language is different than the language of the file. For example, running a Javascript rule on code contained inside of script tags in an HTML document.
+
+## Example of extract mode
+
+Without extract mode, writing rules to validate template, Markdown or configuration files which contain code in another language can be burdensome and require significant rule duplication.
+
+Let's take the following Bash rule as an example (a simplified version of [our `curl-eval` rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
```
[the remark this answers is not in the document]
kopecs: The general form of "<rule id> rule" as a way to refer to a specific rule feels very awkward to me. Do we have a standard convention here? If we'd like to avoid possessive determiners, how about "`curl-eval`, from the Semgrep Registry"?

adamkvitek:
Here is the PR @kopecs! https://github.com/returntocorp/semgrep-docs/pull/607

docs/experiments/extract-mode.md
kopecs: Most of this looks good, but I believe it should still be "a Dockerfile". To the best of my knowledge, Docker doesn't call the "language" Dockerfiles are written in anything specific, and just refers to it as a text format, so I believe this ought to refer to a specific file. See, e.g., the usage https://docs.docker.com/engine/reference/builder/.

docs/experiments/extract-mode.md
kopecs: I wasn't sure how much to talk about the actual mechanics of what we do for a page like this, but the "extraction" essentially comprises the process of reading the text bound to the metavariable from the original file, writing it to a temporary file, and then running rules for the destination language against that file.

It is unclear to me if that description is more or less helpful than the admittedly nebulous "extracts".

The suggestion you have seems good to me wrt your other points.

docs/experiments/extract-mode.md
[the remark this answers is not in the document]
kopecs: [migrated to other branch]

docs/experiments/extract-mode.md
kopecs: [migrated to other branch]

docs/experiments/extract-mode.md
kopecs: [migrated to other branch]

PR #607, opened 2022-07-12
Fix minor stylistic issues in extract mode document

adamkvitek:

### Please ensure

### DIFF
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -14,7 +14,7 @@ Extract mode enables you to run existing rules on subsections of files where the
 
 Without extract mode, writing rules to validate template, Markdown or configuration files which contain code in another language can be burdensome and require significant rule duplication.
 
-Let's take the following Bash rule as an example (a simplified version of [`curl-eval` rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
+Let's take the following Bash rule as an example (a simplified version of the [`curl-eval`](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml) rule form the Semgrep Registry):
 
 ```yaml
 rules:
@@ -35,7 +35,7 @@ rules:
 
 Usually, Semgrep uses this rule only against Bash files. However, a project might contain Dockerfiles or Python scripts that invoke Bash commands&mdash;without an extract mode rule, Semgrep does **not** run any Bash rules against commands contained in files of different languages.
 
-However, with extract mode we can provide Semgrep with instructions on how to extract any Bash commands we might be using in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
+However, with extract mode, you can provide Semgrep with instructions on how to extract any Bash commands used in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
 
 ```yaml
 rules:
@@ -89,7 +89,7 @@ The extract mode specific fields are further explained in two sections below.
 
 ### `extract`
 
-The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and Semgrep extracts the value bound to that metavariable.
+The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s). Semgrep uses the code bound to the metavariable for subsequent queries of non-extract mode rules targeting `dest-language`.
 
 ### `dest-language`
 
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -66,7 +66,7 @@ else:
     print("Success")
 ```
 
-Likewise, if a query included Dockerfile with an equivalent Bash command, Semgrep reports the contained Bash as matching against the `curl-eval` rule. See the following Dockerfile example that contains a Bash command:
+Likewise, if a query included a Dockerfile with an equivalent Bash command, Semgrep reports the contained Bash as matching against the `curl-eval` rule. See the following Dockerfile example that contains a Bash command:
 
 ```dockerfile
 FROM fedora
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -14,7 +14,7 @@ Extract mode enables you to run existing rules on subsections of files where the
 
 Without extract mode, writing rules to validate template, Markdown or configuration files which contain code in another language can be burdensome and require significant rule duplication.
 
-Let's take the following Bash rule as an example (a simplified version of the [`curl-eval`](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml) rule form the Semgrep Registry):
+Let's take the following Bash rule as an example (a simplified version of the [`curl-eval`](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml) rule from the Semgrep Registry):
 
 ```yaml
 rules:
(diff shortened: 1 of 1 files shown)


docs/experiments/extract-mode.md
```diff
@@ -14,7 +14,7 @@ Extract mode enables you to run existing rules on subsections of files where the
 
 Without extract mode, writing rules to validate template, Markdown or configuration files which contain code in another language can be burdensome and require significant rule duplication.
 
-Let's take the following Bash rule as an example (a simplified version of [our `curl-eval` rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
+Let's take the following Bash rule as an example (a simplified version of [`curl-eval` rule](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml)):
```
kopecs, line 17: The general form of " rule" as a way to refer to a specific rule feels very awkward to me. Do we have a standard convention here? If we'd like to avoid possessive determiners, how about "curl-eval, from the Semgrep Registry"?

docs/experiments/extract-mode.md
```diff
@@ -66,7 +66,7 @@ else:
     print("Success")
 ```
 
-Likewise, if our query included a Dockerfile with an equivalent Bash command, such as the example below, Semgrep would report the contained Bash as matching against the `curl-eval` rule.
+Likewise, if a query included Dockerfile with an equivalent Bash command, Semgrep reports the contained Bash as matching against the `curl-eval` rule. See the following Dockerfile example that contains a Bash command:
```
kopecs, line 69: Most of this looks good, but I believe it should still be "a Dockerfile". To the best of my knowledge, Docker doesn't call the "language" Dockerfiles are written in anything specific, and just refers to it as a text format, so I believe this ought to refer to a specific file. See, e.g., the usage https://docs.docker.com/engine/reference/builder/.

docs/experiments/extract-mode.md
```diff
@@ -89,7 +89,7 @@ The extract mode specific fields are further explained in two sections below.
 
 ### `extract`
 
-The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and the value bound to that metavariable will be extracted.
+The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and Semgrep extracts the value bound to that metavariable.
```
kopecs, line 92: > Use active voice - is extracted - Semgrep extracts - what do we mean by extracting it actually? Detected maybe?


I wasn't sure how much to talk about the actual mechanics of what we do for a page like this, but the "extraction" essentially comprises the process of reading the text bound to the metavariable from the original file, writing it to a temporary file, and then running rules for the destination language against that file.

It is unclear to me if that description is more or less helpful than the admittedly nebulous "extracts".

The suggestion you have seems good to me wrt your other points.

docs/experiments/extract-mode.md
adamkvitek: That sounds good, what about: "`curl-eval` rule from the Semgrep Registry!"?

```suggestion
Let's take the following Bash rule as an example (a simplified version of [`curl-eval`](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml) rule from the Semgrep Registry):
```

docs/experiments/extract-mode.md
adamkvitek, line 69:
```suggestion
Likewise, if a query included a Dockerfile with an equivalent Bash command, Semgrep reports the contained Bash as matching against the `curl-eval` rule. See the following Dockerfile example that contains a Bash command:
```

adamkvitek: Good to know, thanks! Changing to "a Dockerfile".

docs/experiments/extract-mode.md
adamkvitek: Interesting mechanics @kopecs! Surely, I would avoid explaining it to external users as it is not needed for the use of it. So is the following suggestion good enough?

```suggestion
The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s), and Semgrep detects the value bound to that metavariable.
```

docs/experiments/extract-mode.md
```diff
@@ -33,7 +33,7 @@ rules:
       - pattern: eval ...
 ```
 
-Usually, this rule would only be run against Bash files. However, our project might contain Dockerfiles or Python scripts which invoke Bash commands&mdash;without an extract mode rule Semgrep won't know about any contained Bash, and therefore, will **not** run any Bash rules against those contained commands.
+Usually, Semgrep uses this rule only against Bash files. However, a project might contain Dockerfiles or Python scripts that invoke Bash commands&mdash;without an extract mode rule, Semgrep does **not** run any Bash rules against commands contained in files of different languages.
 
 However, with extract mode we can provide Semgrep with instructions on how to extract any Bash commands we might be using in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
```
adamkvitek, line 38: @kopecs just suggesting a little simplification again, to avoid the use of `we`.

```suggestion
However, with extract mode, Semgrep can obtain instructions on how to extract any Bash commands used in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
```

s-santillan (approved):
LGTM! For my own knowledge, @kopecs - we'd still need to make a rule per dest-lang/lang pair, right? (For example, html/javascript or html/css)

docs/experiments/extract-mode.md
kopecs: My concern with the proposed phrasing would be that it implies that Semgrep would somehow automatically obtain these instructions, instead of them being provided explicitly by the user or in a rulepack. What about:

"However, with extract mode, you can provide Semgrep with instructions on how to extract any Bash commands used in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function."

docs/experiments/extract-mode.md
kopecs: I would be concerned that "detects" is less clear. I think we could just be more explicit, without going into the level of detail again. Perhaps something like:

The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s). The code bound to the metavariable is used for subsequent queries of non-extract mode rules targeting `dest-language`.

kopecs:
@s-santillan Yep, that's correct, although for "standard" cases I imagine we can vendor rules (e.g., the cases you mention, at least for script/style tags).

docs/experiments/extract-mode.md
adamkvitek: I agree! Only avoid passive sentence to express the agent, so I would suggest:

```suggestion
The `extract` key is required when in extract mode. The value must be a metavariable appearing in your pattern(s). Semgrep uses the code bound to the metavariable for subsequent queries of non-extract mode rules targeting `dest-language`.
```

docs/experiments/extract-mode.md
adamkvitek: I am not sure if we had any convention here before, but I really like your second suggestion.

```suggestion
Let's take the following Bash rule as an example (a simplified version of the [`curl-eval`](https://github.com/returntocorp/semgrep-rules/blob/release/bash/curl/security/curl-eval.yaml) rule form the Semgrep Registry):
```

docs/experiments/extract-mode.md
adamkvitek: I agree!

```suggestion
However, with extract mode, you can provide Semgrep with instructions on how to extract any Bash commands used in a Docker `RUN` instruction or as an argument to Python's `os.system` standard library function.
```

s-santillan:
@kopecs Got it. Thank you!

PR #608, opened 2022-07-13
Fixtests

0xDC0DE:

### Please ensure

- [X] A subject matter expert (SME) reviews the content
- [X] This change has no security implications or else you have pinged the security team

### DIFF
--- a/docs/writing-rules/testing-rules.md
+++ b/docs/writing-rules/testing-rules.md
@@ -41,17 +41,16 @@ To define a test for autofix behavior:
 2. Within the autofix test file, enter the expected result of applied autofix rule to the test code.
 3. Run `semgrep --test` to verify that your autofix test file is correctly detected.
 
-When you use `semgrep --test`, Semgrep applies the autofix rule to the original test code (`path/to/rule.py`), and then verifies whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py)`. If there is a mismatch, the line diffs are printed.```
-   If there is a mismatch, the line diffs will be printed.
+When you use `semgrep --test`, Semgrep applies the autofix rule to the original test code (`path/to/rule.py`), and then verifies whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py)`. If there is a mismatch, the line diffs are printed.
 
 :::info
-Hint: Creating an autofix test for a rule with autofix can take less than a minute with the following flow of commands:
+**Hint**: Creating an autofix test for a rule with autofix can take less than a minute with the following flow of commands:
 ```sh
 cp rule.py rule.fixed.py
 semgrep --config rule.yaml rule.fixed.py --autofix
 ```
 
-These commands apply the rule's autofix to the test code. It is advised to manually inspect whether the fix behaves as expected (e.g. using `vimdiff rule.py rule.fixed.py`).
+These commands apply the autofix of the rule to the test code. After Semgrep delivers a fix, inspect whether the outcome of this fix looks as expected (for example using `vimdiff rule.py rule.fixed.py`).
 :::
 
 ## Example
@@ -103,11 +102,15 @@ No tests for fixes found.
 ```
 
 Semgrep tests automatically avoid failing on lines marked with `# todoruleid` or `# todook`.
+
+## Storing rules and test targets in different directories
 ## Storing rules and test targets in different directories
 
 Creating different directories for rules and tests helps users manage a growing library of custom rules.
+Creating different directories for rules and tests helps users manage a growing library of custom rules.
 To store rules and test targets in different directories use the `--config` option.
-For the following directory structure,
+
+For example, the directory with the following structure:
 
 ```sh
 $ tree tests
@@ -123,13 +126,13 @@ tests
 4 directories, 2 files
 ```
 
-the command
+The following command:
 
 ```sh
 semgrep --test --config tests/rules/ tests/targets/
 ```
 
-produces the same output as in the previous example.
+Produces the same output as in the previous example.
 
 The subdirectory structure of these two directories must be the same for Semgrep to correctly find the associated files.
 
@@ -153,7 +156,7 @@ totally_safe_eval(user_input)
 secure_eval(safe_get_user_input())
 ```
 
-So that the directory structure looks like this:
+So that the directory structure is printed as the following:
 
 ```sh
 $ tree tests
@@ -170,7 +173,7 @@ tests
 4 directories, 2 files
 ```
 
-The command
+The following command:
 
 ```sh
 semgrep --test --config tests/rules/ tests/targets/


docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
 The `.test.yaml` file extension can also be used for test files. This is necessary when testing YAML language rules.
 :::
 
+## Testing autofix
+
+Semgrep's testing mechanism also provides a way to test the behaviour of any fixes defined in the rules.
```
s-santillan, line 36:
```suggestion
Semgrep's testing mechanism also provides a way to test the behavior of any `fix` values defined in the rules.
```

Just making it clear that "fixes" was referring to the `fix` (rule syntax) values.

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
...
+To define a test for autofix behaviour:
```
s-santillan, line 37:
```suggestion
To define a test for autofix behavior: 
```

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
 The `.test.yaml` file extension can also be used for test files. This is necessary when testing YAML language rules.
 :::
 
+## Testing autofix
+
+Semgrep's testing mechanism also provides a way to test the behaviour of any fixes defined in the rules.
+To define a test for autofix behaviour: 
+
+1. Create a new code file with the `.fixed` suffix before the annotation.
```
s-santillan, line 39:
```suggestion
1. Create a new code file with the `.fixed` suffix before the file type. 
```

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
...
+## Testing autofix
+
+Semgrep's testing mechanism also provides a way to test the behaviour of any fixes defined in the rules.
+To define a test for autofix behaviour: 
+
+1. Create a new code file with the `.fixed` suffix before the annotation. 
+   For example, the autofix test file for a rule with test code in `path/to/rule.py` will be named `path/to/rule.fixed.py`.
+2. This file should contain the expected result of applying autofixes to the test code file.
+3. Any future runs of the command `semgrep --test` will apply the autofix to the original testcode (`path/to/rule.py`) and verify whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py`).
+   If there is a mismatch, the line diffs will be printed.
```
s-santillan, line 43: Usually we prefer to have "will" (future tense) be present tense, but I think it's okay in this case? What do you think @adamkvitek ?

adamkvitek: @s-santillan I would avoid using them inside of procedures. These steps do not seem like real steps - why is there "should" in the second step? It must be obvious what needs to be done. See "should" in Google style guide: https://developers.google.com/style/word-list?hl=en#should

For example:
`2. This file should contain the expected result of applying autofixes to the test code file.`
I need directive action inside of procedure. Do we actually want to say, that a user needs to: 
`2. In the test file, include a code that is detected and fixed by the autofix rule.`

Just an illustrative example (sorry if it sounds too simple): These manuals need to be the same as cooking recipes. "When the omelet is still liquid, put some mushrooms on the top of them." Not: "The liquid omelet should contain mushrooms." Ok, so it does not have include mushrooms, only should include them? Should I buy the eggs which already have mushrooms inside (contain?)? Do such eggs exist? Or I need to add them? What should I do? What a difficult recipe. :D

Similarly:
`3. Any future runs of the command `semgrep --test` will apply the autofix to the original testcode (`path/to/rule.py`) and verify whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py`).
   If there is a mismatch, the line diffs will be printed.`
This is actually not a step with an action, but a comment after the user created an autofix rule, right? Removing it from the numbered list, maybe we can change it to:
`When you use `semgrep --test` again, Semgrep applies the autofix rule to the original test code (`path/to/rule.py`) and verifies whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py`). If there is a mismatch, the line diffs are printed.`

I am still not sure what do we mean by the last comment - I would probably need to test it. What exactly we mean by a mismatch here is not clear to me. I am curious to test this out.

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
...
 :::
 
+## Testing autofix
+
+Semgrep's testing mechanism also provides a way to test the behaviour of any fixes defined in the rules.
+To define a test for autofix behaviour: 
+
+1. Create a new code file with the `.fixed` suffix before the annotation. 
+   For example, the autofix test file for a rule with test code in `path/to/rule.py` will be named `path/to/rule.fixed.py`.
+2. This file should contain the expected result of applying autofixes to the test code file.
```
s-santillan, line 41:
```suggestion
2. Within the autofix test file, enter the expected result of applying autofixes to the test code file.
```
Just writing it as a step/verb rather than a state.

docs/writing-rules/testing-rules.md
adamkvitek: Also: Will is more or less fine if there is an action that does not immediately lead to a result, for example "When you send an email, you will receive the answer between 8 or 24 hours." But a user can use `semgrep --test` immediately, so it is better to avoid the use of will here also.

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
...
+3. Any future runs of the command `semgrep --test` will apply the autofix to the original testcode (`path/to/rule.py`) and verify whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py`).
```
s-santillan, line 42:
```suggestion
3. Any future runs of the command `semgrep --test` will apply the autofix to the original test code (`path/to/rule.py`) and verify whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py`).
```

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
...
+   If there is a mismatch, the line diffs will be printed.
+
+:::info
+Hint: creating an autofix test for a rule whose autofix is functioning as expected can take less than a minute with the following flow of commands:
```
s-santillan, line 46:
```suggestion
Hint: Creating an autofix test for a rule with autofix can take less than a minute with the following flow of commands:
```

I removed "whose autofix is functioning as expected" because ... aren't we in the process of testing that the autofix is functioning? I hope I am making sense @pieterdc1 .

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,27 @@ Other than annotations there are three things to remember when creating tests:
...
+```sh
+cp rule.py rule.fixed.py
+semgrep --config rule.yaml rule.fixed.py --autofix
+```
+
+This will apply the fix of the rule to the testcode. It is advised to manually inspect whether the fix behaves as expected (e.g. using `vimdiff rule.py rule.fixed.py`).
```
s-santillan, line 52:
```suggestion
These commands apply the rule's autofix to the test code. It is advised to manually inspect whether the fix behaves as expected (e.g. using `vimdiff rule.py rule.fixed.py`).
```

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +91,115 @@ eval(safe_get_user_input())
...
+1/1: ✓ All tests passed
+No tests for fixes found.
 ```
 
-- True positives (`TP`) correspond to `ruleid`
-- True negatives (`TN`) correspond to `ok`
-- False positives (`FP`) correspond to `todook`
-- False negatives (`FN`) correspond to `todoruleid`
-
+As you can see, semgrep tests automatically avoid failing on lines marked with `# todoruleid` or `# todook`.
```
s-santillan, line 103:
```suggestion
Semgrep tests automatically avoid failing on lines marked with `# todoruleid` or `# todook`.
```

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +91,115 @@ eval(safe_get_user_input())
...
 To avoid failing on TODOs you can specify `--test-ignore-todo`:
```
s-santillan, line 104:
```suggestion

```
Pretty sure we should delete this as this was the old behavior? For confirmation @pieterdc1

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +91,115 @@ eval(safe_get_user_input())
...
-(TODO: 2) rules/detect-eval.yaml
-	✔ insecure-eval-use                                  TP: 1 TN: 1 FP: 0 FN: 0
---------------------------------------------------------------------------------
-final confusion matrix: TP: 1 TN: 1 FP: 0 FN: 0
---------------------------------------------------------------------------------
+semgrep --test --config tests/rules/ tests/targets/
 ```
 
-To store rules and test targets in different directories you can specify `--config`:
+will produce the same output as before.
```
s-santillan, line 129:
```suggestion
produces the same output as in the previous example.
```

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +91,115 @@ eval(safe_get_user_input())
...
+No tests for fixes found.
 ```
 
-- True positives (`TP`) correspond to `ruleid`
-- True negatives (`TN`) correspond to `ok`
-- False positives (`FP`) correspond to `todook`
-- False negatives (`FN`) correspond to `todoruleid`
-
+As you can see, semgrep tests automatically avoid failing on lines marked with `# todoruleid` or `# todook`.
 To avoid failing on TODOs you can specify `--test-ignore-todo`:
```
s-santillan, line 105: ## Storing rules and test targets in different directories

Creating different directories for rules and tests helps users manage a growing library of custom rules.

s-santillan, line 105:
```suggestion
## Storing rules and test targets in different directories

Creating different directories for rules and tests helps users manage a growing library of custom rules.
```

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +91,115 @@ eval(safe_get_user_input())
...
 
+To store rules and test targets in different directories you can specify `--config`.
```
s-santillan, line 106:
```suggestion
To store rules and test targets in different directories use the `--config` option.
```

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +91,115 @@ eval(safe_get_user_input())
...
 ```
 
-To store rules and test targets in different directories you can specify `--config`:
+will produce the same output as before.
 
-```sh
-tree tests
+The subdirectory structure of these two directories must be the same for Semgrep to correctly find the associated files.
+
+To test the autofix behaviour, you can add the autofix test file `rules/detect-eval.fixed.py` to represent the expected outcome of applying the fix to the test code:
```
s-santillan, line 133:
```suggestion
To test the autofix behavior, add the autofix test file `rules/detect-eval.fixed.py` to represent the expected outcome of applying the fix to the test code:
```
Removed "can" because the action is a "must" :)

docs/writing-rules/testing-rules.md
adamkvitek: @s-santillan I see! One hour after I posted this comment according to GitHub. Thank you. :)

docs/writing-rules/testing-rules.md
adamkvitek: @s-santillan Just a nit, I would simplify the second half of the sentence also: `enter the expected result of applied autofix rule to the test code`.

docs/writing-rules/testing-rules.md
pieterdc1: Yes, I meant to delete this, but must have forgotten

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +93,116 @@ eval(safe_get_user_input())
...
---------------------------------------------------------------------------------
+1/1: ✓ All tests passed
+No tests for fixes found.
 ```
 
-- True positives (`TP`) correspond to `ruleid`
-- True negatives (`TN`) correspond to `ok`
-- False positives (`FP`) correspond to `todook`
-- False negatives (`FN`) correspond to `todoruleid`
+Semgrep tests automatically avoid failing on lines marked with `# todoruleid` or `# todook`.
```
adamkvitek, line 105:
```suggestion
Semgrep tests automatically avoid failing on lines marked with `# todoruleid` or `# todook`.

```

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,29 @@ Other than annotations there are three things to remember when creating tests:
...
+Semgrep's testing mechanism also provides a way to test the behavior of any `fix` values defined in the rules.
+To define a test for autofix behavior: 
+
+1. Create a new **autofix test file** with the `.fixed` suffix before the file type extension.  
+   For example, the autofix test file for a rule with test code in `path/to/rule.py` will be named `path/to/rule.fixed.py`.
+2. Within the autofix test file, enter the expected result of applied autofix rule to the test code.
+3. Run `semgrep --test` to verify that your autofix test file is correctly detected.
+
+When you use `semgrep --test`, Semgrep applies the autofix rule to the original test code (`path/to/rule.py`), and then verifies whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py)`. If there is a mismatch, the line diffs are printed.```
+   If there is a mismatch, the line diffs will be printed.
```
s-santillan, line 45:
```suggestion

```
Looks like the line "If there is a mismatch..." was printed twice (see above). Suggesting to remove the second one.

docs/writing-rules/testing-rules.md
```diff
@@ -31,6 +31,29 @@ Other than annotations there are three things to remember when creating tests:
...
+
+Semgrep's testing mechanism also provides a way to test the behavior of any `fix` values defined in the rules.
+To define a test for autofix behavior: 
+
+1. Create a new **autofix test file** with the `.fixed` suffix before the file type extension.  
+   For example, the autofix test file for a rule with test code in `path/to/rule.py` will be named `path/to/rule.fixed.py`.
+2. Within the autofix test file, enter the expected result of applied autofix rule to the test code.
+3. Run `semgrep --test` to verify that your autofix test file is correctly detected.
+
+When you use `semgrep --test`, Semgrep applies the autofix rule to the original test code (`path/to/rule.py`), and then verifies whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py)`. If there is a mismatch, the line diffs are printed.```
```
s-santillan, line 44:
```suggestion
When you use `semgrep --test`, Semgrep applies the autofix rule to the original test code (`path/to/rule.py`), and then verifies whether this matches the expected outcome defined in the autofix test file (`path/to/rule.fixed.py)`. If there is a mismatch, the line diffs are printed.
```
Removing extra backticks.

s-santillan (approved):

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +93,116 @@ eval(safe_get_user_input())
...
+totally_safe_eval = eval
+# todoruleid: insecure-eval-use
+totally_safe_eval(user_input)
+
+# todook: insecure-eval-use
+secure_eval(safe_get_user_input())
 ```
 
-will produce the following output:
+So that the directory structure looks like this:
```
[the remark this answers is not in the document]
adamkvitek: Not sure if rendered is the best word as it may be associated with the process of rendering and different technical terms. Feel free to substitute this word.

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +93,116 @@ eval(safe_get_user_input())
...
 
+The command
+
 ```sh
-python -m semgrep --quiet --test --config /tmp/tests/rules/ /tmp/tests/targets/
+semgrep --test --config tests/rules/ tests/targets/
 ```
 
-will produce the following output:
+will now results in
```
adamkvitek, line 179: Just a comment for @s-santillan - similar to: https://github.com/returntocorp/semgrep-docs/pull/608/files#r920966343

docs/writing-rules/testing-rules.md
```diff
@@ -69,82 +93,116 @@ eval(safe_get_user_input())
...
-(TODO: 2) rules/detect-eval.yaml
-	✔ insecure-eval-use                                  TP: 1 TN: 1 FP: 0 FN: 0
---------------------------------------------------------------------------------
-final confusion matrix: TP: 1 TN: 1 FP: 0 FN: 0
---------------------------------------------------------------------------------
+semgrep --test --config tests/rules/ tests/targets/
 ```
 
-To store rules and test targets in different directories you can specify `--config`:
+produces the same output as in the previous example.
```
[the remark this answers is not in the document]
adamkvitek: Note on previous note: Changing this into a list may show where it begins and ends. Having these explanations outside of a list makes it harder to distinguish the beginning and end of them (bear in mind that people on the internet tend to skim through the page a bit randomly sometimes, so numbered lists and procedures catch their attention - but we can also achieve a similar effect with subheaders and smaller sections without lists maybe).

PR #612, opened 2022-07-14
Update extract mode docs

kopecs:
Documents additional functionality added in <https://github.com/returntocorp/semgrep/pull/5723>.

### Please ensure

### DIFF
--- a/docs/experiments/extract-mode.md
+++ b/docs/experiments/extract-mode.md
@@ -88,7 +88,7 @@ Extract mode rules **also require** two additional fields:
 Extract mode has one **optional** field:
   - `reduce`
 
-The extract mode specific fields are further explained in two sections below.
+The fields specific to extract mode are further explained in the sections below.
 
 ### `extract`
 


docs/experiments/extract-mode.md
```diff
@@ -95,6 +98,14 @@ The `extract` key is required when in extract mode. The value must be a metavari
 
 The `dest-language` key is required when in extract mode. The value must be a [language tag](../../writing-rules/rule-syntax/#language-extensions-and-tags) other than `generic`.
 
+### `reduce`
+
+The `reduce` key is optional when in extract mode. The value of this key specifies a method to combine the ranges extracted by a single rule within a file.
+
+The value must be one of:
+  - `separate` (default)&mdash;treat all matched ranges as separate units for subseqent queries
+  - `concat`&mdash;concatenate all matched ranges together and treat this result as a single unit for subseqent queries
```
kopecs, line 107: I wasn't sure how you would typically information about enumerated options. I figure a table doesn't make sense if there are currently only two options. Is this fine, or is a different format preferred?

adamkvitek: @kopecs what you did is fine, we can use description lists here, I fancy them and advise to use them. :-) See what I wrote about them in our style guide: https://www.notion.so/r2cdev/r2c-technical-communication-style-guide-4d1104e78927460aa8c7624aedd6561f#7161c60e37904848bc30067029cbfbdf

So I suggest the following:

```suggestion
The value of `reduce` key must be one of:
<dl>
    <dt><code>separate</code></dt>
    <dd>Treats all matched ranges as separate units for subsequent queries.</dd>
    <dt><code>concat<code></dt>'
    <dd>Concatenates all matched ranges together and treats this result as a single unit for subsequent queries.</dd>
</dl>
```

Notes:
- Tables are better for data that need to be explained in more than two columns. Description lists are ideal alternatives to two-column tables with longer sentences that explain the data in one column (like a dictionary of terms and explanations).
- You may include capital letters at the beginning and full stops even at the end of sentences in bullet points.
- I used third-person verbs ("treats", and "concatenates") because I guess that Semgrep does these actions when a user uses this option in the extract mode.

docs/experiments/extract-mode.md
```diff
@@ -95,6 +98,14 @@ The `extract` key is required when in extract mode. The value must be a metavari
 
 The `dest-language` key is required when in extract mode. The value must be a [language tag](../../writing-rules/rule-syntax/#language-extensions-and-tags) other than `generic`.
 
+### `reduce`
+
+The `reduce` key is optional when in extract mode. The value of this key specifies a method to combine the ranges extracted by a single rule within a file.
```
adamkvitek, line 103:
```suggestion
The `reduce` key is optional in extract mode. The value of this key specifies a method to combine the ranges extracted by a single rule within a file.
```

docs/experiments/extract-mode.md
adamkvitek: This will be the outcome!
Screenshot 2022-07-15 at 19 30 14

kopecs: What you suggest seems a bit verbose to me; I added a slightly shorted note instead. Let me know if you think there's ambiguity or another issue with this:

https://github.com/returntocorp/semgrep-docs/blob/e50c835bf804a3cef27a0753db2a891775b7bc44/docs/experiments/extract-mode.md?plain=1#L108

adamkvitek: @kopecs good idea! Seems clear and short.

kopecs:
Force push to rebase/resolve merge conflict.

docs/experiments/extract-mode.md
```diff
@@ -85,15 +85,30 @@ Extract mode rules **also require** two additional fields:
   - `extract`
   - `dest-language`
 
+Extract mode has one **optional** field:
+  - `reduce`
+
 The extract mode specific fields are further explained in two sections below.
```
adamkvitek, line 91:
```suggestion
The fields specific to extract mode are further explained in the sections below.
```

PR #617, opened 2022-07-16
Init move away from purple to new greens

pabloest:

### Please ensure


pabloest:
@adamboros This is just a draft PR for now but what do you think of this direction for the docs, so we match the upcoming move towards greens?

Screen Shot 2022-07-18 at 10 28 12 AM

pabloest:
We could closely match the nav like this:

Screen Shot 2022-07-20 at 1 51 40 PM
 
However, I’m not sure if the nav items should be aligned like this (matching the new updated navbar on the homepage) or if they should be left-aligned to better fit the overall Docusaurus framework.

adamboros:
Looks great!
- I'm for left-aligned to match Docusaurus. I think it's okay for the Docs to be distinctive like that. 
- The link color should be 1 or 2 level darker, cause it seems to be low on contrast

pabloest:
Thanks, @adamboros! I’ll make edits per your guidance.

pabloest:
### Preview update

Screen Shot 2022-07-20 at 10 00 39 PM

s-santillan:
The state was moved from draft to normal PR - so does this mean we want to merge this already?

pabloest:
@s-santillan Thanks for noticing. It’s ready to review but, if approved, will wait to merge it.

pabloest:
cc @adamkvitek and @s-santillan this is merging, just tagging you for awareness

adamkvitek (approved):

adamkvitek:
@pabloest approved! I like the bigger top bar now.

Not blocking: Only one note - is the level of contrast for the green hyperlinks ok for you? They seem a little less contrasting to me. (I am fine with them in the dark mode.) I am also not sure what would be a good solution here. 
Screenshot 2022-08-12 at 11 22 10
Screenshot 2022-08-12 at 11 22 24

pabloest:
@adamkvitek Thanks for the review and feedback.

For the link color, do you refer to the link in the main body of your screenshot or the links on the right-side TOC?

adamkvitek:
@pabloest links in the main body mostly. We do not have to make all links suddenly in the company color. For example, [Twilio - their company colors seem to be red](https://www.twilio.com/) (which is a powerful contrasting color to white already, given how humans are sensitive to this color). However, https://www.twilio.com/docs/sms/quickstart/java (maybe because this is somehow a standard on the web in general). 

Btw. one thing I love about their docs - try to highlight a piece of text - inverts text color and the highlight is red!

pabloest:
@adamkvitek I see, thanks. Yes, perhaps we can have a different link color, like blue.

@adamboros What do you think about using a different color for text hyperlinks, like blue? If we do that, would it mean that sidebar nav links should also change to blue (or whatever color)?

adamkvitek:
My feedback is not entirely fruitful here as I do not know a better solution. If we would have another color that gives a better contrast to the white background I would go for it, but we rely on our new website (which looks good!). We use only the black and green combination now. Links cannot be black, so green is the only way now? I am curious if you have any idea now...

pabloest:
Thanks, @adamkvitek. I’m going to merge this PR for now and we can continue the discussion about the link color here or separately.

adamkvitek:
@pabloest I agree. Maybe @adamboros will have an idea how to fix this (ad [this comment](https://github.com/returntocorp/semgrep-docs/pull/617#issuecomment-1213263132)).

adamboros:
@pabloest @adamkvitek I've been thinking about using the blue palette within the app for the same concerns you discuss here and we've been discussing within design to go back to the blue link color. If you want, I'm happy to take some time this week to experiment on this and see how can we include the brand emerald still.

pabloest:
Thanks, @adamboros! That sounds wonderful.

PR #618, opened 2022-07-18
Add Semgrep CI YAML file

adamkvitek:

### Please ensure

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,5 +1,5 @@
-Review CI configuration
-Review our semgrep.yml config file.
+# Review CI configuration
+# Review our semgrep.yml config file.
 
 name: Semgrep
 on:
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,6 +1,3 @@
-# Review CI configuration
-# Review our semgrep.yml config file.
-
 name: Semgrep
 on:
   pull_request: {}
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,26 @@
+Review CI configuration
+Review our semgrep.yml config file.
+
+name: Semgrep
+on:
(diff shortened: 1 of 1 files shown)


holdenmcg (approved):

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,26 @@
+Review CI configuration
+Review our semgrep.yml config file.
```
pabloest, line 2: Are these comments? If so, shouldn’t they be commented lines?

.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,26 @@
...
+    paths:
+      - .github/workflows/semgrep.yml
+  schedule:
+    - cron: '0 0 * * 0'
+jobs:
+  semgrep:
+    name: Scan
+    runs-on: ubuntu-20.04
+    env:
+      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
```
pabloest, line 20: If this connects to Semgrep App, won’t the whole repo be connected and follow the Semgrep App rule board that is used by all of our org?

adamkvitek: Yes, there are workarounds though how to apply only specific rules for us.

.github/workflows/semgrep.yml
adamkvitek: True. Editing.

adamkvitek:
@pabloest should we rather create another org in App only for our docs? I feel that our generic mode rules checking for language and Markdown/HTML stylistics may create a lot of noise in other repositories bound to the current org.

adamkvitek:
Agreed with @pabloest - we will rather create our local YAML file with specific rules - so not all r2c organization rules are applied to us (and vice versa).

adamkvitek:
See: https://github.com/returntocorp/semgrep-docs/pull/618#issuecomment-1234476100

PR #633, opened 2022-07-25
change(writing-rules): clarify pattern-inside example

brandonspark:
**What:**
Currently, the `pattern-inside` example isn't super insightful into what `pattern-inside` actually changes.

**Why**:
This is because it shows off an example where the `pattern` would match regardless of whether the `pattern-inside` rule were to be there. I think the strength of `pattern-inside` is in showing what examples _should not_ match, if a user were to use it. For reference, this confused me over why I should use `pattern-inside`, or care about it.

**How**:
I added a `def foo` which has a `return` which is unmatched, as a result of not being in a `class`.

### Please ensure

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -240,7 +240,7 @@ The `pattern-not` operator is the opposite of the `pattern` operator. It finds c
 
 The `pattern-inside` operator keeps matched findings that reside within its expression. This is useful for finding code inside other pieces of code like functions or if blocks.
 
-<iframe src="https://semgrep.dev/embed/editor?snippet=B4lR" border="0" frameBorder="0" width="100%" height="432"></iframe>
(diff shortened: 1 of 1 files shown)


adamkvitek (approved):
Thanks a lot @brandonspark!

brandonspark:
Is it problematic that this `Check Markdown links` thing is failing?

adamkvitek:
@brandonspark it should not be a major problem (meaning you can ignore it).

@underyx I am not sure how to substitute the dead link to `rule_schema.yaml` on rule_syntax.md document - checked https://github.com/returntocorp/semgrep/search?q=rule_schema but `/semgrep/semgrep_interfaces/rule_schema.yaml` seems to be a submodule now, which is not public as it seems? I am not sure where to find `rule_schema.yaml` now in our repository. Should I remove the link or do we have any substitute for it?

PR #639, opened 2022-07-28
disable ligatures

s-santillan:

### Please ensure

- [ ] Colleague reviews it

For compatibility prefer ligatures off.

Preview


ievans (approved):
I feel like the default ought to be that ligatures are disabled in code blocks but I understand that's out of our hands

PR #645, opened 2022-08-04
Update docs to point to wcmatch

devd:
pathlib doesn't match `**/*` but wcmatch (which semgrep uses) does so update links and add an example to clarify


adamkvitek:
Thank you a lot @devd!

brendongo (approved):

adamkvitek (approved):

PR #649, opened 2022-08-05
focus-metavariable is now considered as GA, add evaluation strategy

IagoAbal:
Plus, explain how `patterns` is evaluated, which is key to writing rules
but most users do not know about.

Closes PA-1683

### Please ensure


IagoAbal:
Maybe we could add a redirect from https://semgrep.dev/docs/experiments/focus-metavariable/ to https://semgrep.dev/docs/writing-rules/rule-syntax/#focus-metavariable.

docs/writing-rules/rule-syntax.md
```diff
@@ -81,6 +81,17 @@ The `patterns` operator performs a logical AND operation on one or more child pa
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=Q83q" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
+### Evaluation strategy
```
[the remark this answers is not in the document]
adamkvitek: I am also not sure if by use of "we" below, you mean we as Semgrep or "we" as if we are describing a procedure together with a user.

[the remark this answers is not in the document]
adamkvitek: Understood. A funnel, internal logic of how Semgrep evaluates rules. Positive patterns are the strongest, negative patterns are filters narrowing down the scope of these patterns. Thank you! I will edit this more soon. @IagoAbal

docs/writing-rules/rule-syntax.md
```diff
@@ -121,6 +132,28 @@ This operator will filter findings that have _any overlap_ with the supplied reg
...
+
+The `focus-metavariable` operator puts the focus, or _zooms in_, on the code region matched by a metavariable. For example, to find all functions arguments annotated with the type `bad` you may write the following pattern:
+
+```yaml
+pattern: |
+  def $FUNC(..., $ARG : bad, ...):
+    ...
+```
+
+This works but it matches the entire function definition. Sometimes, this is not desirable. In particular, if you are using [Semgrep App](https://semgrep.dev/login) and you have triaged a finding generated by this pattern, the same finding will show up again as new if you make any change to the definition of the function!
```
adamkvitek, line 145: @IagoAbal Just to specify why this is beneficial we can add more explanation here. Let me know if what I wrote is correct... feel free to edit my suggestions.

```suggestion
This works but it matches the entire function definition. Sometimes, this is not desirable. If the definition spans hundreds of lines, they are all matched. In particular, if you are using [Semgrep App](https://semgrep.dev/login) and you have triaged a finding generated by this pattern, the same finding shows up again as new if you make any change to the definition of the function!
```

docs/writing-rules/rule-syntax.md
adamkvitek:
```suggestion
### Semgrep rule evaluation strategy
```

docs/writing-rules/rule-syntax.md
```diff
@@ -81,6 +81,17 @@ The `patterns` operator performs a logical AND operation on one or more child pa
...
+2. We evaluate all _negative_ patterns, that is [`pattern-not-inside`](#pattern-not-inside)s, [`pattern-not`](#pattern-not)s, and [`pattern-not-regex`](#pattern-not-regex)es. This gives us a set of _negative ranges_ which are used to filter the positive ranges. This results in a strict subset of the positive ranges computed in the previous step.
+3. We evaluate all _conditionals_, that is [`metavariable-regex`](#metavariable-regex)es, [`metavariable-pattern`](#metavariable-pattern)s, and [`metavariable-comparison`](#metavariable-comparison)s. These conditional operators can only examine the metavariables bound in the positive ranges in step 1, that passed through the filter of negative patterns in step 2. Note that metavariables bound by negative patterns are _not_ available here.
```
adamkvitek, line 90:
```suggestion
3. Semgrep evaluates all _conditionals_, that is [`metavariable-regex`](#metavariable-regex)es, [`metavariable-pattern`](#metavariable-pattern)s, and [`metavariable-comparison`](#metavariable-comparison)s. These conditional operators can only examine the metavariables bound in the positive ranges in step 1, that passed through the filter of negative patterns in step 2. Note that metavariables bound by negative patterns are _not_ available here.
```

docs/writing-rules/rule-syntax.md
```diff
@@ -81,6 +81,17 @@ The `patterns` operator performs a logical AND operation on one or more child pa
...
+4. We apply all [`focus-metavariable`](#focus-metavariable)s, by computing the intersection of each positive range with the range of the metavariable on which we want to focus. Again, the only metavariables available to focus on are those bound by positive patterns.
```
adamkvitek, line 91:
```suggestion
4. Semgrep applies all [`focus-metavariable`](#focus-metavariable)s, by computing the intersection of each positive range with the range of the metavariable on which we want to focus. Again, the only metavariables available to focus on are those bound by positive patterns.
```

docs/writing-rules/rule-syntax.md
```diff
@@ -81,6 +81,17 @@ The `patterns` operator performs a logical AND operation on one or more child pa
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=Q83q" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
+### Evaluation strategy
+
+Note that the order in which the child patterns are declared in a `patterns` operator has no effect on the final result. A `patterns` operator is always evaluated in the same way:
```
adamkvitek, line 86: @IagoAbal expanding on your sentence, let me know if this is true, please. Does it mean that regardless of how a user defines these patterns in a rule, the order of patterns declared in a rule is arbitrary (for example a rule where `pattern-not` is declared before `patterns`)? Semgrep always evaluates `patterns` in the order you mentioned in these steps, not as the user defines them. Is this correct?

```suggestion
This section documents how Semgrep evaluates rules. This is similar to a funnel where some `patterns` take precedent over other `patterns` when Semgrep is processing the rule. Note that the order in which the child patterns are declared in a `patterns` operator of a rule has no effect on the final result. A `patterns` operator is always evaluated in the same way:
```

docs/writing-rules/rule-syntax.md
adamkvitek: WIP title. Open to suggestions.

IagoAbal: It's already nested under `patterns` so I'm not sure we need the extra long title here. It also sounds a bit strange to me to say "evaluation strategy of patterns", wouldn't it be "for" rather than "of", and we should use `patterns` rather than just "patterns", because it refers to the `patterns` operator in particular.

docs/writing-rules/rule-syntax.md
IagoAbal: I feel you're merging "patterns" (as in any pattern operator) and `patterns` (which is a very specific AND operator).

IagoAbal: The section doesn't really describe how we evaluate any arbitrary rule, but only how we evaluate the `patterns` AND operator.

docs/writing-rules/rule-syntax.md
```diff
@@ -81,6 +81,17 @@ The `patterns` operator performs a logical AND operation on one or more child pa
...
+
+1. We evaluate all _positive_ patterns, that is [`pattern-inside`](#pattern-inside)s, [`pattern`](#pattern)s, [`pattern-regex`](#pattern-regex)es, and [`pattern-either`](#pattern-either)s. Each range matched by each one of these patterns is intersected with the ranges matched by the other operators. The result is a set of _positive ranges. The positive ranges carry _metavariable bindings_, for example, in one range `$X` may be bound to the function call `foo()`, and in another range `$X` may be found to the expression `a + b`.
```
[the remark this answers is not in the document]
IagoAbal: If I write `pattern` then I mean `pattern` (the leaf pattern operator) and nothing else.

docs/writing-rules/rule-syntax.md
aryx: What is funnel? Also not sure about the word "strongest".

docs/writing-rules/rule-syntax.md
```diff
@@ -81,6 +81,17 @@ The `patterns` operator performs a logical AND operation on one or more child pa
...
 
+### Evaluation strategy
+
+Note that the order in which the child patterns are declared in a `patterns` operator has no effect on the final result. A `patterns` operator is always evaluated in the same way:
+
+1. Semgrep evaluates all _positive_ patterns, that is [`pattern-inside`](#pattern-inside)s, [`pattern`](#pattern)s, [`pattern-regex`](#pattern-regex)es, and [`pattern-either`](#pattern-either)s. Each range matched by each one of these patterns is intersected with the ranges matched by the other operators. The result is a set of _positive_ ranges. The positive ranges carry _metavariable bindings_. For example, in one range `$X` can be bound to the function call `foo()`, and in another range `$X` can be bound to the expression `a + b`.
+2. Semgrep evaluates all _negative_ patterns, that is [`pattern-not-inside`](#pattern-not-inside)s, [`pattern-not`](#pattern-not)s, and [`pattern-not-regex`](#pattern-not-regex)es. This gives a set of _negative ranges_ which are used to filter the positive ranges. This results in a strict subset of the positive ranges computed in the previous step.
```
aryx, line 89: In theory we should also talk about how the metavariable bindings interact when a metavar is mentioned both in the negative and positive patterns.

aryx (approved):
Seems like a net improvement over what we had.
We can still improve in further PRs.

docs/writing-rules/rule-syntax.md
adamkvitek: Ah, get it now, thanks. It was not heading level 4 (`####`) which is confusing.

docs/writing-rules/rule-syntax.md
adamkvitek: @aryx can you submit this in another PR maybe? Sounds interesting. Thanks.

PR #653, opened 2022-08-05
Update metavariable-comparison docs

IagoAbal:

### Please ensure

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -159,7 +159,7 @@ It is possible to nest `metavariable-pattern` inside `metavariable-pattern`!
 The metavariable should be bound to an expression, a statement, or a list of statements, for this test to be meaningful. A metavariable bound to a list of function arguments, a type, or a pattern, will always evaluate to false.
 :::
 
-### Nested language
+#### `metavariable-pattern` with nested language
 
 If the metavariable's content is a string, then it is possible to use `metavariable-pattern` to match this string as code by specifying the target language via the `language` key.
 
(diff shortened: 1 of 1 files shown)


docs/writing-rules/rule-syntax.md
```diff
@@ -187,6 +187,29 @@ Comparison expressions support simple arithmetic as well as composition with [bo
 
 Building off of the previous example this will still catch code like `set_port(80)` but will no longer catch `set_port(443)` or `set_port(8080)`.
```
adamkvitek, line 188: - Avoid rarely use phrasal verbs (generally avoid all phrasal verbs and idioms unless there is no better word) - https://english.stackexchange.com/questions/102057/what-is-the-meaning-of-build-off-of and see phrasal verbs in this section: https://developers.google.com/style/translation?hl=en#clear-language
- Avoid "will" unless you are talking about a specific time in the future, use present tense for most of the docs. See https://developers.google.com/style/tense?hl=en for more information.
- When using words such as "this" it is better to specify what we mean by it. I guess that here "this" means the rule (looking at the example given above).

- Optional - feel free to remove my suggestions in brackets (`(it is not divisible by two)` and `(the number is larger than 1024)` if you find them misleading. Stating the obvious might help.

```suggestion
Building on the previous example, this rule pattern matches code such as `set_port(80)` but does not match `set_port(443)` (it is not divisible by two) or `set_port(8080)` (the number is larger than 1024).
```

docs/writing-rules/rule-syntax.md
```diff
@@ -187,6 +187,29 @@ Comparison expressions support simple arithmetic as well as composition with [bo
...
+- Boolean, string, integer, and float literals.
+- Boolean operators `not`, `or`, and `and`.
+- Arithmetic operators `+`, `-`, `*`, `/`, and `%`.
+- Comparison operators `==`, `!=`, `<`, `<=`, `>`, and `>=`.
+- Function `int()` to convert strings into integers.
+- Function `str()` to convert numbers into strings.
+- Lists and the `in` infix operator.
+- Function `re.match()` to match a regular expression (without the optional `flags` argument).
+
+And, of course, you can use Semgrep metavariables like `$MVAR`, which are evaluated as follows:
```
adamkvitek, line 201: - @IagoAbal do we have $MVAR documented somewhere? I found a few mentions in release notes, but we do not have anything more. This brings me to another question: Do you think that it makes sense to create an **exhaustive table for all metavariables** (like $FUNC, $ARG) we support or what metavariables to use is obvious for developers? (Can I find all of these metavariables somewhere in our source code to figure it out myself?)
- Please, can you make an example in Playground for use of  `$MVAR` as described below?

```suggestion
You can use Semgrep metavariables such as `$MVAR`, which Semgrep evaluates as follows:
```

adamkvitek (commented):
@IagoAbal thank you as always for your contribution! I left some suggestions and two questions.

docs/writing-rules/rule-syntax.md
adamkvitek: @IagoAbal crossed the text above, understood, metavariables are similar to variables in bash (with $ sign) - I can use anything I want ($X), it is to make the rule filled with meaningful text for people reading the rule, but it does not affect the behavior of Semgrep whether we use $X or $FUNC.

s-santillan: @adamkvitek As discussed - keep $MVAR. All Iago has to do is accept the suggestions. I don't see anything that changes the meaning too much.

adamkvitek: Yes, that is what I tried to achieve, so @IagoAbal can only apply suggestions.

docs/writing-rules/rule-syntax.md
```diff
@@ -159,7 +159,7 @@ It is possible to nest `metavariable-pattern` inside `metavariable-pattern`!
 The metavariable should be bound to an expression, a statement, or a list of statements, for this test to be meaningful. A metavariable bound to a list of function arguments, a type, or a pattern, will always evaluate to false.
 :::
 
-### Nested language
+#### `metavariable-pattern` with nested language
 
 If the metavariable's content is a string, then it is possible to use `metavariable-pattern` to match this string as code by specifying the target language via the `language` key.
```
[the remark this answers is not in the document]
IagoAbal: My PR does not touch this paragraph so I don't think we should be discussing it here. Also, I think it's missing a "the", shouldn't it be "If the content of THE metavariable is ...".

docs/writing-rules/rule-syntax.md
IagoAbal: Again, I did not touch this and I did not even write this. You should open a separate PR for this.

aryx (approved):

aryx:
I agree with iago, reviews should only concern modified text. But adam feel free to open a separate PRs to fix the english
in the sentences you noted!

PR #654, opened 2022-08-05
Add redirect and delete stale files to fix taint search bug

pabloest:

### Please ensure

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ jobs:
           command: yarn build
       - run:
           name: Yarn Deploy Staging
-          command: aws s3 sync build/ s3://staging-semgrep-dev/docs
+          command: aws s3 sync build/ s3://staging-semgrep-dev/docs --delete
   deployProd:
     docker:
       - image: cimg/node:16.14.2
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -58,7 +58,7 @@ jobs:
           command: yarn build
       - run:
           name: Yarn Prod Deploy
-          command: aws s3 sync build/ s3://semgrep-docs/docs
+          command: aws s3 sync build/ s3://semgrep-docs/docs --delete
```
efortuna, line 61: I'm guessing this would only need to be run once, and then we can go back to the original, non-delete version?

.circleci/config.yml
```diff
@@ -39,7 +39,7 @@ jobs:
           command: yarn build
       - run:
           name: Yarn Deploy Staging
-          command: aws s3 sync build/ s3://staging-semgrep-dev/docs
+          command: aws s3 sync build/ s3://staging-semgrep-dev/docs --delete
```
efortuna, line 42: I have not run this myself, but my instinct says we wouldn't need to do this for staging too...

.circleci/config.yml
pabloest: Adding this flag causes `aws s3` to look for local files that are no longer present on the destination, so we’d want to keep it in moving forward. That’ll prevent stale files from remaining in the destination indefinitely.

adamkvitek (approved):

.circleci/config.yml
pabloest: We barely use that staging bucket so I think it doesn’t matter much. I added it here mostly for consistency between staging and the other command.

PR #657, opened 2022-08-09
Experiment with flex dl lists

adamkvitek:

### Please ensure


src/css/custom.css
```diff
@@ -151,16 +151,43 @@ h5 {
...
+  flex-flow: row wrap;
+}
 dt {
-    font-weight:700;
+  flex-basis: 20%;
+  padding: 2px 8px;
+  text-align: left;
+  font-weight: 700;
+}
+dd {
```
s-santillan, line 166:
```suggestion
dd .noun-definition {
```

Try something like `.noun-definition`  then apply the class

adamkvitek:
This works in:
Screenshot 2022-08-15 at 17 55 22


But does not work:
Screenshot 2022-08-15 at 17 54 39

s-santillan:
Created `.description-list__table` class so that it's not global. Apply class as needed. See the screenshot's top list (default layout) and bottom (new layout).

However the top/bottom spacing (i.e. padding) for `.description-list__table` needs work. @adamboros @adamkvitek are there particular docs examples you like?

adamboros:
I prefer the 'additional scan parameters include' one because longer sentences won't break the layout there and it's easy to follow.

adamkvitek:
@adamboros `description_list_table` class (the example with Monitor, Comment, Block) is for short description list entries (few words maximum) - so this will not be a default setup for all description lists. The default description list styling will still be 'additional scan parameters include' (top example). The purpose is to have two styles for different types of description lists.

adamkvitek:
@s-santillan I was playing with the code given in [this article](https://www.the-art-of-web.com/css/format-dl/). Especially here (we can reuse some of the padding and formatting showcased in the article). 
Screenshot 2022-08-29 at 12 43 41

PR #658, opened 2022-08-09
(nano-sized) update instructions for SCM

s-santillan:

### Please ensure

(n/a) A technical writer reviews the content or PR

Removed step 1 because the feature is generally available (no longer behind a feature flag).


adamkvitek (approved):
@s-santillan just quickly checked, LGTM.

PR #661, opened 2022-08-12
fix: documentation around diff-aware scanning

chmccreery:
@s-santillan can you help me make sure this looks ok and add hyperlinks to the main diff-aware page every place we mention diff-aware scanning (should be impossible to miss if you are trying to figure out how to run semgrep in CI, especially if you are using semgrep-app).

### Please ensure


docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
 
 ## Diff-aware scanning (`SEMGREP_BASELINE_REF`)
 
-For [diff-aware scans](overview.md#features), this option filters scan results to those introduced after the git commit, in a branch, or tag. For example, you have a repository with 10 commits. You set the commit number 8 as the baseline. Consequently, Semgrep only returns scan results introduced by changes in commits 9 and 10.
+There are two ways to run a semgrep scan: as a full scan or as a diff-aware scan.
```
s-santillan, line 72:
```suggestion
There are two ways to run a semgrep scan:
```

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+
+A full scan will scan your entire code base and report every finding in the code base. We recommend doing a full scan of your `main` branch at a regular cadence, such as every night or every week, depending on your own needs. This will ensure that Semgrep App has a full list of all findings in your code base, regardless of when they were introduced. To run a full scan, simply run `semgrep ci` without setting `SEMGREP_BASELINE_REF` environment variabel.
```
s-santillan, line 74:
```suggestion
<dl><dt>Full scan</dt>
<dd>A full scan will scan your entire codebase and report every finding in the codebase. It is recommended to perform a full scan of your `main` branch at a regular cadence, such as every night or every week, depending on your own needs. This ensures that Semgrep App has a full list of all findings in your code base, regardless of when they were introduced. To run a full scan, simply run `semgrep ci` without setting `SEMGREP_BASELINE_REF` environment variable.
```

For @chmccreery : I just converted the paragraphs into a "description list" (<dl> tag).

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+
+A diff-aware scan will scan your code before and after some "baseline" and only report findings that are newly introduced in the commits after that baseline. For example, imagine you have a repository with 10 commits and you set commit number 8 as the baseline. Consequently, Semgrep only returns scan results introduced by changes in commits 9 and 10. This is how `semgrep ci` should be run in pull requests and merge requests, since it will report only the findings that are created by those code changes. To run a diff-aware scan, use `SEMGREP_BASELINE_REF=<ref> semgrep ci` where ref can be a commit hash, branch name, or other git ref.
```
s-santillan, line 76:
```suggestion
<dt>Diff-aware scan</dt>
<dd>A diff-aware scan runs on your code before and after some "baseline" and only reports findings that are newly introduced in the commits after that baseline. For example, imagine you have a repository with 10 commits and you set commit number 8 as the baseline. Consequently, Semgrep only returns scan results introduced by changes in commits 9 and 10. This is how `semgrep ci` should be run in pull requests and merge requests, since it will report only the findings that are created by those code changes. To run a diff-aware scan, use `SEMGREP_BASELINE_REF=<ref> semgrep ci` where ref can be a commit hash, branch name, or other git ref.
</dd></dl>
```

docs/semgrep-ci/overview.md
```diff
@@ -71,7 +71,13 @@ To add Semgrep CI to any CI environment, use the [`returntocorp/semgrep`](https:
...
 
 ```sh
-semgrep scan --config auto --config <other rule or rulesets>
+semgrep ci
+```
+
+For diff-aware scans:
+
+```sh
+SEMGREP_BASELINE_REF=<ref> semgrep ci
```
s-santillan, line 80:
```suggestion
<pre class="language-bash">
<code>SEMGREP_BASELINE_REF=<span classname="placeholder">REF</span></code>
semgrep ci
</pre>
```
We use a specific "placeholder" format for variables users should change ("ref" in this case). Just making it match our conventions.

docs/semgrep-ci/overview.md
```diff
@@ -71,7 +71,13 @@ To add Semgrep CI to any CI environment, use the [`returntocorp/semgrep`](https:
 For full project scans:
 
 ```sh
-semgrep scan --config auto --config <other rule or rulesets>
+semgrep ci
+```
+
+For diff-aware scans:
+
+```sh
```
s-santillan, line 79:
```suggestion

```

docs/semgrep-ci/overview.md
```diff
@@ -71,7 +71,13 @@ To add Semgrep CI to any CI environment, use the [`returntocorp/semgrep`](https:
...
+SEMGREP_BASELINE_REF=<ref> semgrep ci
 ```
```
s-santillan, line 81:
```suggestion

```
Removing extraneous backticks (pre tag handles it)

s-santillan:
@chmccreery Done with my review! Thank you, I mostly just removed extra words and formatted stuff.

There was something you said about diff scans vs full scans that surprised me.

These docs say that the presence of `SEMGREP_BASELINE_REF` determines if something is a full scan job or diff job. My understanding is that for GitHub Action (GHA) and GitLab CICD (GL CICD), we don't set SEMGREP_BASELINE_REF and need only one CI job (`semgrep.yml` for example) to trigger diff aware for PRs and full scan for trunkline. However, as far as I know, only GHA and GL CICD behave this way.

That means that for other CI providers, like jenkins, you'd have to make 1 job for diff-aware scans (only for PR/MRs) and 1 job for full scans (scheduled) since the parameters have to be different.

However, that's not what we do when we make CI jobs in Semgrep App for those other providers. We only have them add 1 snippet... so what does that mean? We're always only doing full scans on App, or only doing diff scans? 🤯🤯🤯

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+
+:::note
```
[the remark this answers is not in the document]
chmccreery: They are actually two separate notes. One about only doing full scans for the main branch, and one about never doing full scans on anything but the main branch. They are related, but feel like two different misconfigurations with different repercussions which is why I separated them out. Thoughts?

chmccreery:
> That means that for other CI providers, like jenkins, you'd have to make 1 job for diff-aware scans (only for PR/MRs) and 1 job for full scans (scheduled) since the parameters have to be different.

Yeah, I just checked with Brendon and Sabrina and we all are pretty sure that this is the current state of things. Users either need two jobs or they need to use an "if" condition to set the environment variable only on PR. We also need to work on updating the default config files that are provided by the app during onboarding to follow the advice I've outlined here, but in the meantime we can at least fix the docs :) We will need this "philosophy of how to scan" in the docs even if the default configs are right, so that users of new CI providers can figure things out and also users who go rogue can understand what they shouldn't change. 

>  We're always only doing full scans on App, or only doing diff scans

I think we're currently suggesting something that ends up doing full scans which is leading to lots of performance challenges on the app

docs/semgrep-ci/configuration-reference.md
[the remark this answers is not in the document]
s-santillan: I merged the notes into a list. It was a little hard to set this as a suggestion (because I was also fixing the tags) so I committed it. Resolving this now.

s-santillan (approved):
@chmccreery I built the docs locally to check and corrected some of my incorrect tags/classnames. This is ready for merging! Thank you for contributing to the docs!

adamkvitek:
This works in:
Screenshot 2022-08-15 at 17 55 22


But does not work:
Screenshot 2022-08-15 at 17 54 39

chmccreery:
@adamkvitek did you leave a comment and then delete it? I got an email about a comment I don't see here. @s-santillan can I ask you to carry this over the finish line and address the issue Adam found (if relevant) and then just click merge when you think it's good to go?

s-santillan:
@chmccreery No prob!

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
 
 ## Diff-aware scanning (`SEMGREP_BASELINE_REF`)
 
-For [diff-aware scans](overview.md#features), this option filters scan results to those introduced after the git commit, in a branch, or tag. For example, you have a repository with 10 commits. You set the commit number 8 as the baseline. Consequently, Semgrep only returns scan results introduced by changes in commits 9 and 10.
+Semgrep scans can be classified by scope. The scope of a scan refers to what lines of code are scanned in a codebase. When classifying scans by scope, there are two types of scans:
```
s-santillan, line 72: @adamkvitek This is the only line I've changed since our call since you brought up the ambiguity of "types of scan". Let me know if "scope" is better.

adamkvitek: LGTM!

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+
+<dl>
+    <dt>Full scan</dt>
+    <dd>A full scan will scan your entire codebase and report every finding in the codebase. It is recommended to perform a full scan of your <code>main</code> branch at a regular cadence, such as every night or every week, depending on your own needs. This ensures that Semgrep App has a full list of all findings in your code base, regardless of when they were introduced. To run a full scan, simply run <code>semgrep ci</code> without setting <code>SEMGREP_BASELINE_REF</code> environment variable.</dd>
```
adamkvitek, line 76: - Do not use will in technical writing. Only use will when you are referring to the time that needs to elapse between action and goal. See https://developers.google.com/style/tense?hl=en.
- Since I removed `will`, I am trying to avoid `scan scans` in the first sentence.

```suggestion
    <dd><p>A full scan is scanning your entire codebase and reports every finding in the codebase. It is recommended to perform a full scan of your <code>main</code> branch at a regular cadence, such as every night or every week. This ensures that Semgrep App has a full list of all findings in your code base, regardless of when they were introduced. To run a full scan, run <code>semgrep ci</code> without setting <code>SEMGREP_BASELINE_REF</code> environment variable.</p></dd>
```

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+    <dt>Diff-aware scan</dt>
+    <dd>A diff-aware scan runs on your code before and after some "baseline" and only reports findings that are newly introduced in the commits after that baseline. For example, imagine you have a repository with 10 commits and you set commit number 8 as the baseline. Consequently, Semgrep only returns scan results introduced by changes in commits 9 and 10. This is how <code>semgrep ci</code> should be run in pull requests and merge requests, since it will report only the findings that are created by those code changes. To run a diff-aware scan, use <code>SEMGREP_BASELINE_REF=<span class="placeholder">REF</span> semgrep ci</code> where <span class="placeholder">REF</span> can be a commit hash, branch name, or other git ref.</dd>
```
adamkvitek, line 78: - @chmccreery Ad "should" - `This is how <code>semgrep ci</code> should be run in pull requests and merge requests` - do we want to say: 
A. This is recommended setup for Semgrep in CI.
B. This is how Semgrep should work in CI, but it often does not (uncertainty implied by modal words such as "should").
Avoid these words and rather specify. See https://developers.google.com/style/word-list?hl=en#should.
- @s-santillan I am not sure if we can use `semgrep ci` - some time ago we discussed this with @pabloest that we should rather state Semgrep in CI. I am not sure if there was any additional discussion made that I was not part of regarding this topic.
- git refs are probably [git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References).


```suggestion
    <dd><p>A diff-aware scan runs on your code before and after some "baseline" and only reports findings that are newly introduced in the commits after that baseline.</p>
    <p>For example, imagine a hypothetical repository with 10 commits. You set commit number 8 as the baseline. Consequently, Semgrep only returns scan results introduced by changes in commits 9 and 10. This is how Semgrep in CI can run in pull requests and merge requests, since it reports only the findings that are created by those code changes. To run a diff-aware scan, use <code>SEMGREP_BASELINE_REF=<span class="placeholder">REF</span> semgrep ci</code> where <span class="placeholder">REF</span> can be a commit hash, branch name, or other git reference.</p></dd>
```

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+</dl>
 
 :::note
-It is recommended to perform baseline scans on other branches than your `main` branch. The Semgrep App keeps track of which findings have been fixed on a given branch. If you configure baseline scans on your main branch, and compare the last commit to the penultimate commit, Semgrep wrongly considers all findings as fixed. In this case, Semgrep only reports findings that appear in the last commit.
+* Do not perform diff-aware scans on your `main` branch. The Semgrep App keeps track of which findings have been fixed on a given branch. If you configure diff scans on your main branch, and compare the last commit to the penultimate commit, Semgrep wrongly considers all findings from before the penultimate commit to be fixed.
```
adamkvitek, line 82:
```suggestion
* Do **not** perform diff-aware scans on your `main` branch. The Semgrep App keeps track of which findings have been fixed on a given branch. If you configure diff scans on your main branch, and compare the last commit to the penultimate commit, Semgrep wrongly considers all findings from before the penultimate commit to be fixed.
```

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
+* Do not perform full scans on non-mainline or non-trunk branches. Performing full scans on every branch will slow down your CI jobs, show developers findings they didn't introduce, and result in many duplicated findings in the Semgrep App, resulting in poorer experience there as well.
```
adamkvitek, line 83: - I assume that `show developers findings they didn't introduce` means `displays findings that developers did not introduce` - the meaning was a bit confusing.
- Do not use words such as "show", use display instead when possible.
- Make the not bold when you want to strongly discourage from doing something (otherwise try to frame it positively).

```suggestion
* Do **not** perform full scans on non-mainline or non-trunk branches. Performing full scans on every branch slows down your CI jobs, displays findings that developers did not introduce, and results in many duplicated findings in the Semgrep App, resulting in a poorer experience.
```

docs/semgrep-ci/configuration-reference.md
```diff
@@ -69,17 +69,25 @@ To find more details about some of these configuration options, see the followin
...
 :::
  
 ### Examples of `SEMGREP_BASELINE_REF`
 
 To only report findings newly added
 since branching off from your `main` branch, set the following:
-<pre class="language-bash"><code>SEMGREP_BASELINE_REF=<span className="placeholder">TOPIC-BRANCH-NAME</span></code></pre>
+<pre class="language-bash"><code>SEMGREP_BASELINE_REF=<span className="placeholder">main</span></code></pre>
```
adamkvitek, line 90: Consider making the placeholder upper case as per https://developers.google.com/style/placeholders.

docs/semgrep-ci/configuration-reference.md
adamkvitek: Another note - try to build this, if the output is satisfactory.

adamkvitek (commented):
@s-santillan just a few points made and suggested:
- Avoid will in technical writing. 🙏 https://developers.google.com/style/tense?hl=en
- Avoid should and other modals. 🙏 https://developers.google.com/style/word-list?hl=en#should

docs/semgrep-ci/configuration-reference.md
s-santillan: Honestly, "is scanning" is not better than "will scan".

I think will is fine because this is not a procedure. If the user chooses to use a full scan, then they will scan their codebase. Does that make sense?

docs/semgrep-ci/configuration-reference.md
s-santillan: @adamkvitek Semgrep CI in this context is literally a command, `semgrep ci`.

docs/semgrep-ci/configuration-reference.md
adamkvitek: One of the reasons why we do this is that some engineers are using automatic translations of pages and some languages do not have future tense (so will cause a bit of havoc). @s-santillan See Google style guide suggestion "However, the fact that the reader will be writing and running code in the future isn't a good reason to use future tense. Stick with present tense where possible, as shown in the previous Recommended example." From my experience, TWs are only allowed to use "will" really when you are talking about things such as "The hedgehog will hibernate starting in October." (actual specific future). It might be ok in this case, but I would avoid it here also. I will leave this decision up to you. (Please stick with the division to paragraphs I suggest.)

docs/semgrep-ci/configuration-reference.md
s-santillan: I'm accepting the suggestion but reverting "Semgrep in CI" to `semgrep ci`.

docs/semgrep-ci/configuration-reference.md
s-santillan: I think I'll go with "A full scan runs on your entire codebase..."

PR #668, opened 2022-08-18
Add instructions for adding a new package to the CLI

chmccreery:

### Please ensure

### DIFF
--- a/docs/contributing/semgrep-contributing.md
+++ b/docs/contributing/semgrep-contributing.md
@@ -129,9 +129,9 @@ See the Makefile in `cli/`
 
 ## Adding python packages to `semgrep`
 
-While developing you may find that you would like to add a package as a dependency. You can do so with pipenv. For example, suppose you would like to add the package `pyyaml` to Semgrep.
+While developing you may find that you would like to add a package dependency. Semgrep uses mypy to do static type-checking of its python code, so when you add a new python package, you will also need to add typing stubs for that package. You can add the package you want along with its typing package in three steps. For example, suppose you would like to add the package `pyyaml` to Semgrep.
 
-Semgrep uses mypy to do static type-checking of its python code. This means that when you add a new package, you will also need to add typing stubs for that package or else pre-commit and your IDE will likely complain, as they will be unable to do any type-checking for the functions in the imported package. To fix this, you will need the corresponding package with typing stubs. For our `pyyaml` example, the corresponding package is `types-pyyaml`.
+First, you will need to install the corresponding package with typing stubs. For our `pyyaml` example, the corresponding package is `types-pyyaml`.
 
 ```
 pipenv install --dev types-pyyaml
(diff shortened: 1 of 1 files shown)


s-santillan (approved):

docs/contributing/semgrep-contributing.md
```diff
@@ -127,6 +127,38 @@ make rebuild
...
 See the Makefile in `cli/`
 
+## Adding python packages to `semgrep`
+
+While developing you may find that you would like to add a package dependency. Semgrep uses mypy to do static type-checking of its python code, so when you add a new python package, you will also need to add typing stubs for that package. You can add the package you want along with its typing package in three steps. For example, suppose you would like to add the package `pyyaml` to Semgrep.
+
+First, you will need to install the corresponding package with typing stubs. For our `pyyaml` example, the corresponding package is `types-pyyaml`.
+
+```
+pipenv install --dev types-pyyaml
```
brendongo, line 137: You don't actually need to install types-pyyaml in pipenv. Installing it in the pre-commit config is what matters for mypy

docs/contributing/semgrep-contributing.md
```diff
@@ -127,6 +127,38 @@ make rebuild
...
+Finally, you will need to manually add the original package to `cli/setup.py` in the `install_requires` list variable. You can find the version number either in the `Pipfile.lock` changes or by looking up online the most recent major version of the package.
+
+```
+install_requires = [
+   ...
+   "pyyaml~=6.0",
+]
+```
+
+This change allows docker to find the package when building. Without this change, if you create a pull request you will see a CI job called `build docker image` fail with a `ModuleNotFoundError`, indicating that it is unable to find your package.
```
brendongo, line 160: This isn't just to allow docker to find the package while building, this is so the package is a dependency of published semgrep

adamkvitek (approved):
Hello @chmccreery, I will be reworking this documentation, approving to let this go quickly, but some general stylistic rules for docs should be:

1. Do not use the future tense (only if you mention that something will be done in a specific time. For example: the API will sunset in January 2025.
2. Do not use modal verbs - would, may, might - be strict and directive, or descriptive.

docs/contributing/semgrep-contributing.md
chmccreery: @brendongo isn't pipenv install needed for something, though? Like the code won't run locally until I do the pipenv install of either types-pyyaml or just pyyaml, and I thought you asked me to remove the pyyaml one from my PR so now I'm confused. Can you clarify what we should do, or jump on a call if it's easier to explain that way?

chmccreery:
@adamkvitek thanks for the feedback - I have updated to avoid future tense and modal verbs

docs/contributing/semgrep-contributing.md
chmccreery: ok, fixed

adamkvitek:
Would be great to get some testing VM with compiled Semgrep code so TWs can test this, but we can try to this later. Thanks! @chmccreery

docs/contributing/semgrep-contributing.md
chmccreery: Resolved offline

PR #676, opened 2022-08-24
Add a new section "explaining" code and update changelog management.

mjambon:
This goes with the [updated semgrep PR template](https://github.com/returntocorp/semgrep/pull/5962).

### Please ensure

### DIFF
--- a/docs/contributing/contributing-code.md
+++ b/docs/contributing/contributing-code.md
@@ -173,7 +173,7 @@ Here's a short example:
 *)
 ```
 
-It could be improved by explaining a bit about the code's uses:
+It can be improved by explaining the code's uses:
 ```
 (*
    Generate unique names with a given prefix. This is used to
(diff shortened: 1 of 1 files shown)


docs/contributing/contributing-code.md
```diff
@@ -149,26 +149,80 @@ Once `pre-commit` is working you may commit code and create pull requests as
...
+
+Here's a short example:
+```
+(*
+   Generate unique names with a given prefix.
+*)
+```
 
-A tool called `towncrier` is used for changelog management. At the root of the Semgrep repository, there is a directory `changelog.d`, which contains individual changelog entries, each in their own file. An example of such a file is contained in the repository. See also the description of the format below.
+It could be improved by explaining a bit about the code's uses:
```
s-santillan, line 176:
```suggestion
It can be improved by explaining the code's uses:
```

mjambon:
> So this is why the release notes in Github have been getting more descriptive!

I'm not sure :sweat_smile:... Maybe when faced with an empty file instead of a bullet point, the developer feels the urge to fill it with multiple lines of text :smiley: 

> FYI, we are trying out a **1-shot review** process which Bence explained is used by the App team for their code reviews. In this process we approve the PR in most cases and simply leave suggestions that you can choose to merge or not.

sounds good to me!

PR #685, opened 2022-08-27
Update generic examples + how to ignore comments

mjambon:
I wasn't sure about the internal link so I used the full URL https://semgrep.dev/docs/writing-rules/rule-syntax/#options. Maybe an internal link is better (= less likely to change when the server name changes).

### Please ensure

### DIFF
--- a/docs/experiments/generic-pattern-matching.md
+++ b/docs/experiments/generic-pattern-matching.md
@@ -234,10 +234,10 @@ To match an arbitrary sequence of items and capture their value in the example:
 
 ### Ignoring comments
 
-By default, the generic mode doesn't know about comments or code that
-should be ignored. In the following example, for some reason we want
-to detect CSS code that sets text color to blue. Here's the target
-code:
+By default, the generic mode does **not** know about comments or code
+that can be ignored. In the following example, we are 
+scanning for CSS code that sets the text color to blue. The target code
+is the following:
 
 ```
 color: /* my fave color */ blue;
--- a/docs/experiments/generic-pattern-matching.md
+++ b/docs/experiments/generic-pattern-matching.md
@@ -1,38 +1,72 @@
 ---
 append_help_link: true
-description: "Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data, such as HTML or XML. Generic pattern matching is experimental."
+description: "Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data such as XML. Generic pattern matching is experimental."
 ---
 
 # Generic pattern matching
 ## Introduction
-Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data, such as HTML or XML. Generic pattern matching is experimental.
+Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data such as XML. Generic pattern matching is experimental.
 
 As an example, consider this rule:
 ```yaml
 rules:
-- id: terraform-all-origins-allowed
-  patterns:
-  - pattern-inside: cors_rule { ... }
-  - pattern: allowed_origins = ["*"]
-  languages:
-  - generic
-  severity: WARNING
-  message: CORS rule on bucket permits any origin
+  - id: dynamic-proxy-scheme
+    pattern: proxy_pass $$SCHEME:// ...;
+    paths:
+      include:
+        - "*.conf"
+        - "*.vhost"
+        - sites-available/*
+        - sites-enabled/*
+    languages:
+      - generic
+    severity: WARNING
+    message: >-
+      The protocol scheme for this proxy is dynamically determined.
+      This can be dangerous if the scheme can be injected by an
+      attacker because it may forcibly alter the connection scheme.
+      Consider hardcoding a scheme for this proxy.
+    metadata:
+      references:
+        - https://github.com/yandex/gixy/blob/master/docs/en/plugins/ssrf.md
+      category: security
+      technology:
+        - nginx
+      confidence: MEDIUM
 ```
 
-The above rule matches this code snippet:
+The above rule [matches](https://semgrep.dev/playground/r/generic.nginx.security.dynamic-proxy-scheme.dynamic-proxy-scheme) this code snippet:
 
-```bash
-resource "aws_s3_bucket" "b" {
-  bucket = "s3-website-test-open.hashicorp.com"
-  acl    = "private"
+```
+server {
+  listen              443 ssl;
+  server_name         www.example.com;
+  keepalive_timeout   70;
 
-  cors_rule {
-    allowed_headers = ["*"]
-    allowed_methods = ["PUT", "POST"]
-    allowed_origins = ["*"]  # <--- Matches here
-    expose_headers  = ["ETag"]
-    max_age_seconds = 3000
+  ssl_certificate     www.example.com.crt;
+  ssl_certificate_key www.example.com.key;
+
+  location ~ /proxy/(.*)/(.*)/(.*)$ {
+    # ruleid: dynamic-proxy-scheme
+    proxy_pass $1://$2/$3;
+  }
+
+  location ~* ^/internal-proxy/(?<proxy_proto>https?)/(?<proxy_host>.*?)/(?<proxy_path>.*)$ {
+    internal;
+
+    # ruleid: dynamic-proxy-scheme
+    proxy_pass $proxy_proto://$proxy_host/$proxy_path ;
+    proxy_set_header Host $proxy_host;
+}
+
+  location ~ /proxy/(.*)/(.*)/(.*)$ {
+    # ok: dynamic-proxy-scheme
+    proxy_pass http://$2/$3/$1;
+  }
+
+  location ~ /proxy/(.*)/(.*)/(.*)$ {
+    # ok: dynamic-proxy-scheme
+    proxy_pass https://$1/$2/$3;
   }
 }
 ```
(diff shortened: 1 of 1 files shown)


docs/experiments/generic-pattern-matching.md
```diff
@@ -1,38 +1,77 @@
 ---
 append_help_link: true
-description: "Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data, such as HTML or XML. Generic pattern matching is experimental."
+description: "Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data such as XML. Generic pattern matching is experimental."
```
s-santillan, line 3:
```suggestion
description: "Semgrep can match generic patterns in languages that it doesn’t support yet. You can use generic pattern matching for languages that don’t have a parser, configuration files, or other structured data such as XML. Generic pattern matching is experimental."
```

"Semgrep can match matches generic patterns in languages that it doesn’t yet support." -> Looks like "matches" was a duplication?

docs/experiments/generic-pattern-matching.md
```diff
@@ -1,38 +1,77 @@
...
 ---
 
 # Generic pattern matching
+
+<!-- If you ever need to replace the examples below, a good way is to look
```
s-santillan, line 8: Thank you! This will help us maintain this. :)

docs/experiments/generic-pattern-matching.md
```diff
@@ -1,38 +1,77 @@
...
+     into the semgrep-rules repo under "generic" for an existing rule
+     that makes sense. -->
+
 ## Introduction
-Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data, such as HTML or XML. Generic pattern matching is experimental.
+Semgrep can match matches generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t yet have a parser, configuration files, and other structured data such as XML. Generic pattern matching is experimental.
```
s-santillan, line 13:
```suggestion
Semgrep can match generic patterns in languages that it doesn’t yet support. You can use generic pattern matching for languages that don’t have a parser, configuration files, or other structured data such as XML. Generic pattern matching is experimental.
```
"matches" is a dupe?

docs/experiments/generic-pattern-matching.md
```diff
@@ -49,7 +88,7 @@ Generic pattern matching has the following properties:
 
 ## Caveats and limitations
 
-Generic mode should work fine with any human-readable text, as long as it’s primarily based on ASCII symbols. In practice, it might work great with some languages and less well with others. In general, it’s possible or even easy to write code in weird ways that will prevent generic mode from matching. Note it’s not good for detecting malicious code. For example, in HTML one can write `&#x48;&#x65;&#x6C;&#x6C;&#x6F`; instead of `Hello` and this is not something the generic mode would match if the pattern is `Hello`, unlike if it had full HTML support.
+Generic mode should work fine with any human-readable text, as long as it’s primarily based on ASCII symbols. In practice, it might work great with some languages and less well with others. In general, it’s possible or even easy to write code in weird ways that will prevent generic mode from matching. Note it’s not good for detecting malicious code. For example, in XML one can write `&#x48;&#x65;&#x6C;&#x6C;&#x6F`; instead of `Hello` and this is not something the generic mode would match if the pattern is `Hello`, unlike if it had full XML support.
```
s-santillan, line 91: @adamkvitek I know we're not supposed to use "should" or "might" - but this is a genuinely experimental feature, and we genuinely don't know how it will behave for all possible combinations. I think I'd leave this in because it accurately tells users about the actual state of the product and doesn't mislead them with absolute guarantees?

adamkvitek: @s-santillan than express the uncertainty in other ways. Technical communication is closer to a sign language on roads sometimes than to a real language (similarly to how we mark optional steps as `optional` in our texts, you can just straight say that this is experimental thing without using intricate grammar. We can use admonitions also: https://docusaurus.io/docs/next/markdown-features/admonitions

@mjambon I dared to write some educated guesses on how generic mode works (that it does not really understand the syntax of the given language), please check below. This should make it very obvious to reader.

```suggestion
Semgrep can reliably understand the syntax of natively supported languages. The generic mode is useful for [supported languages](/supported-languages/), bringing specific limitations.

:::caution
- Generic mode is not good for detecting malicious code.
- The quality of results in the generic mode can vary depending on the language you use it for.
:::

The generic mode works fine with any human-readable text, as long as it is primarily based on ASCII symbols. Since the generic mode does not understand the syntax of the language you are scanning, the quality of the result may differ from language to language or even depend on specific code. As a consequence, the generic mode works well for some languages, but it does not give consistent results always. Generally, it's possible or even easy to write code in weird ways that prevent generic mode from matching. Note generic mode is not good for detecting malicious code.

**Example**: In XML, one can write `&#x48;&#x65;&#x6C;&#x6C;&#x6F` instead of `Hello`. If a rule pattern in generic mode is `Hello`, Semgrep is unable to match the `&#x48;&#x65;&#x6C;&#x6C;&#x6F`, unlike if it had full XML support.
```

docs/experiments/generic-pattern-matching.md
adamkvitek: @s-santillan Just changed don't to do **not** - emphasising negative.

docs/experiments/generic-pattern-matching.md
adamkvitek: Yes, thank you!

docs/experiments/generic-pattern-matching.md
```diff
@@ -1,38 +1,77 @@
...
+        - "*.conf"
+        - "*.vhost"
+        - sites-available/*
+        - sites-enabled/*
+    languages:
+      - generic
+    severity: WARNING
+    message: >-
+      The protocol scheme for this proxy is dynamically determined.
+      This can be dangerous if the scheme can be injected by an
```
adamkvitek, line 31:
```suggestion
      This can be dangerous if the scheme is injected by an
```

docs/experiments/generic-pattern-matching.md
```diff
@@ -49,7 +88,7 @@ Generic pattern matching has the following properties:
 
 ## Caveats and limitations
```
adamkvitek, line 89: To make this a bit more specific for Algolia search:

```suggestion
## Caveats and limitations of generic mode
```

docs/experiments/generic-pattern-matching.md
```diff
@@ -193,6 +232,35 @@ To match an arbitrary sequence of items and capture their value in the example:
     severity: WARNING
     ```
 
+### Ignoring comments
+
+By default, the generic mode doesn't know about comments or code that
+should be ignored. In the following example, for some reason we want
+to detect CSS code that sets text color to blue. Here's the target
+code:
```
adamkvitek, line 240:
```suggestion
By default, the generic mode does **not** know about comments or code
that can be ignored. In the following example, we are 
scanning for CSS code that sets the text color to blue. The target code is the following:
```

docs/experiments/generic-pattern-matching.md
```diff
@@ -1,38 +1,77 @@
...
+    # ok: dynamic-proxy-scheme
+    proxy_pass http://$2/$3/$1;
+  }
+
+  location ~ /proxy/(.*)/(.*)/(.*)$ {
+    # ok: dynamic-proxy-scheme
+    proxy_pass https://$1/$2/$3;
   }
 }
 ```
```
adamkvitek, line 77: I am not sure if you would be ok with me committing to your branch @mjambon - let me know. Due to GitHub limitations, I cannot suggest an edit for the following line below this comment - <code>* `...` allows skipping non-matching elements, up to 10 lines down the last match.</code> - please, can we change it to:  <code>* `...` (ellipsis operator) allows skipping non-matching elements, up to 10 lines down the last match.</code>

docs/experiments/generic-pattern-matching.md
```diff
@@ -193,6 +232,35 @@ To match an arbitrary sequence of items and capture their value in the example:
...
+to detect CSS code that sets text color to blue. Here's the target
+code:
+
+```
+color: /* my fave color */ blue;
+```
+
+[`options.generic_comment_style`](https://semgrep.dev/docs/writing-rules/rule-syntax/#options)
+is an option that can be set to ignore
+C-style comments as is the case here. Our simple Semgrep rule is:
```
adamkvitek, line 248:
```suggestion
Use the [`options.generic_comment_style`](/writing-rules/rule-syntax/#options)
to ignore C-style comments as it is the case in our example.
Our simple Semgrep rule is:
```

adamkvitek (commented):
@mjambon thank you for this work! I made some suggestions and comments to Sara's comments as I was whole weekend thinking about this PR. :D Please check and let me know when we can approve and merge this when you are done with them.

s-santillan:
@adamkvitek We're supposed to approve after reviewing, right? I'm going ahead and stamping this.

s-santillan (approved):

docs/experiments/generic-pattern-matching.md
adamkvitek: @mjambon if my suggestion is wrong or misleading in how generic mode works, feel free to rewrite what I wrote.

docs/experiments/generic-pattern-matching.md
mjambon: Reviewers are always welcome to make small edits on branches submitted for review! Editing branches that are still in progress comes with the risk of editing conflicts, but it's not the case here.

mjambon: I made the change.

docs/experiments/generic-pattern-matching.md
mjambon: @adamkvitek it's nice. I'll leave out "Generic mode is not good for detecting malicious code" because it's true of semgrep in general.

PR #686, opened 2022-08-29
experiments: Taint propagators

IagoAbal:
Closes PA-1684

### Please ensure

### DIFF
--- a/docs/experiments/taint-propagators.md
+++ b/docs/experiments/taint-propagators.md
@@ -6,6 +6,10 @@ description: "Taint propagators allow you to customize how taint is propagated."
 
 # Taint propagators
 
+Taint propagators allow you to customize how taint is propagated. 
+
+## Taint propagator example use
+
 Consider the following Python code:
 
 ```python
@@ -17,11 +21,11 @@ def test(s):
     sink(s.pop())
 ```
 
-Here, an unsafe `user_input` is stored into a set data-structure, and then we take
-a random element from that set and pass it into a `sink` function. Of course, this
-random element could be `user_input` itself, leading to an injection vulnerability!
+Code description: An unsafe `user_input` is stored into a `set` data structure.
+A random element from `set` is then passed into a `sink` function. This random
+element can be `user_input` itself, leading to an injection vulnerability!
 
-If we try to find this issue with this rule:
+Let's use the following rule to find the above-described issue: 
 
 ```yaml
 mode: taint
@@ -31,11 +35,11 @@ pattern-sinks:
 - pattern: sink(...)
 ```
 
-It will not find it. This is because, by itself, Semgrep only propagates taint
-through assignments.
+Semgrep only propagates taint through assignments. As a consequence, 
+Semgrep cannot find the issue using the rule displayed above.
 
-Using _taint propagators_ you can tell Semgrep how to propagate taint in these
-scenarios. Taint propagators are specified under the `pattern-propagators` key:
+The use of **taint propagators** enables Semgrep to propagate taint in these scenarios.
+Taint propagators are specified under the `pattern-propagators` key:
 
 ```yaml
 pattern-propagators:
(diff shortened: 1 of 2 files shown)


docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
...
+```yaml
+mode: taint
+pattern-sources:
+- pattern: user_input
+pattern-sinks:
+- pattern: sink(...)
+```
+
+It will not find it. This is because, by itself, Semgrep only propagates taint
+through assignments.
```
adamkvitek, line 35: @IagoAbal what do we mean by propagation through assignments?

```suggestion
Semgrep only propagates taint through assignments. As a consequence, 
Semgrep cannot find the issue using the rule displayed above.
```

docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
...
+
+Using _taint propagators_ you can tell Semgrep how to propagate taint in these
+scenarios. Taint propagators are specified under the `pattern-propagators` key:
```
adamkvitek, line 38:
```suggestion
The use of **taint propagators** enables Semgrep to propagate taint in these scenarios.
Taint propagators are specified under the `pattern-propagators` key:
```

docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
...
+
+```yaml
+pattern-propagators:
```
adamkvitek, line 41: @IagoAbal I was a bit confused by naming - how do we want to call this feature? Is it:

- Pattern propagator (as mentioned in the code snippet here)
- Taint propagator (as in the header)
- Propagator (as we call it in some paragraphs)

We can find one name and stick to it.

docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
...
+- pattern: $X.add($Y)
+  from: $Y
+  to: $X
+```
+
+A propagator consists of any arbitrary pattern, that must bind two metavariables
+of your choice. Then you must specify the `from` and the `to` metavariables.
+In our example, whenever Semgrep finds the pattern `$X.add($Y)`, it will check if
+`$Y` is tainted and, if it is, then it will propagate that taint to `$X`. Thus,
+adding tainted data to a set will mark the set itself as tainted.
```
adamkvitek, line 51:
```suggestion
In the example above, whenever Semgrep finds the pattern `$X.add($Y)`, it checks if
`$Y` is tainted. If `$Y` is tainted, Semgrep propagates that taint to `$X`. Thus,
adding tainted data to a set marks the set itself as tainted.

<iframe src="https://semgrep.dev/embed/editor?snippet=7lNe" border="0" frameBorder="0" width="100%" height="432"></iframe>

## Propagator rule requirements

A propagator rule necessitates:
- Any arbitrary pattern that must bind two metavariables of your choice.
- You must specify the `from` and the `to` metavariables.
```

docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=7lNe" border="0" frameBorder="0" width="100%" height="432"></iframe>
```
adamkvitek, line 53: Thank you for this example! @IagoAbal (Just moving it above.)

adamkvitek:
```suggestion
```

adamkvitek (approved):
@IagoAbal thank you! I made some suggestions, please check them. Approving so you are not blocked, but if my comments require further discussion, feel free to wait before merging (leaving this up to you) - we can always fix anything by future PRs.

docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
...
+    x = user_input
+    s = set([])
+    s.add(x)
+    #ruleid: test
+    sink(s.pop())
+```
+
+Here, an unsafe `user_input` is stored into a set data-structure, and then we take
+a random element from that set and pass it into a `sink` function. Of course, this
+random element could be `user_input` itself, leading to an injection vulnerability!
```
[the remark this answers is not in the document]
IagoAbal: Why "database injection"? This is a generic "injection" vulnerability, I'm just using a generic `sink` method.

docs/experiments/taint-propagators.md
IagoAbal: The feature is "taint propagators" but the operator is `pattern-propagators` because all our operators start with `pattern-`. That we cannot change. Now, "propagator" is just a short form of "taint propagator". We could rewrite those if you insist.

docs/experiments/taint-propagators.md
```diff
@@ -0,0 +1,53 @@
+---
+slug: taint-propagators
+append_help_link: true
+description: "Taint propagators allow you to customize how taint is propagated."
+---
+
+# Taint propagators
+
+Consider the following Python code:
```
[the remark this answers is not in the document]
IagoAbal: I don't like this header much because here the spec and the example are intermixed. If we had a more formal spec of what a propagator is, then followed by an example, then it would make more sense to me.

docs/experiments/taint-propagators.md
adamkvitek: @IagoAbal thanks for the explanation, looks like you fixed it!

PR #691, opened 2022-09-02
Move rule writing articles to the top level

adamkvitek:

### Please ensure


adamkvitek:
TODO: check links to these articles - they all need to be fixed potentially.

sidebars.js
```diff
@@ -175,6 +161,14 @@ module.exports = {
         items: [
             'playground',
             'trophy-case',
+            'writing-rules/overview',
```
s-santillan, line 164: I'd move playground and trophy-case to the bottom of the list. Just a suggestion though.

s-santillan:
@adamkvitek How about we change the header to "Learn Semgrep syntax" so it's clear that the users are learning about the language. (Semgrep isn't a full programming language, so the correct term is syntax, fyi.). Can you also try taking a screenshot so we can see the effect?

s-santillan (commented):
We didn't move any physical files, so the links inside the docs should just work, right? I'll approve once you can confirm. Thanks @adamkvitek

pabloest:
@s-santillan I agree about the header referring to something that we want to achieve.

sidebars.js
adamkvitek: Maybe Playground first and Trophy-case as the last? @s-santillan

adamkvitek: Ah... but there needs to be an explanation about writing rules first before we introduce Playground...

adamkvitek:
@pabloest @s-santillan what do you think about this layout now?
Screenshot 2022-09-08 at 16 48 31

- I still feel like... it is hard to fit the trophy-case anywhere, but it may be fine where it is now in this suggestion.
- Should I rewire the filesystem also? It will take some work, but maybe it will be nicer for maintainers (consistent folders in `docs` directory synced with the left sidebar as we will agree upon it here. I may need to check the `semgrep` repository also for any links which may get broken, here also of course.

s-santillan (approved):
lgtm

PR #694, opened 2022-09-07
Update data-flow overview and status pages

adamkvitek:

### Please ensure


docs/writing-rules/data-flow/overview.md
```diff
@@ -23,4 +27,14 @@ Semgrep strives for simplicity and to deliver on the promise of lightweight and
...
+You can expect both false positives and false negatives. You can remove false positives in different ways, for example, using [pattern-not](../../rule-syntax/#pattern-not) and [pattern-not-inside](../../rule-syntax/#pattern-not-inside). We definitely want to provide you with a way of eliminating false positives, so please [let file an issue if run into some](https://github.com/returntocorp/semgrep/issues/new/choose). We are rather happy to trade false negatives for simplicity and fewer false positives, but you are welcome to open a feature request if Semgrep misses some important bug you would like to catch, and we will certainly consider it.
+
+## Data-flow status
+
+In principle, the data-flow analysis runs on any [language supported by](../../supported-languages.md).
+
+Internally, Semgrep translates the generic syntax tree into an analysis-friendly intermediate language, the _IL_, on which it runs _mostly_ language-agnostic analysis. However, this translation is not fully complete. Unsupported constructs are typically specific to one or a few languages.
```
adamkvitek, line 36: Not sure if users need to know this.

docs/writing-rules/data-flow/overview.md
```diff
@@ -5,22 +5,36 @@ description: >-
...
+
 ## Design trade-offs
 
 Semgrep strives for simplicity and to deliver on the promise of lightweight and fast static analysis. In addition to being intra-procedural, here are some other trade-offs:
 
 - No path sensitivity: All _potential_ execution paths are considered, despite that some may not be feasible.
-- No pointer or shape analysis: _Aliasing_ that happens in non-trivial ways may not be detected, such as through arrays or pointers. Individual elements in arrays or other data structures are not tracked. There is also no proper field sensitivity at present, but this may be developed in the future.
+- No pointer or shape analysis: _Aliasing_ that happens in non-trivial ways may not be detected, such as through arrays or pointers. TODO Individual elements in arrays or other data structures are not tracked. There is also no proper field sensitivity at present, but this may be developed in the future.
 - No soundness guarantees: Semgrep ignores the effects of `eval`-like functions on the program state. It doesn’t make worst-case sound assumptions, but rather "reasonable" ones.
```
adamkvitek, line 28: I am not sure what this whole list means. We are trying to list very specific programmatic limitations. Are they still valid? Not sure about this whole thing. Maybe @IagoAbal can help here?

docs/writing-rules/data-flow/overview.md
```diff
@@ -5,22 +5,36 @@ description: >-
   Semgrep can run data-flow analyses on your code, this is used for constant propagation and for taint tracking.
 ---
 
-# Engine overview
+import SupportedLanguagesTable from '/src/components/reference/_supported-languages-table.mdx'
 
-Semgrep comes with an intra-procedural data-flow analysis engine that provides it with additional capabilities. [Constant propagation](../constant-propagation/) allows Semgrep to, for example, match `return 42` against `return x` when `x` can be reduced to `42` by constant folding. [Taint tracking (aka taint mode)](../taint-mode/) allows writing simple rules that catch complex [injection bugs](https://owasp.org/www-community/Injection_Flaws), such as those that can result in [cross-site scripting (XSS)](https://owasp.org/www-community/attacks/xss/).
+# Data-flow analysis engine overview
```
adamkvitek, line 10: Is data-flow one-to-one synonym with taint analysis, or there can be more data-flows, and taint analysis is one of them? @ievans

IagoAbal:
I wouldn't remove the Status page, at least not for now. At some point the idea is to specify there the level of support for different languages.

docs/writing-rules/data-flow/overview.md
```diff
@@ -5,22 +5,36 @@ description: >-
   Semgrep can run data-flow analyses on your code, this is used for constant propagation and for taint tracking.
 ---
 
-# Engine overview
+import SupportedLanguagesTable from '/src/components/reference/_supported-languages-table.mdx'
```
IagoAbal, line 8: This is not correct, language support for dataflow is not the same as for the matching engine. We actually still have to collect stats to properly define the status of each individual language. It's in my (long) to-do list.

docs/writing-rules/data-flow/overview.md
IagoAbal: The limitations still apply (and will likely apply for the foreseeable future) and advanced users need to know about them.

docs/writing-rules/data-flow/overview.md
IagoAbal: Advanced users can appreciate this so I would keep it.

adamkvitek: Ok! Would like to understand how it works and also if you now in which direction to point me or recommend somebody who can explain... @IagoAbal

docs/writing-rules/data-flow/overview.md
adamkvitek: @IagoAbal thanks! Ok, will fix this!

adamkvitek: I will include the following sentence:

```md
## Data-flow status

In principle, the data-flow analysis may run on any language [supported by Semgrep](/docs/supported-languages.md). However, compared to regular matching engine, data-flow analysis may still not work as expected for a particular language.
```

docs/writing-rules/data-flow/overview.md
efortuna: You may be familiar with the term IR (Intermediate Representation) instead of IL, but the idea is Semgrep performs an analysis of the code, creating an abstract syntax tree, but apparently Semgrep's IL doesn't encode capture some features of some specific languages, which means there are some features in some languages that Semgrep simply ignores/doesn't support.

efortuna (approved):
this all LGTM

docs/writing-rules/data-flow/overview.md
IagoAbal: I am not sure about saying that it "may still not work as expected", sounds too negative perhaps? The level of support is lower than for the matching engine, for now this is still not quantified, and for the record the data-flow engine is not GA yet.

IagoAbal: Although I realize I did write that. 😁

docs/writing-rules/data-flow/overview.md
adamkvitek: @IagoAbal I would avoid making documentation only for a selected circle of enlightened users. What @efortuna explained makes sense. The minimal thing we should do is to spell out the abbreviations in their first occurrence at least...

adamkvitek: Ah ok, the abbreviation is expanded, just in a way I am not used to. Sorry. :D

docs/writing-rules/data-flow/data-flow-overview.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Data-flow status
+
+In principle, the data-flow analysis may run on any language [supported by Semgrep](/docs/supported-languages.md). However, compared to the regular Semgrep matching engine, the data-flow analysis engine may still not work as expected for some languages. Please, report any issue you encounter in our [Semgrep GitHub](https://github.com/returntocorp/semgrep/issues/new/choose).
+
+Internally, Semgrep translates the generic syntax tree into an analysis-friendly intermediate language, the _IL_, on which it runs _mostly_ language-agnostic analysis. However, this translation is not fully complete. Unsupported constructs are typically specific to one or a few languages.
+
+Semgrep does not fail even if it finds an unsupported construct. The analysis simply continues while the construct is ignored. This may result in Semgrep not matching some code which should be matched (false negatives) or matching a code which should not be matched (false positives).
```
adamkvitek, line 38: @efortuna is this ok?

```suggestion
When Semgrep performs an analysis of the code, it creates an **abstract syntax tree** (AST) which is then translated into an analysis-friendly **intermediate language** (IL). Subsequently, Semgrep runs mostly language-agnostic analysis on IL. However, this translation is not fully complete. There can be features of some languages that Semgrep does not analyze correctly. In such a case, Semgrep does not fail even if it finds an unsupported construct. The analysis continues while the construct is ignored. This can result in Semgrep not matching some code that should be matched (false negatives) or matching a code that should not be matched (false positives).
```

docs/writing-rules/data-flow/overview.md
adamkvitek: Resolving, commiting in: https://github.com/returntocorp/semgrep-docs/pull/694#discussion_r965957359

docs/writing-rules/data-flow/data-flow-overview.md
```diff
@@ -0,0 +1,38 @@
+---
+slug: data-flow-overview
+append_help_link: true
+description: >-
+  Semgrep can run data-flow analyses on your code, this is used for constant propagation and for taint tracking.
+---
+
+# Data-flow analysis engine overview
+
+Semgrep provides an intra-procedural data-flow analysis engine that opens various Semgrep capabilites. See the following data-flow analysis tools:
```
adamkvitek, line 10:
```suggestion
# Data-flow analysis engines overview

Semgrep provides intra-procedural data-flow analysis engines that open various Semgrep capabilities. See the following data-flow analysis tools:
```

docs/writing-rules/data-flow/overview.md
adamkvitek: @IagoAbal ok! I would go with:

```
In principle, the data-flow analysis engines can run on any
 language [supported by Semgrep](/docs/supported-languages.md). 
However, the level of support is lower than for the regular
Semgrep matching engine.
```

PR #696, opened 2022-09-07
Update join rule example regexes to be more precise.

minusworld:
Based on conversation, here: https://r2c-community.slack.com/archives/CK86BJ5DW/p1661967762717409

### Please ensure


adamkvitek (approved):
@minusworld thanks! Approving. So the `- pattern-not-regex` value does not have to be marked as a string (in `""` or `''`)?

adamkvitek:
@minusworld thought the original suggestion was `.*(?<!\.html)$`. 🙂

PR #703, opened 2022-09-09
updated circleci yaml

holdenmcg:
PR to improve the sample ci config for circle CI. Gave comments on what the variable groups so as well as adding the SEMGREP_REPO_NAME variable to get auto-set rather than needing to set it manually.
image

### Please ensure

- [ x] A subject matter expert (SME) reviews the content
- [x ] This change has no security implications or else you have pinged the security team
- [x ] Redirects are added if the PR changes page URLs

### DIFF
--- a/docs/semgrep-ci/sample-ci-configs.md
+++ b/docs/semgrep-ci/sample-ci-configs.md
@@ -263,23 +263,16 @@ pipeline {
 
 ## CircleCI
 
-```yaml
+```
 version: 2.1
 jobs:
   semgrep-scan:
     parameters:
-      repo_path:
-        type: string
-        default: myorg/semgrep-test-repo
       default_branch:
         type: string
         default: main
     environment:
-      # Scan with rules set in Semgrep App's rule board.
-      # Get your token at semgrep.dev/orgs/-/settings/tokens
-      SEMGREP_APP_TOKEN: $SEMGREP_APP_TOKEN
-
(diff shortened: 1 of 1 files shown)


docs/semgrep-ci/sample-ci-configs.md
```diff
@@ -263,37 +263,29 @@ pipeline {
 
 ## CircleCI
 
-```yaml
```
s-santillan, line 266: Looks like an accidental yaml removal in the line?

s-santillan (approved):
@holdenmcg thank you!

docs/semgrep-ci/sample-ci-configs.md
holdenmcg: I removed the yaml syntax as I was having issues with a few lines; the yaml syntax wasn't interpreting them as comments and kept highlighting them red, since they were more like bash statements than comments, so I left out the yaml syntax. :/

holdenmcg: image
Those echo statements towards the bottom shouldn't be red, they should be commented out; would you happen to know how to get those to comment?

s-santillan: Looks like the renderer doesn't like pipes (|). Let's go ahead and merge it without the yaml. Thanks!

PR #705, opened 2022-09-12
Sara/improve/tabbed configs

s-santillan:

### Please ensure

This update adds MANY config file variations into the docs, and arranges them by tabs.

**When looking at the code, some aggressive code folding may ease the cognitive load.**


s-santillan:
Hi @holdenmcg!

I decided to stop here and check in with you for feedback first. It's ready for review.

Here's a summary of the changes so you're not so overwhelmed.

1. I just copy pasted the base GitHub and GitLab config files then added or removed env vars as necessary for the App-connected vs non-app-connected scans. Be assured that **no changes were made to the code** for those two, since they aren't affected by the recent changes to env vars.
2. For Jenkins and BitBucket Pipelines, I used [this doc](https://docs.google.com/document/d/1AYr-L4WFgBBOxbiBYRAjToKzOkciDt7ScEggT1ZlTkE/edit#) to determine what env vars are still set in the config file after 0.98.0. I still write out those vars, under Troubleshooting, and added some contextual comments.
3. I added instructions for all of them, making it clear that Jenkins and BitBucket needs a separate job for diff-aware scanning.

holdenmcg (approved):
Love the tabs w/ the double snippets as well as the comments over the variables. 👍

PR #706, opened 2022-09-13
Add guidelines to deal with submodules

mjambon:
This adds a section on how to avoid common problems with submodules.

### Please ensure

### DIFF
--- a/docs/contributing/contributing-code.md
+++ b/docs/contributing/contributing-code.md
@@ -265,7 +265,9 @@ gitup() {
   Good to know:
   - Merging in B can be done with a merge commit or by squashing the
     commits.
-  - If squashing commits in B, the original commit referenced by A becomes
-    orphaned when the branch is deleted but remains cached by git for
-    a while. This is usually sufficient to not require A to point to
-    the newly-squashed commit.
+  - If squashing commits in B, you must know that the original commit
+    referenced by A becomes orphaned when the branch is deleted but
+    remains cached by git for a while. This is usually sufficient to
+    not require A to point to the newly-squashed commit. _If this turns
+    out to be problematic in practice, we may have to disallow
+    commit squashing in the future._
--- a/docs/contributing/contributing-code.md
+++ b/docs/contributing/contributing-code.md
@@ -11,11 +11,11 @@ This README gives an overview of the repository. For further information on buil
 
 ## File structure
 
-Semgrep consists of a Python wrapper (`semgrep-cli`) around an OCaml engine (`semgrep-core`) which performs the core parsing/matching work. Within `semgrep-core`, there are two sources of parsers, `pfff`, linked as a [submodule](https://github.com/returntocorp/pfff), and `tree-sitter-lang`, built using [tree-sitter](https://github.com/tree-sitter/tree-sitter). Additionally, `semgrep-core` contains a subengine, `spacegrep`, for generic matching.
+Semgrep consists of a Python wrapper (`semgrep-cli`) around an OCaml engine (`semgrep-core`) which performs the core parsing/matching work. Within `semgrep-core`, there are two sources of parsers, [`pfff`](https://github.com/returntocorp/pfff), linked as a submodule, and `tree-sitter-lang`, built using [tree-sitter](https://github.com/tree-sitter/tree-sitter). Additionally, `semgrep-core` contains a subengine, `spacegrep`, for generic matching.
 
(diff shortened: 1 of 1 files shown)


docs/contributing/contributing-code.md
```diff
@@ -230,3 +230,42 @@ On M1 macs some `pre-commit` tests may fail.
...
+being tricky to use. To minimize problems, make sure to follow these
+guidelines:
+
+* When checking out a new branch or commit, update the submodules
+  using the command `git submodule update --init --recursive`.
+  Adding a shortcut to your shell can be useful. The following is a
+  Bash function that lets you call `gitup`. It goes into your `~/.bashrc`:
+
+```bash
+gitup() {
```
IagoAbal, line 248: Why not a Git alias? I think you could define `git up` to do that.

docs/contributing/contributing-code.md
```diff
@@ -230,3 +230,42 @@ On M1 macs some `pre-commit` tests may fail.
...
+  echo "git submodule update --init --recursive"
+  git submodule update --init --recursive
+}
+```
+
+* When modifying both a parent repo A and one of its submodules B,
```
IagoAbal, line 254: Doesn't this require that PR  is merged with a merge commit and forbids squash-and-merge? Is that the case with all our dependencies?

docs/contributing/contributing-code.md
```diff
@@ -230,3 +230,42 @@ On M1 macs some `pre-commit` tests may fail.
...
+  2. Make sure PR B is merged **before** PR A.
+     This ensures that other developers will pick up the changes on B
+     when making their own changes.
+  3. After merging PR A, check that submodule B is still up-to-date
+     with respect to its main branch, especially if PR B was merged
+     more than an hour ago.
+  Good to know:
+  - Merging in B can be done with a merge commit or by squashing the
+    commits.
+  - If squashing commits in B, the original commit referenced by A becomes
```
IagoAbal, line 268: Oh it's mentioned here. I personally don't agree with this practice. It may not cause any immediate problems, but I also care about being able to build things in the future.

docs/contributing/contributing-code.md
nmote: Yes, you can have a git alias for this: https://github.com/nmote/dotfiles/blob/420a361/.gitconfig#L19

nmote: Also, we have an internal Notion page for Git tips. Maybe we should move them all here? Notably, I think these config lines also help a lot with dealing with submodules:

https://github.com/nmote/dotfiles/blob/420a361/.gitconfig#L44-L46

mjambon: Let's not bikeshed here. If you want to change this section, please post an edit I can approve with one click.

docs/contributing/contributing-code.md
mjambon: I agree but I don't want to spend more time on this. I want people to understand the consequences of their actions (squashing commits and not pointing to the squashed commit). For some people, it can speed things up a bit because the submodule gets updated frequently. I'll let them decide what's best and hold them responsible for any ensuing success or failure.

mjambon: I added a warning at the end of the paragraph.

nmote:
I don't care very much about the bash function vs git alias decision, but I do care that this adds a second place where we maintain Git advice, especially since the Notion page also has recommendations on dealing with submodules. I would support moving the existing Git advice over to this page, so that it is publicly accessible, but I don't think it's a good idea to have two separate places where we keep this sort of thing.

adamkvitek:
I agree with @nmote.

mjambon:
This is not a guide on all things related to git, this is a tutorial on how to contribute a piece of code to the semgrep project effectively. We can have a tutorial on how to get set up with git for working at r2c, but that's a different concern (covers projects from multiple teams, doesn't have to be public, solves a different problem, ...). I draw all the inspiration for the separation of concerns from https://documentation.divio.com/. I think it's very useful to keep in mind what a document is for and stick to it. And yes, there will be duplication of information because the same material will be presented in different contexts. That's ok.

aryx:
I agree with @mjambon . The Notion page is also accessible only to r2c employees whereas this document is supposed to be public and accessible to external contributors.

IagoAbal:
If we stick to separation of concerns, which I agree it's a good thing, I don't see why we should explain how to create a Bash alias in a contribution guide for Semgrep.

PR #710, opened 2022-09-13
update docusaurus & clean 

s-santillan:

### Please ensure

- [ ] A reviewer runs `yarn build` and `yarn serve` to check that this works.

Accidentally generated package-lock.json a while back. Removing it. Updated docusaurus through yarn to check nothing broke.


pabloest (approved):
@s-santillan Note I got this error at first when running `yarn install`:

`error @docusaurus/core@2.1.0: The engine "node" is incompatible with this module. Expected version ">=16.14". Got "16.13.0"`

I upgraded to Node 16.14.2 and then was able to `yarn install` and build + serve the site just fine. But just something to possibly look out for when this merges into production.

s-santillan:
Ran `yarn install` and `yarn serve` here on node 16.14.2, everything built. will merge shortly.

PR #722, opened 2022-09-29
Add documentation for how many semgrep versions are supported by the app

chmccreery:

### Please ensure

Closes APP-2251

### DIFF
--- a/docs/semgrep-app/getting-started-with-semgrep-app.md
+++ b/docs/semgrep-app/getting-started-with-semgrep-app.md
@@ -138,7 +138,7 @@ Scanning a project from the CLI is a standalone action. This means that you manu
 
 :::info Prerequisites
 - Semgrep CLI must be installed. See [Getting started with Semgrep CLI](../../getting-started).
-- Ensure that you are running a recent version of Semgrep CLI. Semgrep-app supports the 10 most recent minor versions of semgrep.
+- Ensure that you are running a recent version of Semgrep CLI. Semgrep-app supports the 10 most recent minor versions of Semgrep.
 :::
 
 To scan a local repository through Semgrep CLI, follow these steps:
--- a/docs/semgrep-app/getting-started-with-semgrep-app.md
+++ b/docs/semgrep-app/getting-started-with-semgrep-app.md
@@ -29,6 +29,10 @@ Semgrep App supports code scanning from:
 
 This guide walks you through scanning code in both types of environments.
 
+:::info
+Many improvements to the App experience only work with up-to-date Semgrep versions. For this reason, Semgrep App only supports the 10 most recent minor versions of the Semgrep open-source tool. For example, if the latest release was 0.114.0, then versions greater than 0.104.0 are supported and others may be deprecated or error out. It is recommended to use the "latest" tag for all Semgrep CI scans.
+:::
+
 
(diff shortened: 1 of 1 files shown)


docs/semgrep-app/getting-started-with-semgrep-app.md
```diff
@@ -29,6 +29,10 @@ Semgrep App supports code scanning from:
 
 This guide walks you through scanning code in both types of environments.
 
+:::info
+Many improvements to the App experience only work with up-to-date Semgrep versions. For this reason, Semgrep App only supports the 10 most recent minor versions of the Semgrep open-source tool. For example, if the latest release was 0.114.0, then versions greater than 0.104.0 are supported and others may be deprecated or error out. It is recommended to use the "latest" tag for all Semgrep CI scans.
```
adamkvitek, line 33:
```suggestion
Many improvements to the Semgrep App experience only work with up-to-date Semgrep CLI versions. For this reason, Semgrep App only supports the 10 most recent minor versions of the Semgrep open-source tool. For example, if the latest release was 0.114.0, 10 versions greater than 0.104.0 are supported while earlier versions, such as 0.103.0 can be deprecated or can result in failure. Use the **latest** tag for all Semgrep CI scans.
```

adamkvitek, line 33: Thank you @chmccreery! Two questions:

- Where is the `latest` tag? I do not see anything in our docs except for `runs-on: ubuntu-latest`.
- Should we mention this also in https://semgrep.dev/docs/semgrep-ci/running-semgrep-ci-with-semgrep-app/? This potentially sounds like something we should mention in many places.

adamkvitek (approved):
@chmccreery thank you, I left two comments and made one commit to fix some specifics to render this in a bit more flashy manner (I moved one step to Prerequisites also). This is how the Prerequisites will look like:
Screenshot 2022-09-30 at 12 36 46

docs/semgrep-app/getting-started-with-semgrep-app.md
chmccreery: > Where is the latest tag? I do not see anything in our docs except for runs-on: ubuntu-latest.

The "latest" tag is a tag is in docker: https://hub.docker.com/layers/returntocorp/semgrep/latest/images/sha256-31f53cf64ae228692c35572ea733db963b374a5eed50b3de32dc7fdf82352bf0?context=explore It is also the default behavior in CI/CD, but there are some users who have pinned particular versions, or whose CI/CD files are otherwise unusual, who might need to actually specify the "latest" tag.

> Should we mention this also in https://semgrep.dev/docs/semgrep-ci/running-semgrep-ci-with-semgrep-app/? This potentially sounds like something we should mention in many places.

Hmm yep, I agree! Do you mind adding it to https://semgrep.dev/docs/semgrep-ci/running-semgrep-ci-with-semgrep-app/ and anywhere else you think makes sense?

chmccreery: Feel free to merge this as-is now, or push the changes mentioned above first!

chmccreery: @adamkvitek feel free to merge this as-is or make the changes mentioned above and then merge!

PR #742, opened 2022-10-17
Sara/mkt 429 semgrep ci clean up configuration

s-santillan:

### Please ensure

(n/a) A subject matter expert (SME) reviews the content

This mainly reorganizes the CI Configuration Reference to be centered around the environment variables. The configuration to set up how semgrep in CI environments handles errors is separated into a different doc so that users can see it from the TOC (sidebar)


docs/semgrep-ci/configuration-reference.md
```diff
@@ -22,124 +22,243 @@ Object.entries(frontMatter).filter(
...
-## Selecting rules to scan with (`SEMGREP_RULES`)
+## Environment variables for configuring scan behavior
 
-```sh
-SEMGREP_RULES="p/security-audit p/secrets"
+These environment variables configure various aspects of your CI job, such as a job's timeout or source of rules.
+
+### `SEMGREP_APP_TOKEN`
+
+Set `SEMGREP_APP_TOKEN` to send findings to Semgrep App and use rules from the Rule Board. You must have a Semgrep App account to use this environment variable. To generate a token, see [Creating a `SEMGREP_APP_TOKEN`](/docs/semgrep-ci/running-semgrep-ci-with-semgrep-app/#creating-a-semgrep_app_token). `SEMGREP_APP_TOKEN` is incompatible with `SEMGREP_RULES`. Do not set `SEMGREP_RULES` environment variable within the same CI job as `SEMGREP_APP_TOKEN`.
```
adamkvitek, line 37: Optional, not blocking: The - `You must (...)` part and `Do not set (...)` maybe we can move them into a prerequisite admonition and caution admonition?

docs/semgrep-ci/configuration-reference.md
```diff
@@ -22,124 +22,243 @@ Object.entries(frontMatter).filter(
...
-- On GitLab, insert the flag after the `script` key (for example, `script: semgrep ci --suppress-errors` to state the default option).
-- Insert these flags in an equivalent key in configuration files of other CI providers.
- 
-See the [Examples of blocking findings and errors configuration](#examples-of-blocking-findings-and-errors-configuration) below.
- 
-:::info
-- For more information about specific Semgrep exit codes, see [CLI reference](../../cli-reference/#exit-codes).
-- This functionality replaces the audit mode `SEMGREP_AUDIT_ON` (collecting findings silently for [Semgrep App > Findings](https://semgrep.dev/manage/findings)).
-:::
+Set `SEMGREP_GHA_MIN_FETCH_DEPTH` to configure the **minimum** number of commits `semgrep ci` fetches from `remote` when calculating the merge-base in GitHub Actions. Having more commits available helps Semgrep determine what changes came from the current pull request, fixing issues where Semgrep would otherwise report findings that were not touched in a given pull request. This value is set to 0 by default.
```
adamkvitek, line 68: The following sentence lacks a call to action for a user (had to read it a couple of times to figure out one): `Having more commits available helps Semgrep determine what changes came from the current pull request, fixing issues where Semgrep would otherwise report findings that were not touched in a given pull request. This value is set to 0 by default.`

Maybe something like:

```
For optimal performance, use `SEMGREP_GHA_MIN_FETCH_DEPTH` with more commits. Having more commits available helps Semgrep determine what changes came from the current pull request, fixing issues where Semgrep would otherwise report findings that were not touched in a given pull request. This value is set to 0 by default.
```

Note: `More commits` sounds a bit vague also. Do we mean 3 commits or 20?

docs/semgrep-ci/configuration-reference.md
```diff
@@ -22,124 +22,243 @@ Object.entries(frontMatter).filter(
...
-SEMGREP_REPO_NAME="foo/bar"
-SEMGREP_COMMIT="a52bc1ef"
-SEMGREP_PR_ID="44"
+By default, Semgrep App autodetects values such as the name of your repository, which Semgrep uses to generate hyperlinks (URLs) to the specific repository code that generated the finding.
+
+Set any as needed or all of the following environment variables to troubleshoot and override autodetected CI environment values.
+
+### `SEMGREP_BRANCH`
+
+Set `SEMGREP_BRANCH` to define the branch name for the URL used to generate hyperlinks. To avoid hardcoding this value, check your CI provider's documentation for available environment variables that can automatically detect the correct values for every CI job.
```
adamkvitek, line 119: Generate hyperlinks where? To display them on Findings page? I am not sure I understand what this option allows.

adamkvitek (approved):
@s-santillan left some comments, approving, looks great!

docs/semgrep-ci/configuration-reference.md
```diff
@@ -22,124 +22,243 @@ Object.entries(frontMatter).filter(
 
 # Continuous Integration (CI) configuration reference
 
-Configure Semgrep CI by passing various environment variables in your Continuous Integration (CI) jobs.
+Use this reference to configure Semgrep's behavior in CI environments. This reference includes both command-line options and environment variables. You can also define these variables within a CI configuration file. Refer to your CI provider's documentation for the correct syntax. Examples are written in a Bash environment except when otherwise stated.
 
-:::info
-While environment variables are the preferred way to configure Semgrep CI, pass any of these options as command-line options. Refer to the output of `semgrep ci --help` to find the corresponding flags.
+:::tip
+Unless otherwise stated, you can also set many of these environment variables within your local environment for testing or experimental purposes. Run `semgrep ci` in your command line after setting environment variables to test their effect.
```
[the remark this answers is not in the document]
s-santillan: Adam - we agreed this would be a reference doc, right? It's not meant to tell the user how to set up env vars, it's meant to supplement what the user already knows to do. See https://semgrep.dev/docs/semgrep-ci/running-semgrep-ci-with-semgrep-app/ or https://semgrep.dev/docs/semgrep-ci/running-semgrep-ci-without-semgrep-app/. The user is likely to go through those two first before looking in a reference doc.

Maybe what you'd like is a link instead to either of those guides?

docs/semgrep-ci/configuration-reference.md
s-santillan: Good idea. Done

docs/semgrep-ci/configuration-reference.md
s-santillan: Sure, I added in the link to the findings page.

docs/semgrep-ci/configuration-reference.md
s-santillan: Some context for this - it was made to deal with diff-aware findings duplication for a particular use case. We can't say for sure "how many" commits back the user needs. What i have done is to add a note saying "don't set this variable unless you are encountering findings duplication in your diff aware scans".

docs/semgrep-ci/configuration-reference.md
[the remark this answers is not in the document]
adamkvitek: Links might help here! @s-santillan I mean - this is not only about reference or not - what do we mean here by the local environment? Going through both guides linked above, I cannot figure out how to include a local CI with environment variables file (without any SCM if I understand that correctly) and run CI locally. 

In short: Do we want to say here that users can try out any configuration of environment variables (for any SCM provider) also locally, by calling respective configuration files (can it be `.semgrep` in the root of a repository https://semgrep.dev/docs/semgrep-ci/running-semgrep-ci-without-semgrep-app/#adding-rules-to-scan-with-semgrep-ci)? Can we mention this in the guides linked above? What would be the commands to do so? Is it `semgrep ci semgrep.yml` or `semgrep ci .gitlab-ci.yml` for respective SCM file? This can be two lines of added text...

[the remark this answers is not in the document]
s-santillan: So - part of your concern should be addressed. But part of your confusion stems from a misunderstanding of CI and env vars. You don't need any CI provider (and therefore you don't need any yaml file, none of those .semgrep-ci.yml) for env vars. Env vars are just env vars. They can be set in a CI provider's environment (remote) and yours (locally). The idea is that the user can set these locally in their development machine to test so they have an idea of how semgrep works when they DO hook it up to a pipeline and run in a remote environment provided by a CI provider. Now - am I worried that **developers** will not understand what I am writing about? Devs should generally know the difference between setting an env var locally and remote (such as CI).

There is no CircleCI, GitHubActions, etc. when the user sets up env vars locally. That's what the word "locally" means in this context.

If these env vars can be set locally, why aren't they under Semgrep CLI? Because the env vars CAN be set locally to for testing purposes to simulate how semgrep will behave in CI, but are MAINLY used in CI.

I think the actionable change here for me is to spell out "local development environment" instead of "local environment" and specify that the env vars are set in the command line.

docs/semgrep-ci/configuration-reference.md
adamkvitek: @s-santillan I like the added note! ("don't set this variable unless you are encountering findings duplication in your diff aware scans")

docs/semgrep-ci/configuration-reference.md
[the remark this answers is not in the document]
adamkvitek: @s-santillan thank you for the explanation! I think that DevOps and devs experienced in working with this can understand this, but not most of them. I am still not sure if I should create a file for these env vars (different than the ones mentioned above for SCMs), or if I can make a oneliner or both previous options together.

Maybe the actionable item can be to document a small procedural article: Testing Semgrep in CI locally? It can be nice for some users...

PR #746, opened 2022-10-19
Update getting-started.md

adamkvitek:

### Please ensure

### DIFF
--- a/docs/running-rules.md
+++ b/docs/running-rules.md
@@ -42,10 +42,6 @@ Explore the Semgrep Registry by following these steps:
 4. Optional: Run registry rules simultaneously with local rules:
    <pre class="language-bash"><code>semgrep --config="<span className="placeholder">RULESET-ID</span>" --config=<span className="placeholder">PATH/TO/MYRULE.YAML PATH/TO/SRC</span></code></pre>
 
-:::tip
-See Semgrep [Rule Board](semgrep-app/rule-board.md) to select which rules Semgrep uses and what action it undertakes when it generates a finding.
-:::
-
 ### Running Semgrep Registry continuously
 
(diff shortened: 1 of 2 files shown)


docs/running-rules.md
```diff
@@ -38,12 +38,12 @@ Explore the Semgrep Registry by following these steps:
...
 3. Copy the code for local install, and then add the path to the source code you want to check in your terminal:
-    <pre class="language-bash"><code>semgrep --config="<span className="placeholder">RULESET-ID</span>" [<span className="placeholder">PATH/TO/SRC</span>]</code></pre>
+    <pre class="language-bash"><code>semgrep --config="<span className="placeholder">RULESET-ID</span>" <span className="placeholder">PATH/TO/SRC</span></code></pre>
 4. Optional: Run registry rules simultaneously with local rules:
    <pre class="language-bash"><code>semgrep --config="<span className="placeholder">RULESET-ID</span>" --config=<span className="placeholder">PATH/TO/MYRULE.YAML PATH/TO/SRC</span></code></pre>
 
-:::info
-Try Semgrep [Rule Board](semgrep-app/rule-board.md) to determine which rules Semgrep uses and what action Semgrep undertakes when it generates a finding.
+:::tip
+See Semgrep [Rule Board](semgrep-app/rule-board.md) to select which rules Semgrep uses and what action it undertakes when it generates a finding.
```
s-santillan, line 46: Hmm... okay so for running-rules LOCALLY, the implication is we're using Semgrep CLI.

It is technically correct that we can run Semgrep APP rule board locally, but we haven't explained how to connect the LOCAL set up to the APP (webapp) setup.I'd suggest removing this entirely, it's not really related and gives the user the idea that the Rule Board and Registry are somehow related without explaining what the relation is.

s-santillan (approved):
Left only one comment!

docs/running-rules.md
adamkvitek: This is true, I have the same feeling @s-santillan. Will remove this.

PR #749, opened 2022-10-20
Remove the 'Integrating Semgrep with webhooks link

Jason-Lim00:
With the deprecation of Zendesk on our side this link should also be removed from the docs as it is also deprecated. 

* [Integrating Semgrep with webhooks](https://support.semgrep.dev/hc/en-us/articles/4415975269275-Integrating-Semgrep-with-Webhooks)

### Please ensure


s-santillan (approved):
@Jason-Lim00 thank you for this catch!

PR #751, opened 2022-10-25
removed registry license document

s-santillan:

### Please ensure

(n/a) A technical writer reviews the content or PR

**I also grep -lir'ed for any other usage of license to make sure we're good. And we should be.** Licensing in FAQ is linked to the repositories themselves, not docs/licensing.md


pabloest (approved):
It might be nice to add a redirect but I’m not sure what that would be.

Or alternatively we could restate Semgrep’s license here (and the license for the rules) so that it’s discoverable and we don’t send people to a GitHub page to figure it out. That can be a separate PR though, if we decide to do that.

s-santillan:
@pabloest Agree, a new ~~phoenix~~ doc can rise up. Left a possible alternative in Slack :)

PR #757, opened 2022-10-28
Move data-flow analysis and experiments under writing rules in sidenav

adamkvitek:

### Please ensure


adamkvitek:
This was requested by @IagoAbal as Data-flow analysis and Experiments are more related to writing rules. We probably moved them higher in the Sidenav to promote Taint analysis.

sidenav-rule-writing

**Please check and discuss**: This greatly reduces the Semgrep Platform category, I am not sure what this category actually should contain. Screenshot 2022-10-28 at 11 53 05

IagoAbal (approved):
Thanks!

pabloest:
@adamkvitek Overall I think it’s ok to move the content structure in this way, but I think the directory paths of the experiments articles would need to be updated to reflect the new nesting under “writing rules.” 

Also, I’m not sure if the data-flow section should be within “writing rules.” It’s true, one can use this information to write rules, but we also have Registry rules that leverage data-flow. So it’s not strictly about writing rules, IMO.

IagoAbal:
What kind of info about dataflow we want to show to non-rule-writers ? An overview of what can be done via the data-flow engine ? We could perhaps split the docs that way ?

Also, if we change URLs, please remember to add redirects.

adamkvitek:
@pabloest just asking to make sure we understand each other exactly:

1. "_the directory paths of the experiments articles would need to be updated to reflect the new nesting under “writing rules.”_" - This means that I should move all experiments MD files under `writing-rules` folder and update all links leading to them (in the docs and sidebar, maybe check `semgrep` repository if there are some links leading to our experiments).
2. "_I’m not sure if the data-flow section should be within “writing rules.” It’s true, one can use this information to write rules, but we also have Registry rules that leverage data-flow. So it’s not strictly about writing rules, IMO._" Possible solutions:
    A. We can keep data-flow analysis where it is now (under Semgrep Platform).
    B. We can keep data-flow analysis where it is, and we can also modularize and reshuffle some parts of these articles and include them under `writing-rules` category (so data-flow articles under the Semgrep platform will look the same, but some things can be also mentioned differently under writing-rules). There are some parts that seem to motivate users to write rules: http://localhost:3000/docs/writing-rules/data-flow/taint-mode/#mini-cookbook 
    C. Split the docs as @IagoAbal suggests - an overview of what can be done through data-flow analysis, as well as data-flow status will be under the Semgrep platform, but specific details showing different kinds of rules that can be made with data-flow analysis can be moved under writing-rules.

Let me know what you think, maybe I am overthinking this.

adamkvitek:
@pabloest maybe for now 2. A. is the best solution and we can figure out a better layout later. My initial question here was about the whole Semgrep platform category, which seems very strange to me still... what should it include? I know there was some thinking behind this, but I am not getting the idea when looking at this category. As Semgrep platform I would imagine: Semgrep CLI, Semgrep in CI, DeepSemgrep. Also @s-santillan 
Screenshot 2022-11-02 at 11 08 56

s-santillan:
For @adamkvitek - here are the pertinent threads to answer your question:

https://returntocorp.slack.com/archives/C02JT6YA5S7/p1666633199382899
https://returntocorp.slack.com/archives/C032YLMN3NH/p1666382211087109
https://returntocorp.slack.com/archives/C02JT6YA5S7/p1666633199382899

The idea is that Platform = { All features common to Semgrep SAST and SSC }
In the very first thread, we talk about how we want to compare some things (see info about both SAST and SSC together in 1 doc, such as Supported Languages and Pricing and Billing) while still having a dedicated space for both products. Technically "Writing custom rules" should be under Semgrep Platform, because it puts it "above the fold" but I believe you made a change recently. I'm fine with that, it doesn't have to be a rigid nomenclature.

PR #762, opened 2022-11-01
Adamkvitek/mkt 343 improve release notes format

adamkvitek:

### Please ensure


s-santillan:
Looks good, I think this is a Pablo-level approval :)

pabloest (commented):
Can we keep the same URL parent structure, following `/docs/release-notes/*` instead of `/docs/updates/*`?

Also, I think you can use `overview` as the name and path of introduction page instead of `release-notes-introduction` — that follows the convention we use in other subdirectories.

Outside of that, this looks great to me!

adamkvitek:
@pabloest implemented. Notes:

1. I chose to go for `/docs/updates` initially as we include `rule-updates` also in the same directory (but I guess repetition of `updates/rule-updates` is not great. Rule updates are now under `/docs/release-notes/rule-updates`. But at least we will not break some links in our Slack etc. 👍 
2. I shortened `release-notes-introduction` to `introduction` - not sure if we should call it the overview. The overview would say much more, probably giving an outline of documents etc (I also checked [MD](https://www.macmillandictionary.com/dictionary/british/overview and [OD](https://www.oxfordlearnersdictionaries.com/definition/english/overview?q=overview)). at least that is how I perceive it. Please let me know if this is ok. See also another PR where we use Introduction also: https://github.com/returntocorp/semgrep-docs/pull/763/files#diff-f77dbffe5437a9037bf3e9bb15f5cb95ff2bcc2f3203bacac3833450f6abd26eR1-R14

pabloest (approved):

PR #763, opened 2022-11-02
Split overview of Experiments category to Autofix and Deprecated experiments, add introduction

adamkvitek:

### Please ensure

### DIFF
--- a/docs/experiments/introduction.md
+++ b/docs/experiments/introduction.md
@@ -1,6 +1,6 @@
 ---
-id: introduction-experiments
-slug: introduction-experiments
+id: introduction
+slug: introduction
 title: Introduction
 hide_title: true
 append_help_link: true
--- a/docs/experiments/introduction.md
+++ b/docs/experiments/introduction.md
@@ -7,7 +7,7 @@ append_help_link: true
 description: "Introduction of Semgrep experiments that also documents that some experiments can sunset or become GA, which means that particular documents can change their position in docs also."
 ---
 
-# Introduction to Semgrep Experiments
+# Introduction to Semgrep experiments
 
 The experiments category documents experimental features and the way you can use them. In the future, as it is the nature of experiments, some of these experiments can become deprecated, and others can become generally available (GA), meaning that GA features are fully supported parts of Semgrep. If a feature is deprecated, its documentation is moved to the [Deprecated experiments](/experiments/deprecated-experiments/) document. If a feature becomes GA, its docs are moved to a relevant category outside of the experiments section.
 
--- a/docs/experiments/introduction.md
+++ b/docs/experiments/introduction.md
@@ -1,0 +1,13 @@
+---
+slug: introduction-experiments
+title: Introduction
+hide_title: true
+append_help_link: true
+description: "Introduction of Semgrep experiments that also documents that some experiments can sunset or become GA, which means that particular documents can change their position in docs also."
+---
+
+# Introduction to Semgrep Experiments
+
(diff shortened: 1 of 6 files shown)


sidebars.js
```diff
@@ -50,7 +50,7 @@ module.exports = {
           type: 'category',
           label: 'Experiments 🧪',
           items: [
-            'experiments/overview',
```
aryx, line 53: We probably also want an introduction (not overview) section explaining what experiments mean.
It basically means new features we are trying that we think are useful, but that may disappear.
We make no promise on experimental features.
cc @ievans

aryx (changes requested):

sidebars.js
aryx: In particular, some experiments might be promoted to proprietary features that only logged-in users can access.

aryx: cc @dlukeomalley

aryx: and we can also say some experiments may also just disappear, because they are a maintenance burden we can't support.
And some might be promoted to regular features.
We could even give some examples of past experiments who got promoted (I think autofix and generic matching should be for example).

adamkvitek:
@aryx are you happy with the order in which we have these experiments sorted? Should we just sort them out alphabetically?

[Screenshot 2022-11-02 at 12 20 25](https://user-images.githubusercontent.com/26669307/199477000-5678d415-e60e-4160-805c-90a6c18d5ea1.png)

sidebars.js
adamkvitek: @aryx we should not remove docs for unsupported experiments, but move them to the deprecated experiments document (so everyone can see that this thing no longer works) in my opinion.

What do you think about an introductory file with the following content:

# Introduction

The experiments category documents experimental features and the way you can use them. As these features are experimental, some of them can be deprecated, others can become generally available (GA), meaning that GA features are fully supported parts of Semgrep.
If a feature is deprecated, its documentation is moved to (TODO link) the Deprecated experiments document. If a feature becomes GA, their docs are moved to a its relevant category outside of the experiments section.

If you see any issues with the experimental features, please [file bugs](https://github.com/returntocorp/semgrep/issues/new/choose), tweak the code, and most importantly share your thoughts!

adamkvitek: Let me know @dlukeomalley and @aryx what you think about it.

aryx: looks good to me.

aryx (approved):

docs/experiments/introduction.md
```diff
@@ -0,0 +1,14 @@
+---
+id: introduction-experiments
+slug: introduction-experiments
+title: Introduction
+hide_title: true
+append_help_link: true
+description: "Introduction of Semgrep experiments that also documents that some experiments can sunset or become GA, which means that particular documents can change their position in docs also."
+---
+
+# Introduction to Semgrep Experiments
```
pabloest, line 10:
```suggestion
# Introduction to Semgrep experiments
```

pabloest (approved):

aryx:
CI is failing: 
```[ERROR] Unable to build website for locale en.
[ERROR] Error: You are trying to create client-side redirections to paths that do not exist:
- /experiments/overview/
- ```

aryx:
Can we merge this (once CI is fixed)?  We have a few other modififications to do to the Experiments section.

adamkvitek:
@aryx these redirects are the usual error we get, maybe we need to somehow silence them as this link checker is not very helpful and only confuses contributors who think that something went wrong (cc @pabloest).

aryx:
you should fix those errors or remove the check if it's not useful. Having a red cross on each PRs is not great, because maybe sometimes it can detect a real error but it will be passed over by habit.

PR #769, opened 2022-11-04
Sara/mkt 436 - combine ssc support langs into main support langs doc

s-santillan:

### Please ensure

* For @andy-r2c - adds Yarn 2 and Yarn 3 for SSC
* Correctly reformats Supported Languages to conform with TW standards:
    * Use `<ul>` for lists, of, items, like, so
    * Use `<dl>` for definitions of terms
    * Remove instances of "above" 
    * Remove usage of ordered lists for items that are not procedures
* Merges SSC Supported languages into the main supported languages doc
    * Origin of request: https://returntocorp.slack.com/archives/C02JT6YA5S7/p1665447196744859
    * Correctly set up redirects and sidebars.js

I'm asking for a review from both a TW perspective and from @andy-r2c for his awareness.


docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+   <td>Go modules (<code>go mod</code>)</td>
+   <td><code>go.sum</code></td>
+  </tr>
+  <tr>
+   <td rowspan="2">JavaScript / TypeScript</td>
+   <td>npm (Node.js)</td>
+   <td><code>package-lock.json</code></td>
+  </tr>
+  <tr>
+   <td>Yarn, Yarn 2, Yarn 3</td>
```
andy-r2c, line 174: Should we say 'Yarn (all versions)' instead so it's easier on the eyes?

docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+   <td><code>gradle.lockfile</code></td>
+  </tr>
+  <tr>
+   <td>Maven (single-file configurations)</td>
+   <td><code>pom.xml</code>*</td>
+  </tr></tbody>
+</table>
+
+
+_* `pom.xml` is not a lockfile, however Semgrep Supply Chain parses `pom.xml` as the source of truth for dependencies in Maven projects in conjunction with `MANIFEST.mf`._
```
andy-r2c, line 213: May we also add another caveat below to include 'requirements.txt' as well? It's like the precursor to the lockfile

mmcqd:
Two things: 
 1. I think our docs are slightly incorrect when talking about python.
 2. We should mention that we support `requirements.txt`, but this is a bit weird for reasons I explain below

 
Currently we say that `Pipfile.lock` is from the package manager `pip`, but this is not entirely true. `Pipfile.lock` is generated by `pipenv`, which is a tool build _on top of_ `pip`. `pip` does nothing except grab a file from Pypi and install it in some location you specify. `pipenv` handles a creating separate virtual environments for your projects, and then installs packages on them with `pip`.

We want to add that we support the use of `requirements.txt` as a lockfile. There are two complications.

 It's a little annoying to describe what package manager this corresponds to, because it can be generated using several different tools, including `pip`, `pip-tools` and `pipreqs`. I suggest we just say `several tools` or something. Maybe just `pip`?

The other complication is that `requirements.txt` is not _always_ a lockfile, and some projects treat it like a manifest file, and include unpinned versions. So we should mention that we only support `requirements.txt` when it is being used as a lockfile: all packages are set to exact versions and ideally the file was generated automatically.

cc @s-santillan

docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+
+See [Parse rates by language](https://dashboard.semgrep.dev/).
+
+## Semgrep Supply Chain
+
+<SscIntro/>
+
+Semgrep Supply Chain parses **lockfiles** for dependencies, then scans your codebase for reachable findings based on the lockfiles. Some languages, such as Java, have several lockfiles, depending on your repository's package manager. For some languages, such as Javascript and Python, a manifest file is also parsed.
+
+The following tables list lockfiles, their package manager and language, and their support levels.
```
adamkvitek, line 153: We can provide this information just before the table, for ease of reading. (Or provide this intro again before the tables. It is a bit hard to look at the table and understand what it provides at first glance.)

Not sure as well if we are not too much in our internal code perspective here. Lockfiles should belong to the package manager of a specific language, not language to the lockfiles (also, we are reading the table from left to right, so it starts with the language).

For example: Original - `The following tables list lockfiles, their package manager and language, and their support levels.` - Can be - `The following tables list languages, their package manager and lockfiles, respective to their support levels (Generally Available or Beta).`

```suggestion
```

docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+  <tr>
+   <td><strong>Feature</strong>
+   </td>
+   <td><strong>GA</strong>
+   </td>
+   <td><strong>Beta</strong>
+   </td>
+  </tr>
+  <tr>
+   <td>Rule count
```
adamkvitek, line 235:
```suggestion
   <td>Number of rules
```

docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+## Semgrep Supply Chain
+
+<SscIntro/>
+
+Semgrep Supply Chain parses **lockfiles** for dependencies, then scans your codebase for reachable findings based on the lockfiles. Some languages, such as Java, have several lockfiles, depending on your repository's package manager. For some languages, such as Javascript and Python, a manifest file is also parsed.
+
+The following tables list lockfiles, their package manager and language, and their support levels.
+
+### General Availability
+
```
adamkvitek, line 156: It might be better to state the obvious for the ease of reading the table:

```suggestion

This table provides information about fully supported (generally available (GA)) languages, specific package managers, and their lockfiles in Semgrep Supply Chain:

```

docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+   </td>
+   <td>No commitment for new rules based on the latest security advisories.
+   </td>
+  </tr>
+  <tr>
+   <td>Semgrep engine language support
+   </td>
+   <td>Semgrep engine support is GA.
+   </td>
+   <td>Semgrep engine support is at least Beta.
```
adamkvitek, line 255: Just so I understand - by this we mean the support level of https://semgrep.dev/docs/supported-languages/#language-maturity? Maybe we can link it (this is completely optional)...

docs/supported-languages.md
s-santillan: i wrote out all the names so that people can search for them :)

docs/supported-languages.md
```diff
@@ -76,10 +137,123 @@ The following thresholds define each maturity level:
...
+  </tr>
+  <tr>
+   <td>Ruby</td>
+   <td>RubyGems</td>
+   <td><code>Gemfile.lock</code></td>
+  </tr></tbody>
+</table>
+
+### Beta
+
```
adamkvitek, line 194: @s-santillan feel free to write this sentence if it sounds confusing:

```suggestion

This table provides information about the beta level of support for languages, specific package managers, and their lockfiles in Semgrep Supply Chain:

```

docs/supported-languages.md
```diff
@@ -4,30 +4,92 @@ id: supported-languages
...
+    <dt>Feature support</dt>
+    <dd>What <a href='/writing-rules/pattern-syntax/'>Semgrep features</a> are implemented for a given language.</dd>
+    <dt>Ruleset count</dt>
+    <dd>Number of <a href='https://semgrep.dev/explore/'>Semgrep rule groupings</a> in Semgrep Registry.</dd>
+</dl>
+
+#### Levels of maturity
 
-There are 3 levels of maturity: **experimental**, **beta**, and **generally available (GA)**. Each of these maturity levels are combined with a threshold of the factors above. When a language meets the maturity threshold for each of the factors then it’s moved into that maturity level.
+There are 3 levels of maturity:
```
adamkvitek, line 59: Optional:

```suggestion
Semgrep gives 3 levels of maturity: 
```

docs/supported-languages.md
```diff
@@ -4,30 +4,92 @@ id: supported-languages
...
+<dd>Generally available languages support all advanced features such as the following:
+<ul>
+<li>All features supported in Beta</li>
+<li>Regexp</li>
+<li>Equivalence</li>
+<li>Deep expression operator</li>
+<li>Types and typing</li>
+</ul>
+</dd>
+</dl>
```
adamkvitek, line 88: This is great. Thank you. Three questions:

1. Should we remove the support level information given under our modularized table for Semgrep support?
    See below:
    Screenshot 2022-11-07 at 13 41 20

2. I am also not sure about the use of emojis in our table. Do they look ok? I mean... they are definitely cheerful. Not sure if they look completely professional. :D (We don't need to make everything dull though.)
    Screenshot 2022-11-07 at 13 41 37

3. Should I also rewrite the language support table to HTML as in the tables below? The table headers have a central alignment etc. so maybe we can unify the style. If you say yes, can I push it as a commit into this branch? 🙏

docs/supported-languages.md
```diff
@@ -62,8 +124,7 @@ The following thresholds define each maturity level:
     * Parse rate: 99.9%+
     * Rules: 10+
     * Features:
-        * All in experimental
-        * All in beta
+        * All items in Beta
         * `deep_expr_operator`
         * `dots_method_chaining`
         * `equivalence_constant_propagation`
```
adamkvitek, line 130: @s-santillan to unify style with your Supply chain support level tables below, should I convert these three bullet lists with nested lists in lists to three tables?

docs/semgrep-sc/sc-supported-langs.md
```diff
@@ -2,143 +2,11 @@
 slug: supply-chain-supported-languages 
 append_help_link: true
 description: ""
-tags:
-    - Semgrep Supply Chain 
-    - Team & Enterprise Tier
 title: Supported languages 
 hide_title: true
 ---
```
adamkvitek, line 7: Question to understand the workflow - We cannot remove a file completely when we are setting a redirect? I think that we can remove it completely...

adamkvitek (approved):
@s-santillan thank you! Left some comments!

docs/semgrep-sc/sc-supported-langs.md
adamkvitek: Or we are preparing this as a placeholder for possible modularization? So the tables and information can be included in two files? Probably not since the redirect is in place already...

docs/supported-languages.md
adamkvitek: I know that this is longer... but seems more obvious what we are talking about. Feel free to dismiss this. :-)

docs/supported-languages.md
s-santillan: I'll go with "Semgrep defines 3 maturity levels"

s-santillan:
from @mmcqd:

> Two things:
> 
> 1. I think our docs are slightly incorrect when talking about python.
> 2. We should mention that we support `requirements.txt`, but this is a bit weird for reasons I explain below
> 
> Currently we say that `Pipfile.lock` is from the package manager `pip`, but this is not entirely true. `Pipfile.lock` is generated by `pipenv`, which is a tool build _on top of_ `pip`. `pip` does nothing except grab a file from Pypi and install it in some location you specify. `pipenv` handles a creating separate virtual environments for your projects, and then installs packages on them with `pip`.
> 
> We want to add that we support the use of `requirements.txt` as a lockfile. There are two complications.
> 
> It's a little annoying to describe what package manager this corresponds to, because it can be generated using several different tools, including `pip`, `pip-tools` and `pipreqs`. I suggest we just say `several tools` or something. Maybe just `pip`?
> 
> The other complication is that `requirements.txt` is not _always_ a lockfile, and some projects treat it like a manifest file, and include unpinned versions. So we should mention that we only support `requirements.txt` when it is being used as a lockfile: all packages are set to exact versions and ideally the file was generated automatically.
> 
> cc @s-santillan

Does the following image preview work for you, @mmcqd?

docs/supported-languages.md
s-santillan: @andy-r2c Reply with any emoji if this works for you 😺

docs/supported-languages.md
s-santillan: Indeed we can. Added!

docs/semgrep-sc/sc-supported-langs.md
s-santillan: I just forgot to remove it. thanks.

docs/supported-languages.md
s-santillan: Link added.

docs/supported-languages.md
s-santillan: You can try it out and see if you like it!

docs/supported-languages.md
s-santillan: I think the emoji is appropriate for the table.

adamkvitek: @s-santillan What about points 1. and 3.

adamkvitek: Maybe we can remove it and just link to the # Levels of maturity section (by an absolute link from the module).

s-santillan: Support level != features btw so we should probably keep it ("expect bugs at this level, we won't prioritize features at this level", etc.)

andy-r2c (approved):

PR #821, opened 2022-12-13
Update taint tracking docs

IagoAbal:
Closes PA-1685
Closes PA-1784
Closes PA-2251

### Please ensure


docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+def make_tainted(a_set):
+  a_set.add(tainted)
 
-For convenience, Semgrep has a special kind of _not-conflicting_ sanitizer for this purpose, declared with `not_conflicting: true`. A pattern like `$F(...)` matches every function call. If it were acting as a regular sanitizer, it would also apply to any source or sink that had the same function-call shape. In this example, this would sanitize all calls to `sink` and [Semgrep would produce no findings at all](https://semgrep.dev/s/rYrj).
+make_tainted(my_set)
+sink(my_set)
+```
 
-If you use this approach, you instead have to enumerate your taint propagators—with `pattern-not`—if there are any:
+If we wrote the source specification below, the pattern formula would only match the occurrence of `x` that is the actual parameter of `make_tainted`, but it would not affect the occurrence of `x` that is passed to `sink`. They are the same variable, but they are different occurrences, and only the fist one is matched by the source specification. At the same time, Semgrep does not know that `make_tainted` is actually updating the variable `x` by side-effect. Thus, a taint rule using such specification would not produce any finding.
```
adamkvitek, line 95:
```suggestion
Using the source specification below, the pattern formula only matches the occurrence of `x` that is the actual parameter of `make_tainted`, but it does not affect the occurrence of `x` that is passed to `sink`. Both are the same variable but in different occurrences. Only the first one is matched by the source specification. At the same time, Semgrep does not know that `make_tainted` is updating the variable `x` by side-effect. Thus, a taint rule using such specifications that are not producing any finding.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
-<iframe src="https://semgrep.dev/embed/editor?snippet=9rJ2" border="0" frameBorder="0" width="100%" height="432"></iframe>
+```yaml
+pattern-sources:
+- patterns:
+  - pattern: make_tainted($X)
+  - focus-metavariable: $X
+```
 
-Semgrep also considers that indexing an array with a tainted index leads to tainted data. Again, it is easy to disable this via sanitizers:
+Until recently, the recommended workaround has been to write a source specification like the one below. This tells Semgrep that _every_ occurrence of `$X` after `make_tainted($X)` must be considered a source.
```
adamkvitek, line 104: Maybe we should somehow mark this as sort of a **deprecated** workaround.

```suggestion
**Deprecated workaround**: The previously recommended workaround was to write a source specification such as the one specified below. This instructs Semgrep that **every** occurrence of `$X` after `make_tainted($X)` must be considered a source.
```

adamkvitek, line 104: Note (of course this is TW work, just so you comprehend the reasoning): We also should not use words such as "until recently" in docs - the document may not be updated for the next half a year and it will stay there...

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+```yaml
+pattern-sources:
+- patterns:
+  - pattern: |
+      make_tainted($X)
+      ...
+  - pattern: $X
+```
+
+This has two main limitations. First, it will override any sanitization that may be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` will be reported as tainted despite `x` has been re-assigned to a fresh set!
```
adamkvitek, line 115:
```suggestion
This approach had two main limitations. First, it overrides any sanitization that can be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` is reported as tainted despite `x` was re-assigned to a fresh set!
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+
+```python
+make_tainted(x)
+x = set([])
+sink(x) # false positive
 ```
 
-Mini cookbook
--------------
+Also, we must note that [`...` has limitations](/writing-rules/pattern-syntax/#ellipses-and-statement-blocks) and, for example, in the code below we would not get any finding from Semgrep if we used that source specification:
```
adamkvitek, line 123:
```suggestion
Note also that [`...` ellipses operator](/writing-rules/pattern-syntax/#ellipses-and-statement-blocks) has limitations. For example, in the code below Semgrep does not match any finding if such source specification is in use:
```

docs/writing-rules/data-flow/taint-mode.md
adamkvitek, line 115: Just a note - @IagoAbal we are including this comparison between the old and a new best practice so rule creators can update their rules, right?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
 
-Again, keep in mind that sources, sanitizers and sinks are given by arbitrary patterns, so they can be anything that you can match with Semgrep. You can get very creative!
+```python
+if cond:
+    make_tainted(x)
+sink(x) # false negative
+```
 
-### Function argument as a source
+The recommended solution nowadays is to set `by-side-effect: true`, being `by-side-effect` an option that is available for each individual source in a taint rule. When this option is enabled, _and_ the source specification matches a variable (or in general, an l-value) exactly, then Semgrep will assume that the variable (or l-value) becomes tainted by side-effect at the precise places where the source specification produces a match.
```
adamkvitek, line 131:
```suggestion
It is now recommended to set `by-side-effect: true`, where `by-side-effect` is an option available for each individual source in a taint rule. When this option is enabled, and the source specification matches a variable (or in general, an l-value) exactly, then Semgrep assumes that the variable (or l-value) becomes tainted by side-effect at the precise places where the source specification produces a match.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
 
-Taint may come from specific functions that read user input such as `window.prompt()` but it is also easy to specify that taint comes from anywhere else, such as, for example, a specific argument within a function definition:
+So, this is how we specify that `make_tainted` will update its argument making future occurrences tainted:
```
adamkvitek, line 133:
```suggestion
So, this is how to specify that `make_tainted` updates its argument making future occurrences tainted:
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
 
-```javascript
+<iframe src="https://semgrep.dev/embed/editor?snippet=eRvy" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+There are also [sanitizers by side effect](#sanitizers-by-side-effect). For a good example of use of both sources and sanitizers by side-effect, take a look at rule [`c.lang.security.use-after-free.use-after-free`](https://semgrep.dev/playground/r/c.lang.security.use-after-free.use-after-free).
```
adamkvitek, line 137:
```suggestion
There are also [sanitizers by side effect](#sanitizers-by-side-effect). For a good example of the use of both sources and sanitizers by side-effect, take a look at rule [`c.lang.security.use-after-free.use-after-free`](https://semgrep.dev/playground/r/c.lang.security.use-after-free.use-after-free).
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+
+Propagators
+-----------
+
+By default, tainted data automatically propagates through assignments, operators, and function calls (from inputs to output). However, there are other ways taint can propagate, which may require language- or library-specific knowledge that Semgrep does not have built in.
```
adamkvitek, line 142:
```suggestion
By default, tainted data automatically propagates through assignments, operators, and function calls (from inputs to output). However, there are other ways in which taint can propagate, which can require language or library-specific knowledge that Semgrep does not have built-in.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+```python
+def test(s):
+    x = user_input
+    s = set([])
+    s.add(x)
+    #ruleid: test
+    sink(s.pop())
+```
+
+The following rule cannot find the above-described issue. The reason is that Semgrep does not know that executing `s.add(x)` makes `x` one of the elements in the set data structure `s`.
```
adamkvitek, line 155:
```suggestion
The following rule cannot find the above-described issue. The reason is that Semgrep does not evaluate that executing `s.add(x)` makes `x` one of the elements in the set data structure `s`.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+- pattern: $S.add($E)
+  from: $E
+  to: $S
+```
+
+In the example above, Semgrep finds the pattern `$S.add($E)`, and it checks whether the code matched by `$E` is tainted. If it is tainted, Semgrep propagates that same taint to the code matched by `$S`. Thus, adding tainted data to a set marks the set itself as tainted.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=dGRE" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+Another situation where taint propagatos may be useful is to specify in Java that, when we iterate a collection that is tainted, the individual elements must also be considered tainted:
```
adamkvitek, line 179:
```suggestion
Another situation where taint propagators are useful is to specify in Java that, when we iterate a collection that is tainted, the individual elements must also be considered tainted:
```

adamkvitek, line 179: @IagoAbal "we iterate a collection" - substitute "we" with something specific.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+- pattern: $C.forEach(($X) -> ...)
+  from: $C
+  to: $X
+```
+
+In general, a taint propagator must specify:
+1. A pattern containing **two** metavariables. These two metavariables specify where taint is propagated **from** and **to**.
+2. The `to` and `from` metavariables. These metavariables should match a **variable** or an **expression**.
+    - The `from` metavariable specifies the entry point of the taint.
+    - The `to` metavariable specifies where the tainted data is propagated to, typically an object or data structure.
```
adamkvitek, line 192: @IagoAbal If this is the most important information for the reader - which it seems so - move it under the `Propagators-----------` header.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -7,28 +7,32 @@ description: >-
...
 # Taint tracking
 
-Semgrep supports intra-procedural [taint tracking](https://en.wikipedia.org/wiki/Taint_checking). This data-flow analysis tracks the flow of untrusted (**tainted**) data throughout a program. If tainted data are not transformed or checked accordingly (**sanitized**), taint analysis reports a finding whenever tainted data reach a vulnerable function (**sink**).
+Semgrep supports intra-procedural [taint tracking](https://en.wikipedia.org/wiki/Taint_checking). This data-flow analysis tracks the flow of untrusted (**tainted**) data throughout the body of a function or method. Tainted data originate from taint **sources**. If tainted data are not transformed or checked accordingly (**sanitized**), taint analysis reports a finding whenever tainted data reach a vulnerable function (what we call a **sink**). Tainted data flow from source to sink through **propagators** (assignments, function calls, etc).
 
 The following video provides a quick overview of taint mode:
 <iframe class="yt_embed" width="100%" height="432px" src="https://www.youtube.com/embed/6MxMhFPkZlU" frameborder="0" allowfullscreen></iframe>
 
+Getting started
+---------------
```
adamkvitek, line 16: Optional note - should we rather change these headers to `## Getting started` (The same applies to Propagators and other second-level headers in this article - they are easier to search for.) @IagoAbal

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+```
+
+In general, a taint propagator must specify:
+1. A pattern containing **two** metavariables. These two metavariables specify where taint is propagated **from** and **to**.
+2. The `to` and `from` metavariables. These metavariables should match a **variable** or an **expression**.
+    - The `from` metavariable specifies the entry point of the taint.
+    - The `to` metavariable specifies where the tainted data is propagated to, typically an object or data structure.
+
+:::note
+When the `to` metavaraible matches a variable (or in general, an l-value), the taint will be propagated _by side-effect_.
```
adamkvitek, line 195:
```suggestion
When the `to` metavariable matches a variable (or in general, an l-value), the taint is propagated _by side-effect_.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+See [Sources by side effect](#sources-by-side-effect) for an explanation of what this means.
+:::
+
+For example, pattern `$S.add($E)` includes two metavariables `$S` and `$E`. Given `from: $E` and `to: $S`, and with `$E` matching `x` and `$S` matching `s`, when `x` is tainted then `s` will become tainted (by side-effect) with the same taint as `x`.
```
adamkvitek, line 199:
```suggestion
For example, pattern `$S.add($E)` includes two metavariables `$S` and `$E`. Given `from: $E` and `to: $S`, and with `$E` matching `x` and `$S` matching `s`, when `x` is tainted then `s` becomes tainted (by side-effect) with the same taint as `x`.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+
+Sanitizers
+-----------
+
+A sanitizer is specified by a pattern, and _any_ subexpression that is matched by this pattern will be regarded as sanitized.
```
adamkvitek, line 204:
```suggestion
A sanitizer is specified by a pattern, and **any** subexpression that is matched by this pattern is regarded as sanitized.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+
+:::note
+Given the sanitizer specification below, and a piece of code such as `sanitize(sink(source))`, the call `sink(source)` will _NOT_ be reported.
```
adamkvitek, line 207:
```suggestion
Given the sanitizer specification below, and a piece of code such as `sanitize(sink(source))`, the call `sink(source)` is **not** reported.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
+def check_if_safe(x):
+  if not it_is_safe(x):
+    raise Exception("unsafe")
+
+x = tainted
+check_if_safe(x)
+sink(x)
+```
+
+Until recently, the recommended workaround has been to write a sanitizer specification like the one below. This tells Semgrep that _every_ occurrence of `$X` after `check_if_safe($X)` must be considered safe.
```
adamkvitek, line 230:
```suggestion
**Deprecated workaround**: The previously recommended workaround was to write a sanitizer specification such as the one below. This instructs Semgrep that **every** occurrence of `$X` after `check_if_safe($X)` must be considered safe.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
 ```
 
 This way we tell Semgrep that we do not want arrays to be accessed with tainted indexes.
 
-### Sanitized by side-effect
+Minimizing false positives
+--------------------------
 
-Typically a sanitizer will be some function that gets tainted data and returns untainted one. But it does not need to be that way. Sometimes data gets sanitized via side-effect, and taint mode can handle this too, for example:
+Since taint mode is intra-procedural, it does not know what other functions do. By default, Semgrep assumes that taint could always propagate through other functions.
```
adamkvitek, line 276:
```suggestion
Since taint mode is intra-procedural, it does not recognize what other functions do. By default, Semgrep assumes that taint can always propagate through other functions.
```

adamkvitek: Avoid modal verbs...

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
+
+For example, in the code below, `some_safe_function` receives tainted data as input and, to be on the safe side, Semgrep assumes that it will also return tainted data as output. As a result, a finding is produced.
```
adamkvitek, line 278:
```suggestion
For example, in the code below, `some_safe_function` receives tainted data as input, and to be on the safe side, Semgrep assumes that it also returns tainted data as output. As a result, a finding is produced.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
-    - pattern-inside: |
-        $JWT.verify($TOKEN, ...)
-        ...
-    - pattern: $TOKEN
+var x = some_safe_function(tainted);
+sink(x);
 ```
 
-This example just annotates as sanitized all the occurrences of a `$TOKEN` that happen after calling `verify` on it.
+In some codebases, this conservative assumption may produce too many false positives. If that is the case, you could first consider enumerating such functions as sanitizers:
```
adamkvitek, line 285:
```suggestion
In some codebases, this conservative assumption can generate too many false positives. If that is the case, consider enumerating such functions as sanitizers:
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
 pattern-sanitizers:
-  - patterns:
-    - pattern-inside: |
-        if !strings.HasPrefix($PATH, <... $TARGET ...>, ...) {...}
-        ...
-    - pattern: $PATH
+- pattern: some_safe_function(...)
 ```
 
+If this is too cumbersome, then you can easily make Semgrep assume the opposite, that is, to assume that functions are safe by default. This is done by setting `taint_assume_safe_functions: true` under the rule [`options`](/writing-rules/rule-syntax/#options) key. If you use this approach, you instead have to enumerate your taint propagators:
```
adamkvitek, line 292:
```suggestion
If this is too cumbersome, then set Semgrep to assume that functions are safe by default. This is done by setting `taint_assume_safe_functions: true` under the rule [`options`](/writing-rules/rule-syntax/#options) key. If you use this approach, you have to enumerate your taint propagators:
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=gBD0" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+Similarly, by default, Semgrep assumes that indexing an array with a tainted index (that is, `array[tainted]`) is a tainted expression, even if the array itself is not tainted. You can set `taint_assume_safe_indexes: true` to instead assume that these expressions are safe.
```
adamkvitek, line 296:
```suggestion
Similarly, by default, Semgrep assumes that indexing an array with a tainted index (that is, `array[tainted]`) is a tainted expression, even if the array itself is not tainted. Set `taint_assume_safe_indexes: true` to instead assume that these expressions are safe.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
+
+If you want to have very fine-grained control over how taint is propagated, you can set `taint_only_propagate_through_assignments: true`, and then Semgrep will only propagate taint via trivial assignemnts of the form `x = tainted`. You would need to specify taint propagators for anything else.
```
adamkvitek, line 298:
```suggestion
If you want to have very fine-grained control over how taint is propagated, set `taint_only_propagate_through_assignments: true`. As a result, Semgrep only propagates taint through trivial assignments of the form `x = tainted`. In this case, you need to specify taint propagators for anything else.
```

adamkvitek, line 298: What do we mean by "anything else" here? @IagoAbal

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -116,37 +265,42 @@ Remember, anything can be a sink, even the index of an array access:
...
+
+:::note
+Previously we recommended the use of so-called _not conflicting sanitizers_. This feature is no longer recommended and it should be considered deprecated.
```
adamkvitek, line 301:
```suggestion
Previously, the use of so-called **not conflicting sanitizers** was recommended. This feature is now deprecated and no longer recommended.
```

adamkvitek (approved):
@IagoAbal thanks for this great input! I left some comments, suggestions, and questions. Please go through them - maybe do not merge this PR now, as fixing them later in another PR would require a lot of going back to this PR... it should not take too much time to solve these things. :-)

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Fine with me, I just think `--------` it's a bit more visual. There should also exist editor extensions that let you navigate through subsections. But I don't spend that much time in the docs so feel free to change it.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -39,63 +43,208 @@ You can find more examples of taint rules in the [Semrep Registry](https://semgr
...
 -----------------
 
-The taint engine provides basic field sensitivity support:
+The taint engine provides basic field sensitivity support, it can:
 
-- Track `x.a.b` as tainted, while you can specify that `x` and `x.a` are not tainted. If `x.a.b` is tainted, any extension of `x.a.b` (such as `x.a.b.c`) is also considered tainted.
-- If `x.a` is tainted, and later `x.a.b` is sanitized, the engine detects that `x.a.b` is not tainted but `x.a` or `x.a.c` are still tainted.
-- Taint tracking is **not** index sensitive, if `x.a[i]` is tainted, Semgrep considers `x.a` itself also as tainted. However, if `x.a[i]` is sanitized, then `x.a` is also sanitized.
+- Track that `x.a.b` is tainted, but not `x` or `x.a`. If `x.a.b` is tainted, any extension of `x.a.b` (such as `x.a.b.c`) is considered tainted by default.
+- Track that `x.a` is tainted, but remember that `x.a.b` has been sanitized. Thus the engine knows that `x.a.b` is not tainted, but `x.a` or `x.a.c` are still tainted.
```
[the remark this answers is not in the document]
IagoAbal: I picked "records" instead.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Yes. Also, some people may still prefer the old practice for whatever reason, but at least now they are more aware of the problems that it has.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Here `evaluate` sounds weird to me, I am proposing `is not aware`. Otherwise maybe something like "has no information about" is also OK.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Any other expression that acts as a taint propagator. For example, `x = taint_propagator(tainted)` or `x = tainted + "foo"` won't propagate any taint from `tainted` to `x` unless you explicitly say so, whereas in the default setting this happens automatically.

IagoAbal:
Nice suggestions @adamkvitek, as requested I won't merge just yet, give it another pass if you want.

docs/writing-rules/data-flow/taint-mode.md
adamkvitek: Thanks! Just wanted to know the context.

docs/writing-rules/data-flow/taint-mode.md
adamkvitek: I will push in a commit (as the diff here is outdated, no suggestion is possible), and substitute the text with the following (just a small change):

```
In this case, you need to specify taint propagators or any other expression that can act as a taint propagator.
```

adamkvitek: @IagoAbal I will push in a commit (as the diff here is outdated, no suggestion is possible), and substitute the text with the following (just a small change):

```md
If you want to have very fine-grained control over how taint is propagated, set `taint_only_propagate_through_assignments: true`, and then Semgrep only propagates taint through trivial
assignments of the form `x = tainted`. In this case, you need to specify taint propagators or any other expression that can act as a taint propagator. For example, `x = taint_propagator(tainted)` or `x = tainted + "foo"` does not
propagate any taint from `tainted` to `x` unless it is explicitly defined, whereas in this happens automatically in the default taint analysis setting.
```

adamkvitek: If this is misleading, feel free to rewrite this as you find best!

adamkvitek:
@IagoAbal thank you once more, feel free to merge this PR, I left one last comment for you https://github.com/returntocorp/semgrep-docs/pull/821#discussion_r1049518736.

adamkvitek (approved):

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -7,148 +7,295 @@ description: >-
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=gBD0" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+Similarly, by default, Semgrep assumes that indexing an array with a tainted index (that is, `array[tainted]`) is a tainted expression, even if the array itself is not tainted. Set `taint_assume_safe_indexes: true` to instead assume that these expressions are safe.
+
+If you want to have very fine-grained control over how taint is propagated, set `taint_only_propagate_through_assignments: true`, and then Semgrep only propagates taint through trivial assignments of the form `x = tainted`. In this case, you need to specify taint propagators or any other expression that can act as a taint propagator. For example, `x = taint_propagator(tainted)` or `x = tainted + "foo"` does not propagate any taint from `tainted` to `x` unless it is explicitly defined, whereas by default taint would propagate without explicit definition.
```
IagoAbal, line 292:
```suggestion
If you want to have very fine-grained control over how taint is propagated, set `taint_only_propagate_through_assignments: true`, and then Semgrep only propagates taint through trivial assignments of the form `x = tainted`. In this case, you need to specify taint propagators for any other expression that can act as a taint propagator. For example, with this setting `x = taint_propagator(tainted)` or `x = tainted + "foo"` do not propagate any taint from `tainted` to `x` unless it is explicitly specified, whereas by default taint would propagate implicitly.
```

PR #833, opened 2023-01-03
Sara/mkt 480 update algolia scraper and config

s-santillan:

### Please ensure

Previously, we only had general headings appear in Algolia's search results modal: Continuous Integration, CLI, Release Notes, + some others, but we **never** had a working "Semgrep App" or "Semgrep Supply Chain" heading. 

This PR fixes that. The fix is done by fixing what element the scraper picks up.

See before and after:
before-after


pabloest (approved):
Thank you, @s-santillan !

PR #838, opened 2023-01-06
added jenkins docker SAST snippet

holdenmcg:

This is a new Jenkins config tab for running Semgrep in Jenkins with Docker. This is for:

- Consistency, as all of the other configs are using Docker. It's a pain to set up with Jenkins compared to other CI providers (that usually have self-hosted, in-built support for it in the form of a build step), and I don't see people use it often, but it will be good to have it here.
- Recently a prospect asked about it, so figured it's time to add it in.

### Please ensure


s-santillan:
@holdenmcg Do we still need the "old" configuration (using `pip3 install semgrep`) - if we don't need it, we could just remove it? WDYT?

s-santillan (approved):

holdenmcg:
> @holdenmcg Do we still need the "old" configuration (using `pip3 install semgrep`) - if we don't need it, we could just remove it? WDYT?

I'm thinking we can keep it as the main one; from what I've seen, the pip3 config is the more common use case, and its nicer to look at as well 😅

s-santillan:
Ok! LGTM. Merge away! @holdenmcg

PR #850, opened 2023-01-17
set up appendix for maven

s-santillan:

### Please ensure


docs/semgrep-sc/sc-getting-started.md
```diff
@@ -167,4 +167,27 @@ Semgrep Supply Chain enables developers to perform triage and remediation throug
...
+
+Semgrep Supply Chain does not read `pom.xml` files to parse Maven projects. Instead it parses a dependency tree generated by `mvn`. Perform the following steps to enable Semgrep Supply Chain to correctly parse Maven projects:
+
+<ol>
+    <li>Generate a file outlining the project's dependency tree by adding the following command to your build pipeline:<br />
+    <pre class="language-bash">mvn dependency:tree -DoutputFile=maven_dep_tree.txt</pre>
+    For specific steps to add the command into your build pipeline, refer to your CI provider's documentation.</li>
+    <li>Run the Semgrep workflow, action, or step after the dependency tree has been generated.</li>
+</ol>
+
```
s-santillan, line 182: @andy-r2c As far as I understand, a dependency tree MUST be generated per subdirectory? Is this only for monorepositories? 

Also related: how do we handle multi-file configurations?

docs/supported-languages.md
```diff
@@ -214,11 +214,11 @@ This table provides information about the beta level of support for languages, s
   </tr>
   <tr>
    <td>Maven (single-file configurations)</td>
```
s-santillan, line 216: Previously we only handled "single-file configurations" - how has this changed? @andy-r2c 🙏

docs/semgrep-sc/sc-getting-started.md
andy-r2c: yeah, must be generated per subdirectory with a pom.xml whose contents the user wants scanned. More broadly for multi-file configurations (usually in monorepos), there's a few lines in our code that looks for the "closest" lockfile and associates the codefiles in that subdirectory with that, @mmcqd can probably share more details

docs/supported-languages.md
andy-r2c: I'm not sure -- did we only handle single-file configurations previously for maven? cc: @mmcqd

docs/semgrep-sc/sc-getting-started.md
s-santillan: If a dependency tree is generated per subdirectory, the last thing I remember is that ssc will go from the "bottomost" file directory going upward towards root, parsing each folder that has a dependency tree. In short, it will bottom-up parse ALL the dependency tree files additively and the user doesn't have to do anything except generate those `maven_dep_tree.txt` files. Is this correct? @andy-r2c @mmcqd

s-santillan: Do monorepos need to generate MANY maven_dep_tree.txt files is my question :)

docs/semgrep-sc/sc-overview.md
```diff
@@ -111,11 +111,13 @@ The following table displays differences between Semgrep and Semgrep Supply Chai
...
 
-* Semgrep Supply Chain supports scanning for transitive or indirect dependencies for all of its supported languages **except Java**. Findings are collected and displayed in **Semgrep App > Supply Chain**.
-* Semgrep Supply Chain detects transitivity only for **Python and JavaScript **languages.
-* Semgrep Supply Chain generates reachable findings only for direct dependencies.
+See [SSC glossary > Transitivity](/docs/semgrep-sc/sc-glossary/#transitive-or-indirect-dependency) for a definition of a transitive dependency.
+
+* Semgrep Supply Chain does **not** perform reachability analysis for transitive dependencies.
+* Semgrep Supply Chain supports scanning for transitive or indirect dependencies for all of its [supported languages](/docs/supported-languages#semgrep-supply-chain). Findings are collected and displayed in **Semgrep App > Supply Chain**.
+* In most cases, Semgrep Supply Chain generates **reachable findings** for **direct dependencies**. However, there are certain dependencies that are vulnerable simply through their inclusion in a codebase. Semgrep Supply Chain generates reachable findings for these types of dependencies even if they are transitive dependencies.
```
s-santillan, line 120: Hi @kurt-r2c! Do we have an example of a package this horrible?

docs/semgrep-sc/sc-getting-started.md
andy-r2c: yeah @s-santillan exactly

docs/semgrep-sc/sc-overview.md
s-santillan: Resolving with log4shell.

docs/semgrep-sc/sc-getting-started.md
s-santillan: Okay, I added in an additional step in case they have a monorepo like structure! Resolving...

docs/supported-languages.md
mmcqd: I believe this is referring to how we could not understand references within a `pom.xml` file to _other_ `pom.xml` files. I don't think this is a problem anymore. As long as a `maven_dep_tree.txt` file has been generated for every `pom.xml` the user wants us to understand, we're fine. Maven handles the hard part.

docs/semgrep-sc/sc-getting-started.md
```diff
@@ -167,4 +167,28 @@ Semgrep Supply Chain enables developers to perform triage and remediation throug
...
+Semgrep Supply Chain does not read `pom.xml` files to parse Maven projects. Instead it parses a dependency tree generated by Maven (`mvn`). Perform the following steps to enable Semgrep Supply Chain to correctly parse Maven projects:
+
+<ol>
+    <li>Generate a file outlining the project's dependency tree by adding the following command to your build pipeline:<br />
+    <pre class="language-bash">mvn dependency:tree -DoutputFile=maven_dep_tree.txt</pre>
+    For specific steps to add the command into your build pipeline, refer to your CI provider's documentation.</li>
+    <li>For monorepositories or similar directory structures, create additional dependency trees per project subdirectory. Semgrep Supply Chain can detect and parse them all.</li>
+    <li>Run the Semgrep workflow, action, or step after the dependency tree or trees have been generated.</li>
+</ol>
```
mmcqd, line 182: I think we should more clearly state that you need to generate one of these for every single `pom.xml` whose dependencies you want understood by semgrep. My impression is that even non-monorepo projects sometimes have multiple `pom.xml` files. Basically instead of "add this one command, oh and also you have to do it multiple times for monorepos" I'd say "for each pom.xml file in your project, generate a dependency tree for it using this command"

docs/semgrep-sc/sc-overview.md
```diff
@@ -111,11 +111,13 @@ The following table displays differences between Semgrep and Semgrep Supply Chai
...
 
-## Limitations of Semgrep Supply Chain scans
+## Transitive dependencies and reachability analysis
 
-* Semgrep Supply Chain supports scanning for transitive or indirect dependencies for all of its supported languages **except Java**. Findings are collected and displayed in **Semgrep App > Supply Chain**.
-* Semgrep Supply Chain detects transitivity only for **Python and JavaScript **languages.
-* Semgrep Supply Chain generates reachable findings only for direct dependencies.
+See [SSC glossary > Transitivity](/docs/semgrep-sc/sc-glossary/#transitive-or-indirect-dependency) for a definition of a transitive dependency.
+
+* Semgrep Supply Chain does **not** perform reachability analysis for transitive dependencies.
```
mmcqd, line 118: I would add an extra sentence to clarify here like "Meaning we do not scan the source code of your dependencies to determine if any of _their_ dependencies might produce a reachable finding in your code"

mmcqd (changes requested):
Overall looks great. Thank you Sara!

docs/semgrep-sc/sc-overview.md
s-santillan: Added that bit in! Thanks :)

docs/semgrep-sc/sc-getting-started.md
s-santillan: Understood! Changed those bits.

docs/supported-languages.md
s-santillan: Okay, I removed "single file configurations" now that our parser can handle all the things. :) TY!

mmcqd (approved):

PR #856, opened 2023-01-19
Use sass

pabloest:

### Please ensure


pabloest:
@s-santillan Mind checking that this builds locally for you? I used it to test our staging but rather than discard it, I figured it would be useful to have in main.

s-santillan (approved):
@pabloest Happy to use scss moving forward, just curious what we're going to use it for? (mixins, variables, proportional font computing...?) Or is this something like unifying app/main marketing website/docs stack (so that other engineers can help us or we can copy scss snippets from them)?

Approved btw! 🚀

s-santillan:
(And yes it works on my end!)

pabloest:
Thanks, @s-santillan. 

Re: using scss, potentially yes to everything you said, but only as needed. I have no vision of “migrating” to scss, but since css works as-is in sass, we can just use sass opportunistically.

PR #870, opened 2023-02-09
chore: Update the semgrep scan help text

emjin:
Add in the new flags for scan

Question: is there an equivalent ci reference to add?

Test plan: compare to `semgrep scan --help`

### Please ensure


s-santillan (approved):
LGTM! Thank you @emjin. Just curious, `-pro-languages` is ONLY Apex? No JS/TS?

emjin:
Yup. The state is:
JS/TS and Java will come with all flags but be run with normal analysis unless you have the pro engine
Apex requires the pro engine and only runs with normal OSS analysis

PR #895, opened 2023-02-27
Update SSC's Python language support

underyx:
The previous table was wrong in some parts and not matching how Python devs think about their tooling in others.


s-santillan (approved):
thank you!

PR #897, opened 2023-03-02
Update exit codes (add exit codes 3 and 14)

adamkvitek:

### Please ensure

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -338,7 +338,7 @@ export HTTPS_PROXY="http://10.10.1.10:1080"
 
 <!-- Source code reference - the exit codes are located in the Semgrep repository - https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/error.py. -->
 
-Semgrep OSS can finish with the following exit codes:
+Semgrep can finish with the following exit codes:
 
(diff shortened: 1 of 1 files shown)


docs/cli-reference.md
```diff
@@ -338,19 +338,20 @@ export HTTPS_PROXY="http://10.10.1.10:1080"
 
 <!-- Source code reference - the exit codes are located in the Semgrep repository - https://github.com/returntocorp/semgrep/blob/develop/cli/src/semgrep/error.py. -->
 
-Semgrep can finish with the following exit codes:
+Semgrep OSS can finish with the following exit codes:
```
adamkvitek, line 341: Should we say OSS (as these exit codes are coming from the OSS part of our software), or just Semgrep to emphasize that whatever product our users will use, these are always the exit codes they can expect? @milanwilliams

milanwilliams: I'd say just Semgrep! :)

milanwilliams (approved):

docs/cli-reference.md
adamkvitek:
```suggestion
Semgrep can finish with the following exit codes:
```

PR #903, opened 2023-03-04
updated slack app to latest instructions

s-santillan:

### Please ensure

Our slack app is not yet approved (it will be soon) but until then we should let users know they have to be workspace owners to install the app.


vivekkhimani (approved):
LGTM! 🚀 Thanks for keeping these up to date! 🙇

PR #904, opened 2023-03-04
updated with member scoped access token

s-santillan:

### Please ensure

~~I took a look at the screenshot @zyannes and it looks like it doesn't show personal data though am checking with you first Zach.~~ (fixed)

~~BTW - when you say a user can publish rules, I wasn't aware we had a way of publishing rules to the rule board from the CLI? Or did you mean something else?~~

What the above means: member scoped users can't update the rule board. This has been reflected in the docs (mar 13)

Please let me know if the text is all good.


src/css/custom.scss
```diff
@@ -189,6 +189,14 @@ img[src*="bordered"] {
     display: block;
 }
 
+img[src*="md-width"] {
```
pabloest, line 192: Nice!

pabloest (approved):

zyannes (approved):

PR #908, opened 2023-03-15
check version to confirm installation

milanwilliams:
From what I've seen, most tools do a version check to confirm installation instead of running a pattern. While it is nice to see a scan, I worry the syntax is a bit jarring on first impression. What do folks think?

### Please ensure

- [] A subject matter expert (SME) reviews the content
- [] Redirects are added if the PR changes page URLs

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -76,7 +76,7 @@ Semgrep CLI installation requires Python 3.7 or later.
 
 2. Confirm installation by running a simple search pattern. For example, run the following command:
     ```sh
-    semgrep --pattern '127.$A.$B.$C' --lang generic /etc/hosts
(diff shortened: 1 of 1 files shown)


docs/getting-started.md
```diff
@@ -76,7 +76,7 @@ Semgrep CLI installation requires Python 3.7 or later.
 
 2. Confirm installation by running a simple search pattern. For example, run the following command:
```
adamkvitek, line 77:
```suggestion
2. Confirm installation by the following command:
```

adamkvitek (approved):
@milanwilliams thanks! Good idea. Maybe we can simplify the description before the code snippet as we are no longer running a simple search pattern (suggestion is ready).

PR #913, opened 2023-03-23
Fix title rendering for "Removing rules or rulesets"

armchairlinguist:
The "Removing rules or rulesets" heading was showing as raw Markdown with ## instead of a rendered H2, due to a missing line break; that should be fixed now.

### Please ensure

- [x] A subject matter expert (SME) reviews the content (no content changes)


adamkvitek (approved):
@armchairlinguist thanks!

PR #928, opened 2023-03-31
Fix reference to Bitbucket in BuildKite section

armchairlinguist:
There is a reference to Bitbucket in the Buildkite section of "Sample CI configurations" - it should probably just refer to Buildkite.

### Please ensure


adamkvitek (approved):
@armchairlinguist thank you!

PR #933, opened 2023-04-04
update links

s-santillan:

### Please ensure

----

I went through and tested the links, some of them are redirecting on click AND in new tab, some of them are not. I'm not sure if it has something to do with links like: doc-name/#header-sections-in-links.

I tested the links with yarn build then yarn serve.

I also double checked with the docusaurus plugin for redirects.

I decided to simply find and replace the links. I then caught 2 docs we haven't renamed, which are fixed in this PR.


s-santillan:
Update: I have decided to separate this PR into two so that half of it can be merged asap. @pabloest I ran yarn build and no broken redirects appeared, and checked that links are working. There's another half of it coming tomorrow.

pabloest (approved):

PR #940, opened 2023-04-06
last redirects in src components

s-santillan:

### Please ensure

**⚠️ Note!** I noticed some links that had to be updated, which increased the number of files changed but are not relevant to the subject of this PR, **so I will write here in the comment the only files you need to look at ⚠️**

This PR:

1. Removes the step to authorize semgrep-app when a new user first logs in to Semgrep. (`src/components/procedure/_platform-signin-github`)
2. Updates `/semgrep-cloud-platform/user-management` with updated steps for **Adding an organization**
3. Updates the screenshot for Source code managers. (`app-scm.png`)

@zyannes Let me know if this is a good first start. My understanding is that the main change is no longer having to auth semgrep-app right at the beginning AND that joinin an org is now in the user's control. I'm not sure of some ramifications of this new flow, I did however test parts of the update (Create an org flow) for this update.


docs/semgrep-cloud-platform/user-management.md
```diff
@@ -54,7 +54,18 @@ To create an organization account:
...
 1. Sign in to Semgrep Cloud Platform.
 2. On the sidebar, click **your account name > Add org**.
-3. Complete the steps to finish adding your organization account.
+3. Choose either Create an organization or Join an organization.
+4. For **Create an organization**, follow these steps:
+    1. Enter the **Organization Display Name**.
+    2. Click **Create**.
+    3. Click **your account name** > **the new organization** you just created.
+    3. Click **Settings** > **Source code**.
+    4. Click on your Source code manager, for example Connect to GitHub.
```
zyannes, line 63: Do we want to mark these as "optional"? If not, we may want to explain what the purpose of "Connecting an SCM" does: it allows your Semgrep organization's membership to be managed by GH/GL. i.e. after linking your org, other members of your GH Organization/GL Group should be able to join your org by following step 5 below.

s-santillan: I see what you mean, I'll separate the procedure into 2 as they do 2 different things. Thanks!

adamkvitek (approved):
@s-santillan approving so you can merge any time when this PR is ready to go.

zyannes (approved):

PR #969, opened 2023-04-19
Clarify current join mode behavior

armchairlinguist:
### Please ensure

### DIFF
--- a/docs/writing-rules/experiments/join-mode/overview.md
+++ b/docs/writing-rules/experiments/join-mode/overview.md
@@ -196,7 +196,7 @@ Short for references, `refs` is a list of external rules that make up your code 
 
 ### `rule`
 
-This points to an external rule location to use in this join rule. Currently, join mode requires external rules. Additionally, even though Semgrep rule files can typically contain multiple rules under the `rules` key, join mode **will only use the first rule in the file**.
+This points to an external rule location to use in this join rule. Even though Semgrep rule files can typically contain multiple rules under the `rules` key, join mode **will only use the first rule in the provided file**.
 
 Anything that works with `semgrep --config <here>` will work as the value for `rule`.
 
@@ -210,7 +210,7 @@ Renaming is necessary for named expressions, e.g., `$...EXPR`.
 
 ### `as`
 
-An optional key for an object in `refs`, `as` will let you specify an alias for the results collected by this rule for use in the `on` conditions. Without the `as` key, the default name for the results collected by this rule is the rule ID of the rule in `rule`. If you use `as`, the results can be references using the alias specified by `as`.
+An optional key for an object in `refs`, `as` will let you specify an alias for the results collected by this rule for use in the `on` conditions. Without the `as` key, the default name for the results collected by this rule is the rule ID of the rule in `rule`. If you use `as`, the results can be referenced using the alias specified by `as`.
 
 ### `on`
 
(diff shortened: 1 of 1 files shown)


docs/writing-rules/experiments/join-mode/overview.md
```diff
@@ -196,7 +196,7 @@ Short for references, `refs` is a list of external rules that make up your code
 
 ### `rule`
 
-This points to an external rule location to use in this join rule. Currently, join mode requires external rules. Additionally, even though Semgrep rule files can typically contain multiple rules under the `rules` key, join mode **will only use the first rule in the file**.
+This points to an external rule location to use in this join rule. Even though Semgrep rule files can typically contain multiple rules under the `rules` key, join mode **will only use the first rule in the provided file**.
```
s-santillan, line 199: Just checking that I understood this correctly - we no longer require external rules? Does that mean join mode will accept a pattern operator?

docs/writing-rules/experiments/join-mode/overview.md
```diff
@@ -237,7 +237,9 @@ The `on` key is required in join mode. This is where the join conditions are lis
 
 Join mode **is not taint mode**! While it can look on the surface like join mode is "connecting" things together, it is actually just creating sets for each Semgrep rule and returning all the results that meet the conditions. This means some false positives will occur if unrelated metavariable contents happen to have the same value.
 
-Right now, external rules are required for join mode. (This is why the rules are called `refs` - they are references to other locations.) To use join mode, you must define your individual Semgrep rules in independent locations. This can be anything that works with `semgrep --config <here>`, such as a file, a URL, or a Semgrep registry pointer like `r/java.lang.security.some.rule.id`. 
+To use join mode with `refs`, you must define your individual Semgrep rules in independent locations. This can be anything that works with `semgrep --config <here>`, such as a file, a URL, or a Semgrep registry pointer like `r/java.lang.security.some.rule.id`. 
+
+Join mode does not work in the Semgrep playground, as it is an experimental feature.
```
s-santillan, line 242:
```suggestion
Join mode does not work in the Semgrep playground or editor, as it is an experimental feature.
```

s-santillan (approved):
thanks!

docs/writing-rules/experiments/join-mode/overview.md
adamkvitek, line 199:
```suggestion
The `rule` points to an external rule location to use in this join rule. Even though Semgrep rule files can typically contain multiple rules under the `rules` key, join mode **only uses the first rule in the provided file**.
```

docs/writing-rules/experiments/join-mode/overview.md
```diff
@@ -196,7 +196,7 @@ Short for references, `refs` is a list of external rules that make up your code
 
 ### `rule`
```
adamkvitek, line 197: @armchairlinguist @minusworld should this actually be `rules`? I don't see `rule:` key in the examples on this page.

docs/writing-rules/experiments/join-mode/overview.md
```diff
@@ -196,7 +196,7 @@ Short for references, `refs` is a list of external rules that make up your code
```
adamkvitek, line 196: @minusworld on line 193 - <code>### `refs`</code> are these references now? I also don't see `refs:` in any example in this file.

docs/writing-rules/experiments/join-mode/overview.md
adamkvitek: (Just building on Sara's suggestion)
```suggestion
Join mode does not work in the Semgrep Playground or Semgrep Editor, as it is an experimental feature.
```

docs/writing-rules/experiments/join-mode/overview.md
adamkvitek: @armchairlinguist not sure about this edit. @minusworld can you please have a look here?

adamkvitek (commented):
@armchairlinguist thanks for catching some things here, I think we should revisit this document and maybe update it a bit. Left some comments and asked @minusworld for some help.

docs/writing-rules/experiments/join-mode/overview.md
armchairlinguist: It seems a little bit redundant to me to refer back to `rule` when it's also the entirety of the heading. I like the other change, though.

armchairlinguist: I was able to come up with a way to phrase this that's more analogous to the other sections, hoping it works!

armchairlinguist: @s-santillan Yes, the document gets into this a little further down, discussing rules vs refs. I've found that both work, so this restriction doesn't apply anymore.

armchairlinguist:
Thanks everyone for the feedback! I did some style cleanup related to the suggestions (effectively incorporating most of them, but going a little further), so requesting a fresh review before merging.

s-santillan (approved):
lgtm. sorry about the delay.

PR #970, opened 2023-04-20
trueup, license reconcilliation

s-santillan:

### Please ensure

**Changes from draft:**
* I noticed an inconsistency in the draft. We say "number of purchased license**s**" then say "purchase an annual license for 50 developers" in the example. This has been corrected to "purchase annual licenses for 50 developers".
* Created a header for the example and made it clear for the right-side TOC
* Replaced "Should the organization exceed the number..." with "If the organization exceeds the number..." for conciseness and clarity.
* Removed "we recommend" because :
    * If we use "we" we should state who we are referring to before hand (I can update if we choose to do this, something like "Semgrep Inc recommends...")
    * It's not actually a recommendation 😅 - if they plan to go over their plan limits, they will **have** to talk to us. (That being said, the previous alternative stated, "Semgrep Inc. recommends..." is also a viable phrasing and less harsh on a sensitive concern. 

**Given this addition, do we want to keep the "Plan limits" section? (See screenshot - it looks a little deprecated)**

### DIFF
--- a/docs/semgrep-cloud-platform/pricing-and-billing.mdx
+++ b/docs/semgrep-cloud-platform/pricing-and-billing.mdx
@@ -279,11 +279,28 @@ Notes:
 
 ### Number of developers
 
-Within your team or organization, assess the number of members that make commits. That determines the number of developers needed for the plan purchase.
+Within your team or organization, assess the number of members that make commits. That determines the number of **licenses** needed for the plan purchase.
 
-For example, if a project has 4 unique developers who create commits during the billing period while Semgrep is scanning their repositories, only 4 developers are required even if the organization has a total of 10 members. If these unique developers commit to many projects within the same organization, they are counted once, so no additional cost is charged.
+For example, if a project has 4 unique developers who create commits during the billing period while Semgrep is scanning their repositories, only 4 licenses are required even if the organization has a total of 10 members. If these unique developers commit to many projects within the same organization, they are counted once, so no additional cost is charged.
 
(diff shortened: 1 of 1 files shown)


docs/semgrep-cloud-platform/pricing-and-billing.mdx
```diff
@@ -279,11 +279,28 @@ Notes:
...
+For example, if a project has 4 unique developers who create commits during the billing period while Semgrep is scanning their repositories, only 4 licenses are required even if the organization has a total of 10 members. If these unique developers commit to many projects within the same organization, they are counted once, so no additional cost is charged.
+
+All members of the organization, regardless of developer (license) status, have access to paid features for the chosen tier. This means that project managers and other non-programming roles can still view the Semgrep Cloud Platform dashboard.
+
+### Semgrep add-on reconciliation of licenses
+
+If the organization exceeds the number of purchased licenses, the organization will be charged based on the number of licenses that exceeded the amount purchased. The additional charge starts the month after the use of licenses exceeds the contracted amount.
```
adamkvitek, line 290: Optional:
```suggestion
If the organization exceeds the number of purchased licenses, the organization will be charged based on the number of licenses that exceeded the purchased amount. The additional charge starts the month after the use of licenses exceeds the contracted amount.
```

pabloest:
I don’t have a strong opinion on the plan limits section so I’ll approve so you can move forward in the way you choose.

pabloest (approved):

PR #981, opened 2023-05-02
Add branching limitation for symbolic propagation

armchairlinguist:
Adds a known limitation of symbolic propagation with branching, for clarity.

### Please ensure

### DIFF
--- a/docs/writing-rules/experiments/symbolic-propagation.md
+++ b/docs/writing-rules/experiments/symbolic-propagation.md
@@ -27,3 +27,24 @@ If we tried to match the pattern `pandas.DataFrame(...).index.set_value(...)` ag
 Symbolic propagation is a generalization of [constant propagation](/writing-rules/data-flow/constant-propagation/) that addresses this limitation. It enables Semgrep to perform matching modulo variable assignments. Thus, Semgrep is then able to match both `test1` and `test2` with the same simple pattern. This feature needs to be enabled explicitly via rule `options:` by setting `symbolic_propagation: true`.
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=JeBP" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+## Limitations
+
+Currently, symbolic propagation does not cross branching boundaries, such as `if` clauses or loops. This helps avoid difficulties with managing the parse graph for the code. Consider the following Python code, adapted from the example shown above:
+
+```python
+import pandas
+
+def test1():
+    # ruleid: test
+    pandas.DataFrame(x).index.set_value(a, b, c)
+
+def test2():
+    if (x < 5):
+        df = pandas.DataFrame(x)
+    ix = df.index
+    # ruleid: test
+    ix.set_value(a, b, c)
+```
+
+In this case, even if `symbolic_propagation: true` is used, Semgrep does not match `test2`, because the assignment of `df` to `pandas.DataFrame(x)` is within a conditional, and is not propagated to the final two lines.
--- a/docs/writing-rules/experiments/symbolic-propagation.md
+++ b/docs/writing-rules/experiments/symbolic-propagation.md
@@ -28,9 +28,9 @@ Symbolic propagation is a generalization of [constant propagation](/writing-rule
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=JeBP" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
-## Limitations
+## Limitations of symbolic propagation
 
-Currently, symbolic propagation does not cross branching boundaries, such as `if` clauses or loops. This helps avoid difficulties with managing the parse graph for the code. Consider the following Python code, adapted from the example shown above:
+Currently, symbolic propagation does not cross branching boundaries, such as `if` clauses or loops. Consider the following Python code, adapted from the example shown above:
 
 ```python
 import pandas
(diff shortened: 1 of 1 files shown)


docs/writing-rules/experiments/symbolic-propagation.md
```diff
@@ -27,3 +27,24 @@ If we tried to match the pattern `pandas.DataFrame(...).index.set_value(...)` ag
 Symbolic propagation is a generalization of [constant propagation](/writing-rules/data-flow/constant-propagation/) that addresses this limitation. It enables Semgrep to perform matching modulo variable assignments. Thus, Semgrep is then able to match both `test1` and `test2` with the same simple pattern. This feature needs to be enabled explicitly via rule `options:` by setting `symbolic_propagation: true`.
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=JeBP" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+## Limitations
+
+Currently, symbolic propagation does not cross branching boundaries, such as `if` clauses or loops. This helps avoid difficulties with managing the parse graph for the code. Consider the following Python code, adapted from the example shown above:
```
armchairlinguist, line 33: @IagoAbal I'd particularly love your feedback on the best way to phrase this (or if it's even important to explain the reason).

IagoAbal: I would drop this (and it's not 100% correct anyways):
> This helps avoid difficulties with managing the parse graph for the code.

docs/writing-rules/experiments/symbolic-propagation.md
```diff
@@ -27,3 +27,24 @@ If we tried to match the pattern `pandas.DataFrame(...).index.set_value(...)` ag
...
+
+def test2():
+    if (x < 5):
+        df = pandas.DataFrame(x)
+    ix = df.index
+    # ruleid: test
+    ix.set_value(a, b, c)
+```
+
+In this case, even if `symbolic_propagation: true` is used, Semgrep does not match `test2`, because the assignment of `df` to `pandas.DataFrame(x)` is within a conditional, and is not propagated to the final two lines.
```
IagoAbal, line 50: I would use this example instead:
```python
 def test2():
     df = pandas.DataFrame(x)
     if (x < 5):
        pass
     ix = df.index
     # ruleid: test
     ix.set_value(a, b, c)
 ```
So even if `df` is defined outside the `if` block, its symbolic value is lost after the `if` block.

File: docs/writing-rules/experiments/symbolic-propagation.md [hunk 3]
```diff
@@ -27,3 +27,24 @@ If we tried to match the pattern `pandas.DataFrame(...).index.set_value(...)` ag
 Symbolic propagation is a generalization of [constant propagation](/writing-rules/data-flow/constant-propagation/) that addresses this limitation. It enables Semgrep to perform matching modulo variable assignments. Thus, Semgrep is then able to match both `test1` and `test2` with the same simple pattern. This feature needs to be enabled explicitly via rule `options:` by setting `symbolic_propagation: true`.
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=JeBP" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+## Limitations
```
adamkvitek, line 31: Maybe specific header can help our search:

```suggestion
## Limitations of symbolic propagation
```

docs/writing-rules/experiments/symbolic-propagation.md
armchairlinguist: That sounds good, I'll fix it up by ✂️ that out entirely.

docs/writing-rules/experiments/symbolic-propagation.md
armchairlinguist: @adamkvitek I hadn't done this because it looks a little strange in the page context and isn't done in other similar pages like [Join mode](https://semgrep.dev/docs/writing-rules/experiments/join-mode/overview/#limitations), but if you think it's worth it for search, I don't mind!

docs/writing-rules/experiments/symbolic-propagation.md
armchairlinguist: I see, is that to show that it doesn't have to be within the conditional, there just has to be a conditional present? That is a more general limitation, so it makes sense to me, just want to confirm.

IagoAbal: Yes, any symbolic value that is known before the `if (x < 5)`, or introduced in its body, will be lost after the `if`. The control-flow graph has "join" points, where two or more branches get "merged"/"joined". At those "join" points, right now Semgrep forgets every symbolic-propagated value. As of now symbolic values are only propagated through sequential control-flow.

IagoAbal: I'll try to relax this restriction soon, maybe next week, so maybe we need to update the docs soon. 🤷‍♂️

armchairlinguist: If it's that soon maybe it's not worth merging! I wasn't sure how long it might be. If there will still be a restriction of some sort we can go ahead and then patch it later, though.

docs/writing-rules/experiments/symbolic-propagation.md
armchairlinguist: OK! added that in my update, just wanted to check.

docs/writing-rules/experiments/symbolic-propagation.md
IagoAbal: I can't make promises unless that issues becomes high-prio, and the work here is done I think, so I would just merge this.

adamkvitek (approved):
@armchairlinguist lgtm!

PR #995, opened 2023-05-09
Semgrep Pro Engine - instruction fixes and wording clarity

armchairlinguist:
### Please ensure

### DIFF
--- a/docs/semgrep-code/semgrep-pro-engine-intro.md
+++ b/docs/semgrep-code/semgrep-pro-engine-intro.md
@@ -191,8 +191,8 @@ If many repositories cause scan issues:
 
 Semgrep Pro Engine is different from [join mode](/writing-rules/experiments/join-mode/overview/), which also allows you to perform interfile analyses by letting you join on the metavariable matches in separate rules. Join mode is an experimental feature which is not actively developed or maintained. You may encounter many issues while using join mode.
 
-### Future development of Semgrep Pro Engine
+### Feedback for Semgrep Pro Engine
 
-We’re excited to hear what’s on your mind. As you explore Semgrep Pro Engine, we want to know what you'd like to be able to capture with Pro Engine. We believe that this deeper analysis helps users find more vulnerabilities, build trust with developers, and enforce code standards quickly. Let us know what you think about the results in the <a href="https://go.semgrep.dev/slack">Semgrep Community Slack</a>.
+The team at Semgrep is excited to hear what’s on your mind. As you explore Semgrep Pro Engine, we want to know what you'd like to be able to capture with Pro Engine. We believe that this deeper analysis helps users find more vulnerabilities, build trust with developers, and enforce code standards quickly. Let us know what you think about the results in the <a href="https://go.semgrep.dev/slack">Semgrep Community Slack</a>.
 
 <MoreHelp />
--- a/docs/semgrep-code/semgrep-pro-engine-intro.md
+++ b/docs/semgrep-code/semgrep-pro-engine-intro.md
@@ -1,7 +1,7 @@
 ---
 slug: semgrep-pro-engine-intro
 append_help_link: true
-description: "This article introduces Semgrep Pro Engine, guides you through installation and provides some additional information."
+description: "This article introduces Semgrep Pro Engine, guides you through installation, and provides some additional information."
 hide_title: true
 title: Semgrep Pro Engine overview
 ---
@@ -39,7 +39,7 @@ To install and run Semgrep Pro Engine in the CLI, follow these steps:
     ```sh
     semgrep login
     ```
-1. Follow the link that Semgrep CLI printed in the command line.
+1. Follow the link that Semgrep CLI printed on the command line.
 1. To install Semgrep Pro Engine use the following command:
     ```sh
     semgrep install-semgrep-pro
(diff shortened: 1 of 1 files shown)


docs/semgrep-code/semgrep-pro-engine-intro.md
```diff
@@ -179,20 +179,20 @@ rules:
...
 
 ### Future development of Semgrep Pro Engine
 
-We’re excited to hear what’s on your mind. As users explore the limits of Semgrep Pro Engine, we want to know what they’re failing to express. We believe that this deeper analysis helps users find more vulnerabilities, build trust with developers, and enforce code standards quickly. Let us know what you think about the results in the <a href="https://go.semgrep.dev/slack">Semgrep Community Slack</a>.
+We’re excited to hear what’s on your mind. As you explore Semgrep Pro Engine, we want to know what you'd like to be able to capture with Pro Engine. We believe that this deeper analysis helps users find more vulnerabilities, build trust with developers, and enforce code standards quickly. Let us know what you think about the results in the <a href="https://go.semgrep.dev/slack">Semgrep Community Slack</a>.
```
armchairlinguist, line 196: I'd love to hear other thoughts on how to structure this paragraph. For example, maybe we want to move the "We believe" to the beginning, or even remove it? It doesn't seem directly relevant to the topic "Future development".

s-santillan: Hi @armchairlinguist, thank you for looking into this! My thoughts are:

* I think what this heading is really asking for is feedback and is building a case for transparency (hence the "we believe..." line.) Maybe we could use a header such as "Developing Semgrep Pro Engine through feedback" or simply "Feedback for Semgrep Pro Engine"
* Speaking from a "strict" TW standpoint, marketing copy ("We believe X") is used in marketing collateral. I personally am not very strict when a doc is created around launch time (as this one was), as the only collateral to send to users **is** the docs and we try to be more approachable. As the product matures, the content usually evolves organically to be more of an "official handbook". @milanwilliams What do you think about removing it?
* Optional change: We could identify "we" at first use, such as "The team at Semgrep is excited to hear what's on your mind." (This is in line with the [Google Documentation Style Guide](https://developers.google.com/style/pronouns).)

Thank you for opening this discussion.

docs/semgrep-code/semgrep-pro-engine-intro.md
```diff
@@ -48,11 +48,11 @@ To install and run Semgrep Pro Engine in the CLI, follow these steps:
     ```bash
     semgrep --pro --config "p/default" 
     ```
-1. Optional: We appreciate your help gathering data as we improve Semgrep Pro Engine! If you are fine with sending r2c usage metrics, run the following command:
+1. Optional: If you prefer not to send usage metrics to Semgrep, run the following command:
```
s-santillan, line 51: Looking for Milan's opinion on this one.

docs/semgrep-code/semgrep-pro-engine-intro.md
```diff
@@ -138,11 +138,11 @@ To enable Semgrep Pro Engine in the Semgrep Cloud Platform, follow these steps:
 1. Sign in to [Semgrep Cloud Platform](https://semgrep.dev/login).
 1. Select **[Settings](https://semgrep.dev/orgs/-/settings)**.
 1. Enable the <i class="fa-solid fa-toggle-large-on"></i> **Semgrep Pro Engine beta** toggle.
-1. Ensure that you have the **default ruleset** added in your **[Rule Board](https://semgrep.dev/orgs/-/board)**. In the Rule Board, the **default ruleset** is in the **Monitor column**. If this ruleset is **not** added, go to [https://semgrep.dev/p/default](https://semgrep.dev/p/default), and then click **Add to Rule Board**.
+1. Ensure that you have the **default ruleset** added in your **[Rule Board](https://semgrep.dev/orgs/-/board)**. In the Rule Board, the **default ruleset** is recommended for the **Monitor column**. If this ruleset is **not** added, go to [https://semgrep.dev/p/default](https://semgrep.dev/p/default), and then click **Add to Rule Board**.
```
s-santillan, line 141: For @armchairlinguist :

The difference between your change ("recommended for") and the existing content is that it's unclear if you're telling the user "put default ruleset in Monitor" or "default ruleset is typically found in Monitor". What the step was trying to say is that "this is where you usually find it, make sure it's there, if it's not there, add it and put it in Monitor." Hope this helps.

s-santillan (approved):
Approved on TW end with some minor comments!

docs/semgrep-code/semgrep-pro-engine-intro.md
armchairlinguist: I think the way it's currently written doesn't say what it's trying to either, though it's possible I didn't make things better. 😆 

"the default ruleset is in the Monitor column" to me means that it **is** there, and cannot/will not be anywhere else if you have added it.

The situation, as I understand it, is that default and comment are the two rulesets that are currently targeted for Pro rules. The recommended placement is Monitor for default and Comment for comment. If you don't have the default ruleset at all, you might not get many results. (This is the primary situation the direction is trying to avoid.) If you have it somewhere (but not in the recommended place), you will get results, but they might be noisy. So the most important thing is that you have it, but also the placement that is recommended is Monitor.

Perhaps one option is:

```suggestion
1. Ensure that you have the **default ruleset** added in your **[Rule Board](https://semgrep.dev/orgs/-/board)**. If this ruleset is **not** added, go to [https://semgrep.dev/p/default](https://semgrep.dev/p/default), and then click **Add to Rule Board**. For best results, this ruleset should be placed in the Monitor column.
```
Thoughts?

s-santillan: Yes, agreed. Feel free to commit your suggestion.

docs/semgrep-code/semgrep-pro-engine-intro.md
milanwilliams: For my 2c - I like the update you made @armchairlinguist! And agree we should update the heading to make it clearer, I liked `Feedback for Semgrep Pro Engine` feels short & to the point :)

docs/semgrep-code/semgrep-pro-engine-intro.md
milanwilliams: Good by me!

milanwilliams (approved):

PR #997, opened 2023-05-10
add workflow dispatch

s-santillan:

### Please ensure

Minor update that does not require other items in the list.

~~- [ ] A technical writer reviews the content or PR~~
~~- [ ] This change has no security implications or else you have pinged the security team~~
~~- [ ] Redirects are added if the PR changes page URLs~~
~~- [ ] If you have changed any header tag links (doc/#this-kind-of-anchor), update all instances of that link~~

In addition to adding workflow dispatch, I also renamed "Semgrep Code" config snippets to "default" as it now runs both products and is equivalent to the default config file that is seen in SCP (with additional comments.)

Snag_2dddef8


adamkvitek (approved):
@s-santillan lgtm, thank you!

armchairlinguist (approved):
LGTM! If you want to be super proactive, you could also update the name of the job to `semgrep/ci` from `Scan` - I recently did a follow-up PR to update that too :) But I can also do that in a separate change

PR #1011, opened 2023-05-18
[NEW KB] How to run all available rules within the registry.

Jason-Lim00:
How to run all available rules within the registry.

This kb talks about the use of `--config=r/all` when wanting to run all available rules within the registry on a repository.
  
### Please ensure

### DIFF
--- a/run-all-available-rules.md
+++ b/run-all-available-rules.md
@@ -1,0 +1,51 @@
+---
+description: How do I run all available rules on my repository?
+tags:
+  - Rules
+  - Semgrep Registry
+  - Semgrep Code
+---
+
+# How to run all available rules on a repository.
+
+If you would like to run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) and are not willing to add all available rules/rulesets to the Rule Board or add a variety of `--config` entries, what you can do is run the following command:
+
+```
+semgrep --config=r/all
+```
+
+This command will pull out all avaialable public, private, and pro tiered rules if you have access to them. Below is an example output:
+
+```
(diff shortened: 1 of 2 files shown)


run-all-available-rules.md
```diff
@@ -0,0 +1,51 @@
...
+
+# How to run all available rules on a repository.
+
+If you would like to run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) and are not willing to add all available rules/rulesets to the Rule Board or add a variety of `--config` entries, what you can do is run the following command:
+
+```
+semgrep --config=r/all
+```
+
+This command will pull out all avaialable public, private, and pro tiered rules if you have access to them. Below is an example output:
```
s-santillan, line 17:
```suggestion
This command will pull out all available public, private, and pro tiered rules if you have access to them. Below is an example output:
```
Curious about the scope of this command: Is the goal of r/all to also run private rules from _local_ sources, such as .semgrep?

Thank you Jason.

s-santillan (approved):
Thanks for pinging us, @Jason-Lim00 !

This looks super helpful, am approving it. So that our user's expectations of this command are managed, maybe let's also mention any caveats (what does r/all NOT include, aside from the Rule board? Does it run local rules from the dev machine? What are private rules - are they those rules that are _not_ in the Rule board, but in the library of custom self-written / org-written rules? Questions like that.)

Note: private rules != local rules, which is why I didn't want to assume its behavior.

Anyway, this is the KB, so the TWs are in an advisory role. Looking forward to the first release!

PR #1012, opened 2023-05-19
Rename interfile and interprocedural

adamkvitek:

### Please ensure


IagoAbal (changes requested):
We agreed that in technical docs we'd use inter-procedural and inter-file mainly, but cross-function and cross-file would appear near those terms in some key places as in "inter-file (cross-file)".

PR #1016, opened 2023-05-23
Sara/cse 422 suggestions

s-santillan:

### Please ensure

Hi @Jason-Lim00 ! I made a separate branch from yours so you could see the kind of edits 

The content is fantastic and I can see the need for `r/all`. The edits I'm suggesting are from a TW standpoint, focusing on 2 goals:

* Skimmability (content is short and easy to read).
* All docs and kb have a unified "tone". This is so that people aren't jarred when they move from one doc to another. 

I'll put my comments in the lines I've changed so you can go over them and resolve if you like.

Here's what the changes look like:

Here's your original:

Thanks again for your contribution!

### DIFF
--- a/docs/kb/rules/run-all-available-rules.md
+++ b/docs/kb/rules/run-all-available-rules.md
@@ -18,7 +18,7 @@ This command runs the following rules from the Semgrep Registry:
 
 * All public or community-written rules
 * Your organization's private rules, if you are logged in 
-* Semgrep Pro rules, if you have access to them.
+* Semgrep Pro rules, if you have access to them
 
 Refer to the following snippet for a sample output of `config=r/all` while logged in through `semgrep ci`:
 
--- a/docs/kb/rules/run-all-available-rules.md
+++ b/docs/kb/rules/run-all-available-rules.md
@@ -6,7 +6,7 @@ tags:
   - Semgrep Code
 ---
 
-# Run all available rules on a repository
+# Running all available rules on a repository
 
 To run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) without using the Rule board, enter the following command:
 
--- a/docs/kb/rules/run-all-available-rules.md
+++ b/docs/kb/rules/run-all-available-rules.md
@@ -6,15 +6,21 @@ tags:
   - Semgrep Code
 ---
 
-# How to run all available rules on a repository.
+# Run all available rules on a repository
 
-If you would like to run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) and are not willing to add all available rules/rulesets to the Rule Board or add a variety of `--config` entries, what you can do is run the following command:
+To run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) without using the Rule board, enter the following command:
 
 ```
 semgrep --config=r/all
 ```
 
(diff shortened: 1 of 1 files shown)


docs/kb/rules/run-all-available-rules.md
```diff
@@ -6,15 +6,21 @@ tags:
   - Semgrep Code
 ---
 
-# How to run all available rules on a repository.
+# Running all available rules on a repository
```
s-santillan, line 9: We typically remove the "how to" because it's superfluous. (Also, espectially in the kb, everything is a how-to.)

docs/kb/rules/run-all-available-rules.md
```diff
@@ -6,15 +6,21 @@ tags:
...
 
-If you would like to run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) and are not willing to add all available rules/rulesets to the Rule Board or add a variety of `--config` entries, what you can do is run the following command:
+To run all rules that are available within the [Semgrep Registry](https://semgrep.dev/explore) without using the Rule board, enter the following command:
```
s-santillan, line 11: I just removed as many words as I could while keeping the goal clear.

docs/kb/rules/run-all-available-rules.md
```diff
@@ -6,15 +6,21 @@ tags:
...
 
 ```
 semgrep --config=r/all
 ```
 
-This command will pull out all available public, org private, and pro tiered rules if you have access to them. Below is an example output:
+This command runs the following rules from the Semgrep Registry:
```
s-santillan, line 17: I separated the items into a list so that users can skim quickly.

docs/kb/rules/run-all-available-rules.md
```diff
@@ -34,17 +40,20 @@ This command will pull out all available public, org private, and pro tiered rul
                                                                                                                         
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:02:31   
 ```
-The above example output is when you are running `--config=r/all` while logged in via `semgrep ci`. If you are not logged in you will just use the public rules within `r/all`.
 
-However there is one error that you should be wary about which is the following:
+If you are not logged in, `config=r/all` runs public rules only.
+
+## Troubleshooting invalid configuration file
```
s-santillan, line 46: Added a header to signpost in case they run into problems.

docs/kb/rules/run-all-available-rules.md
```diff
@@ -34,17 +40,20 @@ This command will pull out all available public, org private, and pro tiered rul
...
 
 ```
 [ERROR] invalid configuration file found (1 configs were invalid)
 ```
 
-This error may come up when running `r/all` if there is a current error within any of your custom/private rules.
+This error appears if there is a syntax or other error within any of your custom rules.
 
-To get around this you can use the following two commands as a workaround:
+Enter the following two commands as a workaround:
```
s-santillan, line 56: IT might be worth it to explain to users what the commands do so that they don't think they're arbitrarily deleting all their custom rules.

s-santillan: Snag_19e6d52

See above for explanation for what the commands do.

PR #1017, opened 2023-05-24
remove mention of old toggle for parity rules

s-santillan:

### Please ensure


andy-r2c (changes requested):
Can we add something that says results from these rules go in the 'Undetermined' filter, and you can not look at them by just untoggling it? 
Screen Shot 2023-05-25 at 12 11 18 PM

s-santillan:
@andy-r2c I apologize for taking so long to get back at this, I felt like I had to investigate the product.

This PR removed the section entirely, then I went through the triage & remediation docs and compared them to what is visible in the product currently. When you remove any mention of parity rules in the docs, and there's no mention of them in the product, it's like there's no distinction for the users to make between the two - to the user, there are simply "rules" running underneath the hood. So it didn't make any sense for me to mention "parity rules" anywhere and just explain that undetermined means we haven't performed reachability analysis on it.

Does that make sense? LMK!

andy-r2c (approved):

PR #1035, opened 2023-06-03
Reference material for Aliengrep

mjambon:
The main goal of this document is to let adventurous users experiment with these new features and collect their feedback so we can make adjustments.

### Please ensure

### DIFF
--- a/docs/writing-rules/experiments/aliengrep.md
+++ b/docs/writing-rules/experiments/aliengrep.md
@@ -1,0 +1,271 @@
+---
+slug: aliengrep
+append_help_link: true
+description: "Aliengrep is a variant of the generic mode that is more
+configurable than spacegrep."
+---
+
+:::caution
+This is an experimental matching mode for Semgrep.
+Many of the features described here are subject to change.
+Your feedback is important and will help us make desirable adjustments.
+:::
+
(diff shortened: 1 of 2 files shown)


adamkvitek (approved):
@mjambon thank you for this reference document, before we merge it, can we remove or work on the three remaining TODOs you left in the document? I pushed some small edits that did not change the meaning of what you originally wrote. I also included Aliengrep to sidenav...

Just listing the small edits I did:
- I tried to use more active voice instead of passive voice when possible.
- I removed future tense for easier reading (standard of simple English in docs - use future tense only when referring to a specific time in the future).
- I formatted the paragraphs into one-liners. I know that in code soft-wrapping is not welcomed much, but it is beneficial for docs as it is much easier to edit paragraphs that are not split after 80 (or any number of) characters.
- I made small edits such as more consistent use of `target input` instead of just using `target` or `input` (but I left only `input` in a few occurrences where it can be too repetitive).

mjambon:
Thanks @adamkvitek.
* I removed one TODO that needed to be removed and converted the two other ones into "note" admonitions. They're calls for feedback and it's the best I could find (list of Docusaurus admonitions: https://docusaurus.io/docs/markdown-features/admonitions).
* I used vscode instead of emacs so as to not break the lines.

PR #1036, opened 2023-06-03
feat!: new rule syntax docs

brandonspark:
## What:
This PR adds in new rule syntax docs, which are basically a straight translation of the old docs, but with the new syntax.

## Why:
We are using this for the purpose of testing the new syntax.

## How:
Just manually replaced a lot of things, and the Playground links.

### Please ensure


milanwilliams:
@adamkvitek Can we reopen this PR? I just realized I didn't go through to approve it, but it looks good to me! Would be helpful so we can refer to it during user testing :)

PR #1037, opened 2023-06-04
callout for turbo mode

s-santillan:

### Please ensure

Minor update to mention turbo mode. We emphasize our gentle push for its use through a tip box (in green) and mention it is not on by default ([per this slack thread](https://semgrepinc.slack.com/archives/C04RL2BBL2V/p1684351574049819)) and is in beta. I did test that it runs on both simple and advanced mode.

Snag_9c76b0


s-santillan:
Have added T's suggestion in.

src/components/procedure/_enable-turbo-mode.md
```diff
@@ -0,0 +1,9 @@
+:::tip Try Turbo mode (beta)
+Semgrep can automatically scan for findings using the Semgrep OSS Engine, on any change in the Rule or Sample Code panes.
```
adamkvitek, line 2: I was just checking [the pane definition](https://developers.google.com/style/ui-elements#term-windows) also in [Stack Exchange](https://ux.stackexchange.com/questions/124588/whats-the-difference-between-a-panel-and-pane). Looks correct - as pane is a section of a window. So the Editor is a page, with a window that takes roughly four-fifths of its size, with a left sample code pane and a right test code pane. Feel free to resolve this comment, and if you want send me more details so I can be sure we shouldn't use a "panel". :D

adamkvitek (approved):
@s-santillan LGTM. 🎉

s-santillan:
Thank you all. Will wait for Turbo mode to be available in Playground before merging.

PR #1075, opened 2023-06-14
Add info about findings triaged across refs

adamkvitek:

### Please ensure

### DIFF
--- a/src/components/reference/_triage-states.mdx
+++ b/src/components/reference/_triage-states.mdx
@@ -4,12 +4,12 @@
 | **Ignored** | Findings that are ignored are present in the code, but have been labeled as unimportant. Ignore findings that are false positives or deprioritized issues. Mark findings as ignored through Semgrep Cloud Platform (see [Managing finding status](/semgrep-code/findings/#managing-finding-status) or by adding a code comment (see [nosemgrep comments](/ignoring-files-folders-code/#reference-summary). |
 | **Fixed** | <p>Fixed findings were detected in a previous scan, but are no longer detected in the most recent scan of that same branch. </p> <p> To change the state of a finding to **fixed**: <ol type="a"><li>Fix the code so the rule does not match it.</li><li>Edit the Semgrep rule so it no longer matches the code.</li></ol></p> |
 
+:::info Findings triaged across refs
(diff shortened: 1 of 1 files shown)


src/components/reference/_triage-states.mdx
```diff
@@ -4,6 +4,10 @@
...
 | **Fixed** | <p>Fixed findings were detected in a previous scan, but are no longer detected in the most recent scan of that same branch. </p> <p> To change the state of a finding to **fixed**: <ol type="a"><li>Fix the code so the rule does not match it.</li><li>Edit the Semgrep rule so it no longer matches the code.</li></ol></p> |
 
+:::info Findings triaged across refs
+Findings triaged (ignored, reopened, resolved) in a specific branch, PR, or MR are also triaged in all other branches, PRs, and MRs of a specific repository. In addition, if you filter for any other [Git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References) (refs) on the [Findings](https://semgrep.dev/orgs/-/findings?tab=open) page, and then triage a finding, it is also automatically triaged in all other branches, PRs, MRs, and refs.
```
armchairlinguist, line 8: @chmccreery does this same thing apply for resolved? I would think if the finding cropped up again after being resolved that it would reappear, but maybe I'm confused.

chmccreery: Removed isn't a triage behavior, it just indicates that a finding no longer exists when we scan (e.g. if the rule that found that finding is no longer being used). I don't think we need to specify that here, though.

adamkvitek: @chmccreery what about resolved findings? I mean - if a code is detected again, it should be detected again - but if a finding is fixed (I should use "fixed" instead of "resolved" to be consistent) - it is only the fixed finding that is triaged.

Therefore, I have to remove "resolved" and just say that this applies to ignored and reopened findings, is this true?

chmccreery: Oh yes, that is correct

s-santillan:
Feel free to merge when SME review is done, this is xs-sized PR so no need for a TW stamp. 👍

src/components/reference/_triage-states.mdx
chmccreery, line 8:
```suggestion
Findings triaged (ignored, reopened) in a specific branch, PR, or MR are also triaged in all other branches, PRs, and MRs of a specific repository. In addition, if you filter for any other [Git references](https://git-scm.com/book/en/v2/Git-Internals-Git-References) (refs) on the [Findings](https://semgrep.dev/orgs/-/findings?tab=open) page, and then triage a finding, it is also automatically triaged in all other branches, PRs, MRs, and refs.
```

chmccreery (approved):

PR #1076, opened 2023-06-14
add admonition to reflect changes

s-santillan:

### Please ensure

~~- [ ] A technical writer reviews the content or PR~~ xs size pr

Preview:
Snag_1533518

### DIFF
--- a/docs/semgrep-cloud-platform/user-management.md
+++ b/docs/semgrep-cloud-platform/user-management.md
@@ -92,7 +92,7 @@ You have successfully joined an org in Semgrep Cloud Platform.
 Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on **Team or Enterprise tiers**.
 
 :::caution Changes to RBAC default roles
-Previously, new users in organizations with RBAC enabled were assigned a `member` role. Starting **June 14, 2023**, the default role is set through the **Defaults** configuration and defaults to `admin`.
+Previously, new users in organizations with RBAC enabled were assigned a `member` role. Starting **June 14, 2023**, the default role is set through the **Defaults** configuration.
 
 Change your default role assignment for new users by following the steps in [Setting a default role](#setting-a-default-role).
 :::
@@ -105,7 +105,7 @@ Semgrep Cloud Platform divides users into two roles:
 * `member`
 
 :::info
-* Users are assigned a role based on your [organization's default role](#setting-a-default-role).
+* Users are assigned a role based on your [organization's default role](#setting-a-default-role). New organizations are created with a default role of `admin`.
 :::
 
 The following table displays features available to each role:
--- a/docs/semgrep-cloud-platform/user-management.md
+++ b/docs/semgrep-cloud-platform/user-management.md
@@ -89,7 +89,13 @@ You have successfully joined an org in Semgrep Cloud Platform.
 
 ## Controlling access through roles 
 
-Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on **Team/Enterprise tiers**.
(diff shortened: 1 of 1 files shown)


docs/semgrep-cloud-platform/user-management.md
```diff
@@ -89,7 +89,13 @@ You have successfully joined an org in Semgrep Cloud Platform.
 
 ## Controlling access through roles 
 
-Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on **Team/Enterprise tiers**.
+Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on **Team or Enterprise tiers**.
+
+:::caution Changes to RBAC default roles
+Previously, new users in organizations with RBAC enabled were assigned a `member` role. Starting **June 14, 2023**, the default role is set through the **Defaults** configuration and defaults to `admin`.
```
armchairlinguist, line 95: Something like "defaults to `admin` for any organization that did not previously make use of RBAC" is probably where we'll land here

leifdreizler: I somewhat question if we need to make a change to the docs. 

- Customers that never used RBAC aren't impacted
- New customers aren't impacted
- Previous RBAC customers with 0 members in their deployment probably don't care
- Customers that did have members have had their default role set to member for their current deployment
- The few customers that had their behavior changed & invited someone during the 1 day window have been individually notified

I think the above message would really just apply to customers that are creating a second+ deployment that are used to the existing RBAC behavior on their previous deployment.

We already mention in the docs that the default behavior is `admin`, although that is lower in the doc.

Option for a simpler change:
[On line 108](https://github.com/returntocorp/semgrep-docs/pull/1076/files#diff-2c21bbf6c1c2f90aa4e383bdf649ed1239454bc0896df6abb7cca9c1ca8a17e7R108), add "New organizations are created with a default role of `admin`".

leifdreizler (approved):

docs/semgrep-cloud-platform/user-management.md
```diff
@@ -89,7 +89,13 @@ You have successfully joined an org in Semgrep Cloud Platform.
 
 ## Controlling access through roles 
 
-Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on **Team/Enterprise tiers**.
+Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role. This **role-based access control (RBAC)** feature is available for organizations on **Team or Enterprise tiers**.
```
leifdreizler, line 92:
```suggestion
Access control in Semgrep Cloud Platform determines the resources and features that are available to users based on their role.
```

leifdreizler: This is available to all customers 🙂

docs/semgrep-cloud-platform/user-management.md
s-santillan: @leifdreizler suggestion: how about keeping this admonition for a month? 

For @armchairlinguist we still default to admin for any org regardless of case (new, old, previously didn't use RBAC) so I think the original phrasing is fine

armchairlinguist: Hmm, I don't know if we have different definitions of "default". We've set a default of "Member" for folks previously making use of RBAC. The "default default" is still admin.

I think Leif's change might be a good choice.

s-santillan: I'll add Leif's suggestion, remove the admonition a month, after which we can mention the change in behavior in the release notes. [Reason being that we once deprecated an old Jira integration](https://semgrepinc.slack.com/archives/C0370TA46PL/p1678301197065599), and people were looking for when that happened.

LMK what yall think.

leifdreizler: sgtm!

s-santillan:
Thanks! Merging.

PR #1077, opened 2023-06-15
Update admonition to reflect Policies as default

armchairlinguist:
On June 14 we shifted from beta-testing the new Policies (Rule management) to making it the default. This update reflects that in the tips at the top of each relevant page.

[Linear issue](https://linear.app/semgrep/issue/MKT-680/update-rule-board-and-policies-docs-to-reflect-new-policies-as-default)

### Please ensure

### DIFF
--- a/docs/semgrep-code/policies.md
+++ b/docs/semgrep-code/policies.md
@@ -23,9 +23,9 @@ Object.entries(frontMatter).filter(
 # Policies
 
 :::tip
-The Policies page is a new version of the [Rule board](https://semgrep.dev/orgs/-/board).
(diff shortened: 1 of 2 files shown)


pangngan (approved):
LGTM! 🥳 Thanks for updating the docs 😄

s-santillan (approved):
This is a small sized PR, so I'm approving it in Adam's absence.

For Alexis: As you've seen with the RBAC PR, I insisted on an admonition to reflect the change so I'm glad we have something like this for the policies page. Usually though after a feature has had some time to settle it will stop being new so we may have to update the docs, and move the notice of the change to release notes, same as we will do for the RBAC. Thanks for contributing as always!

PR #1084, opened 2023-06-21
[New KB Article] Understand rule severities

armchairlinguist:
Provide context on how rule severities are assigned.

### Please ensure

### DIFF
--- a/docs/kb/rules/understand-severities.md
+++ b/docs/kb/rules/understand-severities.md
@@ -17,4 +17,4 @@ As a best practice, severity for Semgrep Registry rules in the `security` catego
 
 ## Semgrep Supply Chain 
 
-Semgrep Supply Chain rules have one of four severity levels: Critical, High, Medium or Low. The score assigned to the CVE using the [Common Vulnerability Scoring System (CVSS) score])(https://nvd.nist.gov/vuln-metrics/cvss), or the severity value set by the GitHub Advisory Database, determines the severity in Semgrep Supply Chain. If a vulnerability is given a 9.0 or higher it is assigned CRITICAL.
+Semgrep Supply Chain rules have one of four severity levels: Critical, High, Medium or Low. The score assigned to the CVE using the [Common Vulnerability Scoring System (CVSS) score])(https://nvd.nist.gov/vuln-metrics/cvss), or the severity value set by the GitHub Advisory Database, determines the severity in Semgrep Supply Chain. If a vulnerability is given a 9.0 or higher it is assigned Critical.
--- a/docs/kb/rules/understand-severities.md
+++ b/docs/kb/rules/understand-severities.md
@@ -11,10 +11,10 @@ tags:
 
 Semgrep Code rules have one of three severity levels: `ERROR` (High), `WARNING` (Medium), or `INFO` (Low). The severity indicates how critical the issues are that a rule potentially detects.
 
-The severity of a given rule is assigned by the rule author. For custom and third-party rules, the rule author's assignment is the source of truth.
+The rule author assigns the rule severity. For custom and third-party rules, their severity assignment is the source of truth.
 
 As a best practice, severity for Semgrep Registry rules in the `security` category should be assigned by evaluating the combination of [likelihood](/docs/contributing/contributing-to-semgrep-rules-repository/#likelihood) and [impact](/docs/contributing/contributing-to-semgrep-rules-repository/#impact). 
 
 ## Semgrep Supply Chain 
 
-Semgrep Supply Chain rules have one of four severity levels: Critical, High, Medium or Low. The score assigned to the CVE using the [Common Vulnerability Scoring System (CVSS) score])(https://nvd.nist.gov/vuln-metrics/cvss), or the severity value set by the GitHub Advisory Database, determines the severity in Semgrep Supply Chain. For example, if a vulnerability is given a 9.0 or higher it is assigned Critical.
+Semgrep Supply Chain rules have one of four severity levels: Critical, High, Medium or Low. The score assigned to the CVE using the [Common Vulnerability Scoring System (CVSS) score])(https://nvd.nist.gov/vuln-metrics/cvss), or the severity value set by the GitHub Advisory Database, determines the severity in Semgrep Supply Chain. For example, if a vulnerability is given a CVSS score of 9.0 or higher it is assigned Critical.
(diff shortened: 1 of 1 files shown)


docs/kb/rules/understand-severities.md
```diff
@@ -0,0 +1,20 @@
...
+
+Semgrep Code rules have one of three severity levels: ERROR (High), WARNING (Medium), or INFO (Low). The severity indicates how critical the issues are that a rule potentially detects.
+
+The severity of a given rule is assigned by the rule author. For custom and third-party rules, the rule author's assignment is the source of truth.
+
+As a best practice, severity for Semgrep Registry rules in the `security` category should be assigned by evaluating the combination of [likelihood](/docs/contributing/contributing-to-semgrep-rules-repository/#likelihood) and [impact](/docs/contributing/contributing-to-semgrep-rules-repository/#impact). 
+
+## Semgrep Supply Chain 
+
+Semgrep Supply Chain rules have one of four severity levels: Critical, High, Medium or Low. The score assigned to the CVE using the Common Vulnerability Scoring System (CVSS) score, or the severity value set by the GitHub Advisory Database, determines the severity in Semgrep Supply Chain. If a vulnerability is given a 9.0 or higher it is assigned CRITICAL.
```
sebasrevuelta, line 20: Good article. I would add a link to the CVSS scoring: https://nvd.nist.gov/vuln-metrics/cvss

armchairlinguist: Good call! I'll do that.

sebasrevuelta (approved):

docs/kb/rules/understand-severities.md
```diff
@@ -0,0 +1,20 @@
...
+
+Semgrep Code rules have one of three severity levels: ERROR (High), WARNING (Medium), or INFO (Low). The severity indicates how critical the issues are that a rule potentially detects.
+
+The severity of a given rule is assigned by the rule author. For custom and third-party rules, the rule author's assignment is the source of truth.
+
+As a best practice, severity for Semgrep Registry rules in the `security` category should be assigned by evaluating the combination of [likelihood](/docs/contributing/contributing-to-semgrep-rules-repository/#likelihood) and [impact](/docs/contributing/contributing-to-semgrep-rules-repository/#impact). 
+
+## Semgrep Supply Chain 
+
+Semgrep Supply Chain rules have one of four severity levels: Critical, High, Medium or Low. The score assigned to the CVE using the [Common Vulnerability Scoring System (CVSS) score])(https://nvd.nist.gov/vuln-metrics/cvss), or the severity value set by the GitHub Advisory Database, determines the severity in Semgrep Supply Chain. If a vulnerability is given a 9.0 or higher it is assigned CRITICAL.
```
sinat101, line 20: I feel like we could either go without defining what Critical's score range is, or if we do, do it for all the other severities as well. Also I think "CRITICAL" should just be "Critical", like the first sentence.

docs/kb/rules/understand-severities.md
```diff
@@ -0,0 +1,20 @@
...
+  - Semgrep Registry
+---
+
+# How does Semgrep assign severity levels to rules?
+
+## Semgrep Code
+
+Semgrep Code rules have one of three severity levels: ERROR (High), WARNING (Medium), or INFO (Low). The severity indicates how critical the issues are that a rule potentially detects.
+
+The severity of a given rule is assigned by the rule author. For custom and third-party rules, the rule author's assignment is the source of truth.
```
sinat101, line 14:
```suggestion
The severity of a given rule is assigned by the rule author. For custom and third-party rules, the rule author is the source of truth for the severity.
```

sinat101: I was initially a little confused by the wording, does this sound any better?

docs/kb/rules/understand-severities.md
```diff
@@ -0,0 +1,20 @@
...
+
+As a best practice, severity for Semgrep Registry rules in the `security` category should be assigned by evaluating the combination of [likelihood](/docs/contributing/contributing-to-semgrep-rules-repository/#likelihood) and [impact](/docs/contributing/contributing-to-semgrep-rules-repository/#impact).
```
sinat101, line 16: I feel like this instills some doubt in our severity assigning process. Why do we say "as a best practice" and "should be". Would it be accurate to just say:

sinat101, line 16:
```suggestion
As a best practice, severity for Semgrep Registry rules in the `security` category is assigned by evaluating the combination of [likelihood](/docs/contributing/contributing-to-semgrep-rules-repository/#likelihood) and [impact](/docs/contributing/contributing-to-semgrep-rules-repository/#impact). 
```

sinat101 (changes requested):

docs/kb/rules/understand-severities.md
```diff
@@ -0,0 +1,20 @@
...
+tags:
+  - Rules
+  - Semgrep Registry
+---
+
+# How does Semgrep assign severity levels to rules?
+
+## Semgrep Code
+
+Semgrep Code rules have one of three severity levels: ERROR (High), WARNING (Medium), or INFO (Low). The severity indicates how critical the issues are that a rule potentially detects.
```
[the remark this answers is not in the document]
armchairlinguist: I don't object to this change, but since they're already in all caps I'm not sure it's necessary?

docs/kb/rules/understand-severities.md
armchairlinguist: Currently we're not super consistent about it, so "is" might be a little strong. One day, I hope. :)

docs/kb/rules/understand-severities.md
armchairlinguist: yeah you're right, the severity writing should be consistent.

I'll see if I can find out more about the other severities before we merge this :) I only got the 9+ score info from what we had internally.

docs/kb/rules/understand-severities.md
armchairlinguist: Huh, I thought it made sense - the severity is assigned by the author, and that assignment is the source of truth. Though I probably shouldn't be writing in passive voice anyway!

What about:

> The rule author assigns the rule severity. For custom and third-party rules, their severity assignment is the source of truth.

docs/kb/rules/understand-severities.md
armchairlinguist: I was able to verify that when providing that example we were also referencing the same scoring rubric shown in the linked article, that just happened to be the level called out in the comment. I changed the wording to indicate that it's an example.

sinat101 (approved):

PR #1085, opened 2023-06-21
Update "Installing and running..." section: provide correct Docker commands

armchairlinguist:
The steps under "To install and run Semgrep OSS Engine, use one of the following options" previously had a single step that was in a tab set - the installation step for the platform - followed by two steps outside the tab set, to verify and run that were the same for each platform.

However, this doesn't work for Docker - if you're running Semgrep locally via Docker, the invocation is different. With this revision, all steps are in a tab set - which leads to some duplication - but the steps are also all correct. 😄 

### Please ensure

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -22,65 +22,102 @@ Install and run the [Semgrep command-line interface](https://github.com/returnto
 Semgrep CLI requires Python 3.7 or later.
 :::
 
-1. To install Semgrep OSS Engine, use one of the following options:
-    <Tabs
-        defaultValue="macOS"
-        values={[
-        {label: 'macOS', value: 'macOS'},
-        {label: 'Linux', value: 'Linux'},
-        {label: 'Windows Subsystem for Linux (WSL)', value: 'Windows Subsystem for Linux (WSL)'},
-        {label: 'Docker', value: 'Docker'},
-        ]}
-    >
+To install and run Semgrep OSS Engine, use one of the following options:
 
-    <TabItem value='macOS'>
+<Tabs
+    defaultValue="macOS"
+    values={[
+    {label: 'macOS', value: 'macOS'},
+    {label: 'Linux', value: 'Linux'},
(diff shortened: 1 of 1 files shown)


armchairlinguist:
If we'd also like to factor out that duplication, I can do that, but I might need some guidance :)

adamkvitek (approved):
@armchairlinguist thank you!

docs/getting-started.md
```diff
@@ -22,65 +22,103 @@ Install and run the [Semgrep command-line interface](https://github.com/returnto
...
+   
+  2. Confirm version by the following command:
+      ```sh
+      docker run --rm returntocorp/semgrep semgrep --version
+      ```
+
+  3. Run recommended Semgrep Registry rules:
     ```bash
     docker run --rm -v "${PWD}:/src" returntocorp/semgrep semgrep --config=auto
     ```
```
adamkvitek, line 114:
```suggestion
    <pre class="language-bash"><code>docker run --rm -v "${PWD}:<span className="placeholder">/SRC</span>" returntocorp/semgrep semgrep --config=auto</code></pre>

```

adamkvitek: @armchairlinguist This will not show a proper code highlighting, but it will display the SRC as a placeholder (italics, changed color also). This approach is inspired by https://developers.google.com/style/placeholders - but they also have some JavaScript for users to write the code immediately in the placeholder, and then copy it with their changed placeholder.

docs/getting-started.md
```diff
@@ -22,65 +22,103 @@ Install and run the [Semgrep command-line interface](https://github.com/returnto
...
 
-    </TabItem>
+    The provided `-v` option mounts the current directory into the container to be scanned. Change directories locally or provide a specific local directory in the command to scan a different directory.
```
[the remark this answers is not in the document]
armchairlinguist: Thanks for the thoughts! I'll leave this for now, since the meaning I'm trying to convey is "the option as provided in the command".

PR #1096, opened 2023-06-27
[New KB Article] Why are there more Semgrep findings when the code hasn't changed?

armchairlinguist:
Answer the common question of how findings can increase even if code hasn't changed, due to ruleset updates.

### Please ensure


Jason-Lim00 (approved):
LGTM on a content standpoint. Would maybe refer to the tech-writers if there is any format changes that are necessary.

docs/kb/semgrep-code/unexpected-new-findings.md
```diff
@@ -0,0 +1,26 @@
...
+  - Rules
+  - Semgrep Code
+description: This may be occurring because rule coverage has increased.
+---
+
+# Why are there more Semgrep findings when the code hasn't changed?
+
+If the rules you're using in Semgrep have changed since you last scanned your code, you may see more findings even if your code has not changed.
+
+For Semgrep-curated rulesets ("by Semgrep"), if you add a ruleset to one of your policies, the policy receives updates and additions to the ruleset on an ongoing bases. So if Semgrep adds a rule to a ruleset, or makes a change to make a rule more comprehensive (or more precise), your policy automatically picks up those changes, and the subsequent scan shows new findings for the new or updated rules.
```
sinat101, line 12: Since we are moving away from policies, should we update this to better reflect current rule management? Also, non-semgrep rulesets can get updated as well.

sinat101, line 12:
```suggestion
Rulesets can receive updates and additions to their rules on an ongoing basis. So if Semgrep (or a third-party rule author) adds or makes changes to a rule, Semgrep automatically picks up those additions or changes and the subsequent scan can show new findings for the new or updated rules.
```

docs/kb/semgrep-code/unexpected-new-findings.md
```diff
@@ -0,0 +1,26 @@
...
+
+For these rulesets, you can view each rule's history to see recent changes:
```
sinat101, line 14:
```suggestion
For Semgrep authored rulesets, you can view each rule's history to see recent changes:
```

sinat101 (changes requested):

docs/kb/semgrep-code/unexpected-new-findings.md
armchairlinguist: Actually we're moving back toward policies, that's the official name of the new rule management :)

Let's chat sync about the question of ruleset source, this turned out to be complicated and I'm definitely not sure I got it all correct.

sinat101 (approved):

PR #1100, opened 2023-06-28
update scan dedupe behavior

s-santillan:

### Please ensure

~~- [ ] A technical writer reviews the content or PR~~ small-sized PR


chmccreery (commented):
This isn't really accurate until/unless we change our API. How do you feel about specifying that deduplication happens in the UI but not in the API or waiting until we fix the API to mimic the UI behavior?

s-santillan:
@chmccreery sorry for the late reply - what's the state of the API now? If it's not fixed, I'll just add the caveat so we can merge this ASAP.

chmccreery:
The API now let's you specify "dedup" behavior - by default it does not dedup across branches, but by specifying "dedup = True" you can get the same behavior. I think it's fine to merge this or add a caveat about that.

chmccreery (approved):

s-santillan:
@chmccreery Thanks. I separated the two and linked to the dedup flag in the Semgrep API docs. Merging :)

PR #1103, opened 2023-06-30
Merging KB into main and publishing the Knowledge Base!

sinat101:
Hello, please take a look at the PR. I've looked over everything myself and there are no conflicts. My local build works as expected. If you agree, let's get this merged and publish the KB :)


armchairlinguist (approved):
I have some nitpicks, but they can wait til after launch! Let's do this thing! 🚀

pabloest:
Thanks to the CSE team for this PR! It’s super exciting to see all this content about to drop!

Some suggestions and a proposal:

#### Generated index cards
I added a min-height to the cards so that they align nicely. See before and after below. 

##### Before
CleanShot 2023-07-01 at 10 51 56@2x

##### After
CleanShot 2023-07-01 at 10 59 25@2x


#### Nav bar

I suggest placing the KB link underneath the docs and making a simple drop-down (similar to how the products drop-down works here currently). Alternatively, we can rename that docs dropdown → Resources (or something similar) and include both docs and KB as sub-links underneath. Thoughts?

pabloest:
@s-santillan In the search, will we be ready to distinguish KB results from docs results once this merges?

sinat101:
Oh yea the varying card heights was bothering me, thanks for that!

Regarding the placement of the KB link on the top nav bar, I kind of think it should have its own top level menu item, because I'm hoping to get users to notice it as a resource they can use in addition to the docs. Just my personal thoughts though.

sinat101:
@pabloest i spoke to @s-santillan beforehand and she mentioned that once merged, Algolia would begin indexing the KB and show articles in search results.

Any chance we can get this PR approved today?

s-santillan:
@pabloest I'll have to see if it indexes correctly but this is something I can fix post-merge!

s-santillan (approved):

PR #1104, opened 2023-07-01
Add version check to Prerequisites due to recent change in comment auth

armchairlinguist:
Strictly speaking the Cloud Platform support is already beyond the called-out version, but there was a specific change that will make comments not work with lower versions, which (I think) is worth calling out for now.

### Please ensure


pangngan (approved):
LGTM!

armchairlinguist:
I'm actually going to close this without merging - based on our data, there are few or no scans from versions earlier than 1.13.0 currently, so this callout is essentially redundant.

PR #1124, opened 2023-07-11
Add private publication steps for Semgrep Editor

armchairlinguist:
Clarifies the steps necessary to publish a rule as private to the Registry from the Editor.

### Please ensure


s-santillan (approved):
Not me ugly crying in the distance because of all your updates!! Thank you so much.

PR #1133, opened 2023-07-18
add caveat about go.mod

mmcqd:
See change. This has always been true, but it should have been documented.

### Please ensure


mmcqd:
Actually, we totally do support this. lol

PR #1134, opened 2023-07-18
Add definition of Undetermined to SSC overview

armchairlinguist:
We get a fair number of questions about this term, and previously it was only defined in the filter descriptions. This PR adds it to the overview page along with the other possibilities.

### Please ensure

### DIFF
--- a/docs/semgrep-supply-chain/overview.md
+++ b/docs/semgrep-supply-chain/overview.md
@@ -40,7 +40,7 @@ Semgrep Supply Chain generates a **finding** when it detects a match. If the dep
 
 A finding is **unreachable** if the dependency contains a known vulnerability, but the vulnerable matching code is not used in your codebase.
 
-A finding is **undetermined** if Semgrep detects a match for the dependency in the vulnerable version range, but does not scan for the reachability of this vulnerability. This is most common with older or lower severity vulnerabilities.
+A finding is **undetermined** if Semgrep detects a match for the dependency in the vulnerable version range, but does not have a reachability rule for this vulnerability. This is most common with older or lower severity vulnerabilities.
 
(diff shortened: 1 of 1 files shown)


s-santillan (approved):
Thank you!

docs/semgrep-supply-chain/overview.md
```diff
@@ -40,6 +40,8 @@ Semgrep Supply Chain generates a **finding** when it detects a match. If the dep
 
 A finding is **unreachable** if the dependency contains a known vulnerability, but the vulnerable matching code is not used in your codebase.
 
+A finding is **undetermined** if Semgrep detects a match for the dependency in the vulnerable version range, but does not scan for the reachability of this vulnerability. This is most common with older or lower severity vulnerabilities.
```
andy-r2c, line 43: I think we can also just say older and lower severity vulnerabilities don't have reachability rules? I feel like some users might interpret "does not scan for the reachability of this vulnerability" as a parsing/code issue

armchairlinguist: Would you prefer:

A finding is **undetermined** if Semgrep detects a match for the dependency in the vulnerable version range, but does not have a reachability rule for this vulnerability.

or something else? Open to suggestions!

armchairlinguist: @andy-r2c got a sec to take a look at this alternative?

andy-r2c: "A finding is undetermined if Semgrep detects a match for the dependency in the vulnerable version range, but does not have a reachability rule for this vulnerability."

works perfectly!

andy-r2c (approved):

PR #1140, opened 2023-07-20
[New KB Article] Matching multiple tokens with ellipsis metavariables

armchairlinguist:
* Add a guide for using ellipsis metavariables to match multiple word tokens in generic mode (a relatively common question).
  * As part of this addition, I checked whether "ellipsis metavariable" is the desired standard, and brought the new/experimental Aliengrep mode document into that fold.
  * While I was editing the Aliengrep doc I also did some other light Vale style linting on it. I didn't attempt to change the first-person usages because they were extensive, but I fixed uses of will and other smaller suggestions.

### Please ensure

### DIFF
--- a/docs/kb/rules/ellipsis-metavariables.md
+++ b/docs/kb/rules/ellipsis-metavariables.md
@@ -31,7 +31,7 @@ If you remove the ellipsis in the `$...ID` variable, the second example no longe
 
 ## Alternative: try the Aliengrep experiment
 
-Semgrep's generic matching mode is [somewhat limited](/docs/writing-rules/generic-pattern-matching/#caveats-and-limitations-of-generic-mode). To address some of the limitations, the team is experimenting with a new mode called [Aliengrep](/docs/writing-rules/experiments/aliengrep/). 
+Semgrep's generic matching mode contains [caveats and limitations](/docs/writing-rules/generic-pattern-matching/#caveats-and-limitations-of-generic-mode). To address some of the limitations, the team is experimenting with a new mode called [Aliengrep](/docs/writing-rules/experiments/aliengrep/). 
 
 With Aliengrep, you can [configure what characters are allowed as part of a word token](/docs/writing-rules/experiments/aliengrep/#additional-word-characters-captured-by-metavariables) as well as [have even more fun with ellipses](/docs/writing-rules/experiments/aliengrep/#ellipsis-).
 
--- a/docs/kb/rules/ellipsis-metavariables.md
+++ b/docs/kb/rules/ellipsis-metavariables.md
@@ -15,11 +15,11 @@ Using ellipsis (`...`) to match a sequence of items is one of the most common co
 
 Most commonly, ellipsis metavariables like `$...ARGS` are used for purposes like matching multiple arguments to a function, [while capturing the values for later re-use](/docs/writing-rules/pattern-syntax/#ellipsis-metavariables).
 
-However, they can also be used to match multiple word tokens. This is especially handy in [Generic pattern matching mode](/docs/writing-rules/generic-pattern-matching/), where a token is defined as a sequence of characters in the set `[A-z0-9_]`. 
+However, they can also be used to match multiple word tokens. This is especially handy in [Generic pattern matching mode](/docs/writing-rules/generic-pattern-matching/), where a token is defined as a sequence of characters in the set `[A-z0-9_]`, which is not a fit for all languages.
 
 ## Capturing multiple tokens with ellipsis metavariables
 
-In `generic` mode, `ABC_DEF` is one token, and a metavariable such as `$VAR` captures the entire sequence. However, `ABC-DEF` is two tokens, and a metavariable such as `$VAR` does not capture the entire sequence. 
+In `generic` mode, `ABC_DEF` is one token, and a metavariable such as `$VAR` captures the entire sequence. However, `ABC-DEF` is two tokens, and a metavariable such as `$VAR` does not capture the entire sequence.
 
 If the language you're working with allows other characters in tokens, using ellipsis metavariables can prevent problems with matching too little of the pattern.
 
@@ -33,7 +33,7 @@ If you remove the ellipsis in the `$...ID` variable, the second example no longe
 
 Semgrep's generic matching mode is [somewhat limited](/docs/writing-rules/generic-pattern-matching/#caveats-and-limitations-of-generic-mode). To address some of the limitations, the team is experimenting with a new mode called [Aliengrep](/docs/writing-rules/experiments/aliengrep/). 
 
-With Aliengrep, you can [configure what characters are allowed as part of a word token](/docs/writing-rules/experiments/aliengrep/#additional-word-characters-captured-by-metavariables) as well as have [even more fun with ellipses](/docs/writing-rules/experiments/aliengrep/#ellipsis-). 
+With Aliengrep, you can [configure what characters are allowed as part of a word token](/docs/writing-rules/experiments/aliengrep/#additional-word-characters-captured-by-metavariables) as well as [have even more fun with ellipses](/docs/writing-rules/experiments/aliengrep/#ellipsis-).
 
 Give it a try and share your thoughts!
 
--- a/docs/kb/rules/ellipsis-metavariables.md
+++ b/docs/kb/rules/ellipsis-metavariables.md
@@ -1,0 +1,36 @@
+---
+description: Ellipsis metavariables can help with matching multiple word tokens.
+tags:
+  - Rules
+  - Semgrep Code
+---
+
+import MoreHelp from "/src/components/MoreHelp"
+
+# Using ellipsis metavariables to match multiple tokens
+
+Ellipses (`...`) to match a sequence of items are one of the most-used constructs in Semgrep rules, and they can even be used with metavariables ($VAR) to increase matching scope!
+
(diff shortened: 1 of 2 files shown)


emjin (approved):
This looks good to me from a quick glance but Martin is the aliengrep expert

docs/kb/rules/ellipsis-metavariables.md
```diff
@@ -0,0 +1,40 @@
...
+description: Ellipsis metavariables can help with matching multiple word tokens.
+tags:
+  - Rules
+  - Semgrep Code
+---
+
+import MoreHelp from "/src/components/MoreHelp"
+
+# Matching multiple tokens with ellipsis metavariables
+
```
s-santillan, line 11:
```suggestion
Match multiple tokens, such as arguments of a function or elements of an array, through the use of two Semgrep rule constructs:

* Ellipsis metavariables
* Aliengrep
```

Just adding an intro and roughly defining a "token". Your subsequent definition of a token is not replaced.

docs/kb/rules/ellipsis-metavariables.md
```diff
@@ -0,0 +1,40 @@
...
+## What are ellipsis metavariables?
```
s-santillan, line 12:
```suggestion
## What are ellipsis metavariables?
```

I'd suggest avoiding hypophora in technical writing, whether body text or headings. We use two types of headings:
* Task-based
* Conceptual

Alternatives
* Introduction to ellipsis metavariables
* Defining ellipsis metavariables

docs/kb/rules/ellipsis-metavariables.md
```diff
@@ -0,0 +1,40 @@
...
+
+Using ellipsis (`...`) to match a sequence of items is one of the most common constructs in Semgrep rules, and ellipses can even be used with metavariables ($VAR) to increase matching scope!
```
s-santillan, line 14: Are items the same as tokens?

docs/kb/rules/ellipsis-metavariables.md
```diff
@@ -0,0 +1,40 @@
...
+
+Most commonly, ellipsis metavariables like `$...ARGS` are used for purposes like matching multiple arguments to a function, [while capturing the values for later re-use](/docs/writing-rules/pattern-syntax/#ellipsis-metavariables).
+
+However, they can also be used to match multiple word tokens. This is especially handy in [Generic pattern matching mode](/docs/writing-rules/generic-pattern-matching/), where a token is defined as a sequence of characters in the set `[A-z0-9_]`, which is not a fit for all languages.
```
s-santillan, line 18:
```suggestion
However, they can also be used to match multiple word tokens. This is especially handy in [Generic pattern matching mode](/docs/writing-rules/generic-pattern-matching/), where a token is defined as a sequence of characters in the set `[A-z0-9_]`, which is not a fit for all languages.
```

"Not a fit for all languages" - what are you trying to say here? I feel like there's a comparison being made between the `generic` language and other programming languages.

docs/kb/rules/ellipsis-metavariables.md
```diff
@@ -0,0 +1,40 @@
...
+
+To match `ABC-DEF` in `generic` mode, use an ellipsis metavariable, like `$...VAR`. Here is an example rule:
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=J6Ro" title="pattern-not rule for unverified transactions" width="100%" height="432px" frameBorder="0"></iframe>
+
+If you remove the ellipsis in the `$...ID` variable, the second example no longer matches.
+
+## Alternative: try the Aliengrep experiment
+
+Semgrep's generic matching mode is [somewhat limited](/docs/writing-rules/generic-pattern-matching/#caveats-and-limitations-of-generic-mode). To address some of the limitations, the team is experimenting with a new mode called [Aliengrep](/docs/writing-rules/experiments/aliengrep/).
```
s-santillan, line 34:
```suggestion
Semgrep's generic matching mode contains [caveats and limitations](/docs/writing-rules/generic-pattern-matching/#caveats-and-limitations-of-generic-mode). To address some of the limitations, the team is experimenting with a new mode called [Aliengrep](/docs/writing-rules/experiments/aliengrep/). 
```
Just a suggestion to try to match the title of of the doc or section the user will go to when they click.

docs/writing-rules/experiments/aliengrep.md
```diff
@@ -70,11 +70,11 @@ In Semgrep rule syntax, an ellipsis is a specific pattern written as three dots
 
 Ellipses at the beginning or at the end of a pattern are anchored. For example, ellipses must match the beginning or the end of the target input, respectively. For example, `...` alone matches the whole input and `a ...` matches the whole input starting from the first occurrence of the word `a`.
 
-### Metavariable ellipsis (capturing ellipsis)
+### Ellipsis metavariable (capturing ellipsis)
 
-A metavariable ellipsis `$...X` matches the same contents as an ordinary ellipsis `...` but additionally captures the contents and assigns it to the metavariable `X`.
+An ellipsis metavariable `$...X` matches the same contents as an ordinary ellipsis `...` but additionally captures the contents and assigns them to the metavariable `X`.
```
s-santillan, line 75: Nice catch :)

s-santillan (changes requested):
Hey Alexis, apologies but I had to ask for some clarifications on this one! The meaning is getting there but needs a little more specification. 🤓

docs/kb/rules/ellipsis-metavariables.md
armchairlinguist: This isn't something I want to have upfront, in particular because this KB is about a less typical usage of ellipsis metavariables that isn't either of those things. But you have very quickly identified something that definitely isn't clear, and most of your comments basically center on this: what is a token?

Most of our docs only talk about tokens in the context of authentication (which is a special use/type of token), so we don't have a lot more on this, except in a few places. The [caveats section](https://semgrep.dev/docs/writing-rules/generic-pattern-matching/#caveats-and-limitations-of-generic-mode) that I link to is somewhat helpful:

> metavariable support is limited to capturing a single “word”, which is a token of the form [A-Za-z0-9_]+

Tokenization in language processing in general is a phase of analysis where you separate the language into its more basic elements. The details of the process, and what constitutes a token, depend on the language being analyzed and the purpose of the analysis. In the case of the generic language processor for Semgrep, it tokenizes `hello, world` as 3 tokens (as in the example), but only two of those tokens are "words"; the comma is a separator/punctuation. `ABC-DEF` is also broken down into 3 (two "words", ABC and DEF, and one separator, `-`), but not ABC_DEF - that's one token, because underscore is considered part of a word in that processor.

@emjin might have a good way to break this down more accessibly so we can present a brief definition that makes sense...

docs/kb/rules/ellipsis-metavariables.md
armchairlinguist: I think probably if we get tokens defined correctly above, yes. :)

docs/kb/rules/ellipsis-metavariables.md
```diff
@@ -0,0 +1,40 @@
...
+
+Using ellipsis (`...`) to match a sequence of items is one of the most common constructs in Semgrep rules, and ellipses can even be used with metavariables ($VAR) to increase matching scope!
+
+Most commonly, ellipsis metavariables like `$...ARGS` are used for purposes like matching multiple arguments to a function, [while capturing the values for later re-use](/docs/writing-rules/pattern-syntax/#ellipsis-metavariables).
+
+However, they can also be used to match multiple word tokens. This is especially handy in [Generic pattern matching mode](/docs/writing-rules/generic-pattern-matching/), where a token is defined as a sequence of characters in the set `[A-z0-9_]`, which is not a fit for all languages.
+
+## Capturing multiple tokens with ellipsis metavariables
+
+In `generic` mode, `ABC_DEF` is one token, and a metavariable such as `$VAR` captures the entire sequence. However, `ABC-DEF` is two tokens, and a metavariable such as `$VAR` does not capture the entire sequence.
```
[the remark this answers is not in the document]
armchairlinguist: Yeah, this is a version of the standard regex notation for the character class including all letters and digits, plus underscore. It's sometimes written [A-Za-z0-9] for clarity but that just makes it clearer that letters in both lower and uppercase are included.

It's actually not my definition, it comes from the generic mode caveat that I link to and which I quoted in the comment thread above. It doesn't include dash unless you add a dash after the underscore - a dash between two range endpoints in a character class can only be a range specifier, not a character on its own.

docs/kb/rules/ellipsis-metavariables.md
armchairlinguist: Thanks, I adjusted it!

docs/kb/rules/ellipsis-metavariables.md
s-santillan: > This isn't something I want to have upfront, in particular because this KB is about a less typical usage of ellipsis metavariables that isn't either of those things.

Yeah, I guess the array / function example doesn't match the rest of the article. What would make a better example or use case in this case for generic "words"?

s-santillan (approved):
Approving as there's only 1 last question left (about tokens) that can be addressed w/o the need for a review. Thank you!

docs/kb/rules/ellipsis-metavariables.md
armchairlinguist: I ended up reorganizing the content at the top and adding more clear definitions/examples of terms to try to resolve this. Hopefully it make sense. :)

armchairlinguist:
@s-santillan Thanks for the approval - I'll merge this now even though I did make some changes to content organization as part of clarifying definition/usage of token and other terms. Very happy for later feedback or direct changes if you have any other concerns!

PR #1142, opened 2023-07-24
Update installation commands to Google style (describe result/effect of command)

armchairlinguist:
This came up when I did a copy-pasta of the Getting Started doc into a KB and @s-santillan pointed out that the sentence sounded a bit strange without a verb for running the command. It turns out Google prefers if you just describe the effect or result of the command so simplifying to that.

### Please ensure


s-santillan (approved):
thank you!

PR #1150, opened 2023-07-31
Add callout to clarify that transitive can be unreachable if not reachable by inclusion

armchairlinguist:
Add a callout box mentioning that it's possible for transitive dependencies to show as unreachable if they have a reachability rule, but aren't used directly in the first-party codebase (which would be the case for the vast majority of transitive dependencies).

### Please ensure


s-santillan:
(Asked additional questions for now in the slack!)

armchairlinguist:
@s-santillan ooh you're too fast. I talked to Matthew about this further and this is not right, we had crossed wires. I'll close this and submit a new PR later on.

s-santillan (approved):
lgtm!

PR #1156, opened 2023-08-10
updated based on pa team

s-santillan:

### Please ensure

~~- [ ] A technical writer reviews the content or PR~~ XS size PR, not needed

I set Rust as experimental for intrafile because we have not extensively tested it.
Removed beta from terraform as discussed.

### DIFF
--- a/src/components/reference/_supported-languages-table.mdx
+++ b/src/components/reference/_supported-languages-table.mdx
@@ -14,8 +14,8 @@
 | PHP        | GA | Beta | -- |                   
 | Python     | GA | Beta | -- |                        
 | Scala      | GA | Beta | -- |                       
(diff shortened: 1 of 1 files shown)


IagoAbal (approved):

src/components/reference/_supported-languages-table.mdx
```diff
@@ -14,8 +14,8 @@
 | PHP        | GA | Beta | -- |
```
IagoAbal, line 14: I don't think `JSON` has any special "cross-function" support.

s-santillan: Makes sense. Removed beta in JSON. Thank you for the catch.

PR #1165, opened 2023-08-16
improve docs user story for members to join orgs

s-santillan:

### Please ensure

- [x] **no todos or tks remain for this PR**

* Initially, Raj brought up the need to have docs specifically for members to be "invited" (or as close to "invited" as we can make them feel).
* While we don't have an explicit invite feature yet, we do have live improvements to a member's first sign in that haven't been documented.
* When I reviewed the docs, they are out of sync with the product in the sense that the current docs focus solely on the first-person-in-an-org to use Semgrep and didn't accommodate "team members". **With this update, orgs and team members are better accommodated**.
* I opted to rewrite the doc focusing on **guiding** both user paths, with **steps** rather than conceptual info that can be read elsewhere.


sachishah:
@s-santillan , for Joining an Org, should we link out to the SSO instructions?

For scanning, we're making several improvements to zero-config scanning (goal is to get it to a Public Beta state by the end of the week) as well as the recommended path for new users — should we hold off on the changes here until the product updates are live over the next two weeks?

docs/pricing-and-billing.mdx
```diff
@@ -28,7 +28,7 @@ Semgrep's various functionalities are available through several offerings:
 <dt>Semgrep OSS</dt>
 <dd>The OSS (open source software) offering includes Semgrep OSS Engine, a fast static analysis command line tool for finding bugs and and enforcing code standards. It is licensed under LGPL 2.1. </dd>
 <dt>Semgrep Team tier</dt>
-<dd>The Team tier includes Semgrep Code and Semgrep Supply Chain (SSC), which enables users to scan both first-party and third-party open source dependencies. It also includes Semgrep Cloud Platform, a web app enabling users to manage users, organizations, repositories, and scans.</dd>
+<dd>The Team tier includes Semgrep Code and Semgrep Supply Chain (SSC), which enables users to scan both first-party and third-party open source dependencies. It also includes Semgrep Cloud Platform, a web app enabling users to manage users, organizations, repositories, and scans. Team tier also includes <strong>cross-function (intrafile)</strong> and <strong>cross-file (interfile) analysis</strong> through <a href="/semgrep-code/semgrep-pro-engine-intro/">Semgrep Pro Engine</a>.</dd>
```
armchairlinguist, line 31: Can we turn the inclusions here into a bullet list? I _think_ that's allowed within a definition list, and it would be better now that there are three:

The Team tier includes:
* Semgrep Code and Semgrep Supply Chain (SSC), which enable users to scan both first-party and third-party open source dependencies. 
* <strong>Cross-function (intrafile)</strong> and <strong>cross-file (interfile) analysis</strong> through <a href="/semgrep-code/semgrep-pro-engine-intro/">Semgrep Pro Engine</a>.
* Semgrep Cloud Platform, a web app enabling users to manage users, organizations, repositories, and scans.

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,82 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+5. Your organization is created and you are prompted to start a scan. Instead of starting a scan, click **<i class="fa-solid fa-gear"></i> Settings** in the navigation menu.
+6. Click **Source code managers**.
+7. Select the source code manager that contains the org to connect to Semgrep Cloud Platform.
+![DESCRIPTION](/img/source-code-manager.png#md-width)
+8. Follow the steps to connect your source code manager. You can review permissions needed by Semgrep in [Requested permissions for GitHub and GitLab](#requested-permissions-for-github-and-gitlab).
+
+## Joining an organization
+
+:::info Prerequisites
+- Your admin or inviter should have an existing organization (org) in Semgrep that is connected to the org in either GitHub or GitLab.
```
armchairlinguist, line 103:
```suggestion
- Your admin or inviter should have an existing organization (org) in Semgrep that is connected to an org in either GitHub or GitLab.

```

armchairlinguist: "Inviter" is a bit weird here (funny since we use "invitee" a lot, but I don't hear "inviter" much). I might go with "Your admin or the person inviting you to SCP".

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,82 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+- You must be a member of the GitHub or GitLab org that is connected to Semgrep.
+:::
+
+For **members to join an organization**, perform the following steps to be added to the org:
```
armchairlinguist, line 107:
```suggestion
For **members**, perform the following steps to join the org:
```
I'm open to other alternative phrasing here, but the current "to join an organization...to be added to the org" feels heavy and redundant.

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,82 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+
+1. Sign in to Semgrep Cloud Platform. A **Welcome to Semgrep** dialog box appears.
+2. Check that the organization under the **Join an organization** button is the correct org to join. If there is more than 1 org, select the org to join.
```
armchairlinguist, line 110:
```suggestion
2. Check that the organization under the **Join an organization** button is the correct org to join. If there is more than one org, select the org to join.

```

armchairlinguist: I don't think this sentence meets any of the exception criteria in the Google style guide for small numbers, but let me know if I'm mistaken!

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,82 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+
+:::tip Product-specific information
+* To learn more about SAST scans on your codebase, see [Getting started with Semgrep Code](/semgrep-code/getting-started).
+* To learn more about SCA scans for your third-party dependencies, see [Getting started with Semgrep Supply Chain](/semgrep-supply-chain/getting-started).
+* Both products are **free for up to 10 contributors**. See [Usage limits](/usage-limits) to learn more about contributors and usage limits.
+:::
+
+### Starting a local repository scan and sending findings to SCP
+
+You can send scan results from a local repository to Semgrep Cloud Platform. The local repository is a separate **Project** from its remote counterpart so the findings records do not overwrite each other and are kept separate.
```
armchairlinguist, line 132: I think this is usually true, but is it always true?

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,82 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+
+To send findings from a local repository, perform the following steps in your command-line interface:
+
+1. Login to Semgrep:
```
armchairlinguist, line 136:
```suggestion
1. Log in to Semgrep:

```

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,82 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+```
+2. Click the login URL provided. Your are taken to your web browser to complete the login process.
+3. Follow any additional steps.
+4. After logging in, start a scan in your CLI:
+```
+semgrep ci
+```
+
+:::info
+Many improvements to the Semgrep Cloud Platform experience only work with up-to-date Semgrep CLI versions. For this reason, Semgrep Cloud Platform only supports the 10 most recent minor versions of Semgrep CLI. For example, if the latest release was 0.114.0, all versions greater than 0.104.0 are supported while earlier versions, such as 0.103.0 can be deprecated or can result in failures.
```
armchairlinguist, line 148:
```suggestion
Many improvements to the Semgrep Cloud Platform experience only work with up-to-date Semgrep CLI versions. For this reason, Semgrep Cloud Platform only supports the 10 most recent minor versions of Semgrep CLI. For example, if the latest release was 0.114.0, all versions greater than 0.104.0 are supported, while earlier versions, such as 0.103.0, can be deprecated or can result in failures.

```

src/components/procedure/_platform-detect-ghrepos.md
```diff
@@ -0,0 +1,8 @@
+To ensure that your GitHub repository is **detected** by Semgrep Cloud Platform:
+
+1. Log into GitHub.
+2. Click your **profile photo > Settings > Applications**.
```
armchairlinguist, line 4: I don't think this set of steps will work for organizations as you need to select your org first. What would be the best way for us to handle that in this snippet?

docs/semgrep-cloud-platform/getting-started.md
armchairlinguist: https://developers.google.com/style/word-list#login

I'll drop a comment on our internal example that uses `login` too :)

armchairlinguist:
@sachishah I wouldn't recommend that we try to lump this in with changes for cloud scanning. It will delay us from making this update, which is an important one, and make it harder to review the changes.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: Nice catch. This is one of those things I forget about when editing my work! Thank you.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: There is one case I haven't verified, which is a case when someone pulls an org repo (say semgrep) then uploads findings from semgrep to the org - what happens when another user does that? I'll ask the cloud folks.

armchairlinguist: I think it might overwrite in that case, potentially. My observation is it's usually `org_name/repo_name` from CI and just `repo_name` locally. 

I was also thinking about if someone goes and reads about environment variables and sets the name, but that's pretty unlikely when they're just getting started.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: I thought it fell under "**Technical quantities**, such as amounts of memory, amounts of disk space, numbers of queries, or usage limits" which is e

armchairlinguist: Ah I see. I don't feel strongly about it, so feel free to ignore.

src/components/procedure/_platform-detect-ghrepos.md
s-santillan: Nice catch, I created separate steps for personal vs org users :)

docs/pricing-and-billing.mdx
s-santillan: Yep, resolved in latest :)

docs/semgrep-cloud-platform/getting-started.md
s-santillan: Fixed in latest, I'll add a screenshot in the main comment body.

s-santillan:
@sachishah I have added some links to SSO configuration to ensure they don't fall off the user path.



@armchairlinguist I added a section explaining how project slugs behave between local and remote scans

Snag_c15dd4

Thank you both for your notes, I feel that we have polished this doc quite a bit! Please approve if all good.

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+6. Click **Source code managers**.
+7. Select the source code manager that contains the org to connect to Semgrep Cloud Platform.
+![DESCRIPTION](/img/source-code-manager.png#md-width)
+8. Follow the steps to connect your source code manager. You can review permissions needed by Semgrep in [Requested permissions for GitHub and GitLab](#requested-permissions-for-github-and-gitlab).
+9. Optional: Refer to the instructions in [SSO configuration](/semgrep-cloud-platform/sso/) to set up SSO for your organization.
+
+## Joining an organization
+
+:::info Prerequisites
+- Your admin or person inviting you to Semgrep Cloud Platform should have an existing organization (org) in SCP that is connected to an org in either GitHub or GitLab.
```
armchairlinguist, line 104: We may want to avoid use of the word "invite" here because folks are not actually invited. Maybe just "Your admin" should be enough?

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+- You must be a member of the GitHub or GitLab org that is connected to Semgrep.
+:::
+
+For **members**, perform the following steps to join the org:
+
+1. Sign in to Semgrep Cloud Platform. If your administrator has configured SSO, click **Use SSO**. A **Welcome to Semgrep** dialog box appears.
+2. Check that the organization under the **Join an organization** button is the correct org to join. If there is more than 1 org, select the org to join.
+![DESCRIPTION](/img/join-an-org.png#md-width)
```
armchairlinguist, line 112: Should DESCRIPTION still be here?

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+:::
+
+To send findings from a local repository, perform the following steps:
+
+1. Ensure that you are signed into Semgrep Cloud Platform in the account you want to send findings to. It is recommended to send local repository findings to your **personal** account.
+2. Log in to Semgrep:
+```
+semgrep login
+```
+2. Click the login URL provided. Your are taken to your web browser to complete the login process.
```
armchairlinguist, line 146: A lot of CLIs don't allow a direct click on this link - but not sure if that distinction really matters?

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+3. Follow any additional steps.
+4. After logging in, start a scan in your CLI:
+```
+semgrep ci
+```
+
+#### Projects separation between local and remote repositories
```
armchairlinguist, line 153:
```suggestion
#### Project separation between local and remote repositories
```
Projects sounds a bit odd to me. If we do change this, the link above would need to be changed too.

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+
+The Project slug for a remote repository always takes the form `account-name/repository-name`. The Project slug for a local repository always takes the form `repository-name`. Refer to the following screenshot for an example of both remote and local Projects in a single personal account.
```
armchairlinguist, line 155:
```suggestion
The Project slug for a remote repository takes the form `account-name/repository-name`. The Project slug for a local repository takes the form `repository-name`. Refer to the following screenshot for an example of both remote and local Projects in a single personal account.
```
we could add "by default", but mostly I just don't think it makes sense to say "always" even if it's just in a getting started page, since it's a configurable value.

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+
+* **For personal accounts:** A local repository scan **never** overwrites findings records of its remote counterpart. They are two separate Projects. Personal accounts only have one team member or user: you.
```
armchairlinguist, line 157:
```suggestion
* **For personal accounts:** A local repository scan does not overwrite the findings records of its remote counterpart. They are two separate Projects. Personal accounts only have one team member or user: you.
```

docs/semgrep-cloud-platform/getting-started.md
```diff
@@ -94,6 +76,95 @@ Using Semgrep Supply Chain with Semgrep Cloud Platform provides users with the V
...
+* For **organization accounts**: A local repository scan never overwrites findings records of its remote counterpart, but if two members both send local repository findings, their records can overwrite each other's findings. This is because org accounts can have more than one team member, but all local scans are sent to the same Project slug.
```
armchairlinguist, line 158:
```suggestion
* For **organization accounts**: A local repository scan does not overwrite the findings records of its remote counterpart, but if two members both send local repository findings, their records can overwrite each other's findings. This is because org accounts can have more than one team member, but all local scans are sent to the same Project slug.
```
"their records can overwrite each other's findings" sounds weird here but I can't think of a great replacement. If you can, though, I'm for using it :)

armchairlinguist (approved):
This is a huge improvement for sure, thank you for all the hard work, @s-santillan ! I still had a few comments, mainly about the use of always/never about project names, but I leave it to your discretion.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: You're right, it's awkward. 

Alternative: if two members both send local repository findings, one set of results may overwrite the other unintentionally.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: Click or copy. In the latest client, I was amazed that my WSL actually automatically opened my windows program. I am truly living in the future, 2 OS in 1. Cygwin could never.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: This is so tricky. I used the term "invite" because that's the language people used to roughly and casually describe their goal, and is the gap we're trying to address because we don't have the functionality in the product per se.

Alternatively, I could remove the above usage of "invite", then explicitly state that Semgrep does not give out email invites, and then instruct admins to tell their team members to sign in after the admin has made an org.

What do you think?

docs/semgrep-cloud-platform/getting-started.md
s-santillan: Nice catch. Fixed in latest.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: Makes sense. I think removing the word "always" or "never" solves it :) Thank you.

docs/semgrep-cloud-platform/getting-started.md
s-santillan: Update: Yeah, I went in this direction :)

PR #1166, opened 2023-08-16
CircleCI snippet fix: SAST snippet was running SSC instead

armchairlinguist:
SAST standalone snippets should not use the `--supply-chain` argument :)
There are some more complex issues with this snippet but I've raised those in a separate internal issue for discussion - this fix at least gets us running the right product.

### Please ensure


s-santillan (approved):
Nice catch!

PR #1179, opened 2023-08-31
Move VS Code + Semgrep on Windows to integrations category

armchairlinguist:
We were looking for this article today and had trouble finding it in SCP. Even though it requires login, it's more of an integration guide than part of SCP.

### Please ensure

### DIFF
--- a/docs/kb/integrations/semgrep-vs-code-windows.md
+++ b/docs/kb/integrations/semgrep-vs-code-windows.md
@@ -6,9 +6,10 @@ tags:
   - WSL
 ---
 
-# Visual Studio Code and Semgrep on Windows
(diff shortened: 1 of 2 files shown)


sinat101 (approved):

armchairlinguist:
@s-santillan I realized that out of habit I deleted the thing about page URLs, but this would change the page URL - do we need to add a redirect?

s-santillan:
One of the ways to check is to yarn build. I'll take care of it :)

s-santillan:
Btw, terms and taxonomy are quite complex and ever-changing things. I'll leave a message in Slack for reasons of product development :)

s-santillan (approved):

PR #1195, opened 2023-09-26
Update SAST standalone snippets with --code

armchairlinguist:
At the moment this change won't have any effect in the user-facing docs, since we removed these snippets from the CI configuration reference earlier, but this prepares us for adding them to the KB in the future.

The change is to remove reference to SEMGREP_RULES which is not compatible with `semgrep ci`, and instead execute with the `--code` argument, which was added in [1.40.0](https://github.com/returntocorp/semgrep/releases/tag/v1.40.0).

### Please ensure


s-santillan (approved):
Why do I get this feeling something is going to change with this PR very soon :D

armchairlinguist:
@s-santillan Do it fast! And then do it again!

PR #1196, opened 2023-09-26
Update a KB and doc for Semgrep Supply Chain branch awareness

armchairlinguist:
Semgrep Supply Chain is now branch aware, so users can see findings from other branches than the default/common ones through a branch selector. An article and doc needed updates to reflect that the requirement for branch identification is now less strict (recommended rather than required).

### Please ensure


s-santillan (approved):
Thank you for the reword as well!

PR #1204, opened 2023-10-05
Branch clarifications and wording improvements for CircleCI instructions

armchairlinguist:
The primary goal here is to clarify that the use of `main` in the new snippet represents main as the default branch, and users should substitute their own default branch if needed.

I also made some wording adjustments and linted the file (which mostly meant replacing BitBucket with Bitbucket in this case).

### Please ensure

- [ ] ~A subject matter expert (SME) reviews the content~ xs change


s-santillan (approved):
lgtm!

armchairlinguist:
Seems like GitHub is in a bad state here with the CLA check, so I'll close and re-open to see if that helps.

PR #1224, opened 2023-10-17
Policies: Update of Cloud Platform URL from /board & other remaining bits

armchairlinguist:
We'd mostly moved over to "Policies" from "Rule board" but a few things were left, including the SCP URL (now live with no redirect) and some other references to the rule board in obscure places.

### Please ensure


s-santillan (approved):
LGTM!

PR #1225, opened 2023-10-17
Add note on SSC directory structure assumption, reword info note

mmcqd:
This PR adds a description of the assumptions that semgrep supply chain makes about the relationship between lockfiles and code files. I added it to the main docs page because if these assumptions are violated by a customer's project, which has happened, then SSC will not be able to produce correct results. I think my description may be overly wordy, a picture/diagram might be better. Let me know what you think.

I re-worded the info note because I found it confusing. Semgrep doesn't treat individual folders in monorepos as repositories, it's just that supply chain findings aren't just grouped by repository, but by lockfile too. And manifest files are never used for grouping findings, only lockfiles. If a folder had a manifest file but no lockfile, it would not do anything. If you think my change is not clarifying, feel free to change it back.

### Please ensure


docs/semgrep-supply-chain/getting-started.md
```diff
@@ -38,8 +38,11 @@ Semgrep Supply Chain is **automatically enabled** for all repositories that you
 
 When the scan finishes, Semgrep Cloud Platform displays an overview of findings in the **Supply Chain** page for further triage and remediation. See [Triaging and remediating dependency findings](/semgrep-supply-chain/triage-and-remediation).
 
+### Warning on Project Directory Structure
```
armchairlinguist, line 41: I don't think this is exactly a warning. I'd probably describe it as something like "Repository structure requirements" (maybe not exactly that, Sara might have an idea).

docs/semgrep-supply-chain/getting-started.md
```diff
@@ -38,8 +38,11 @@ Semgrep Supply Chain is **automatically enabled** for all repositories that you
...
+Semgrep Supply Chain assumes your repository contains a [lockfile](/semgrep-supply-chain/glossary/#lockfile). It also assumes that all code files using the dependencies in this lockfile are nested in the same directory as the lockfile, potentially within deeper directories. For example, if your lockfile is in `my-project/running/lockfile.json`, Semgrep Supply Chain assumes that all code files using the dependencies in `my-project/running/lockfile.json` are nested in `my-project/running/` or deeper directories. If you have code files in `my-project/biking`, Semgrep Supply Chain will not associate them to the dependencies in `my-project/running/lockfile.json`. If there is another lockfile in `my-project/running`, like `my-project/running/uphill/lockfile.json`, then this will override the original `my-project/running/lockfile.json` for all code files in `my-project/running/uphill/` or deeper directories.
```
armchairlinguist, line 42: I think this is probably feeling wordy because "in the same directory, potentially within deeper directories" is a pretty long way to describe this.

The description I used in the KB article is:

> Semgrep Supply Chain searches the parent directories of any code files for the nearest relevant lockfile. Monolithic repositories (monorepos) have their findings grouped based on the lockfiles present in subdirectories.

I think wording similar to that would work here, along with your example. Your example might be better as or supplemented with a visual or ASCII representation of the directory tree and notations of which code goes to which lockfiles, though. As much as I like words, they aren't great for descriptions of non-linear things like tree structures. :)

s-santillan:
@mmcqd Are you ok with me taking over & rewriting to address the notes above? 🙏

khorne3 (approved):
LGTM!

PR #1244, opened 2023-10-31
We should not tell people to use --output

aryx:
The --output is a bit redundant with shell redirection
so we should not advise it. It was added a long time ago
and we're not even sure anymore if it's that useful.
People in theory can use it to post the result to an URL
but even that I'm not sure it's used. In any case
when a user just want to save the output to a file, he
can simply use shell redirection.

test plan:
wait for green CI checks

### Please ensure


aryx:
For reference: https://github.com/martinabrle/app-service-java-demo/issues/13#issuecomment-1785583735

aryx:
ping

s-santillan:
For future readers, note that release notes with `--output` (August 2023, December 2021, June 2022) and FAQ.md are intentionally kept untouched.

s-santillan (approved):

PR #1267, opened 2023-11-18
Update user management for clarity on connection

armchairlinguist:
Clarify some language and processes around connecting Semgrep orgs to SCM orgs.

### Please ensure


s-santillan (approved):
When someone connects with SSO, do they first set up the SSO connection, then the GH/GL org integration, or do they do the org creation first, usually?

armchairlinguist:
@s-santillan That's an important question in general, which is one reason I was thinking about retooling the instructions more significantly to take SSO into account. Currently it's super important to set up GH first - I even added a callout on the SSO page about it recently if you remember that change. (GL is more agnostic since the flow doesn't install an app and those with SSO don't necessarily need the GL member connection.)

Should we try to account for that here? I have some ideas about re-writing, but I'm also wondering if it's worth just merging this version and then doing that separately.

s-santillan:
@armchairlinguist Katie is rewriting the TOFU docs, so thanks for that bit of information. We don't need to include it in the scope of this PR, so feel free to merge.

@khorne3 See comment above por favor

PR #1273, opened 2023-11-23
improve the azure sample with tested configuration

sebasrevuelta:

### Please ensure

### DIFF
--- a/src/components/code_snippets/_azure-semgrep-app-sast.mdx
+++ b/src/components/code_snippets/_azure-semgrep-app-sast.mdx
@@ -1,20 +1,25 @@
 ```yaml
-# trigger:
-#  - master
-
-pool:
-  vmImage: ubuntu-latest
 variables:
-- group: Semgrep Cloud Platform token group
+- group: Semgrep_Variables
 
-steps: 
-
+steps:
+- checkout: self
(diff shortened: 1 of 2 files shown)


s-santillan (approved):

src/components/code_snippets/_azure-semgrep-app-sast.mdx
```diff
@@ -1,20 +1,25 @@
...
 variables:
-- group: Semgrep Cloud Platform token group
-
-steps: 
+- group: Semgrep_Variables
 
+steps:
+- checkout: self
+  clean: true
+  fetchDepth: 100000
```
armchairlinguist, line 8: @sebasrevuelta I'm a bit worried about setting fetchDepth that high by default. It could result in taking quite some time to fetch large repos. Did you test it in those conditions?

armchairlinguist: I'd say 20-100 is probably a more reasonable range (GitLab currently defaults to 20 and we only occasionally get reports of that being an issue).

sebastianrevuelta: I will change that to 20

PR #1301, opened 2023-12-18
[wip] revisions to top-level SCP docs

khorne3:
currently for reference only.

### DIFF
--- a/docs/semgrep-cloud-platform/overview.md
+++ b/docs/semgrep-cloud-platform/overview.md
@@ -1,0 +1,22 @@
+---
+slug: overview
+append_help_link: true
+title: Overview
+hide_title: false
+description: Learn how Semgrep Cloud Platform can help you scan your repositories, manage your Semgrep findings, and integrate Semgrep into your software development lifecycle.
+tags:
+  - Semgrep Cloud Platform
+  - Team & Enterprise Tier
+---
+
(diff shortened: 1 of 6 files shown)


docs/semgrep-cloud-platform/overview.md
```diff
@@ -0,0 +1,22 @@
...
+tags:
+  - Semgrep Cloud Platform
+  - Team & Enterprise Tier
+---
+
+Semgrep Cloud Platform (SCP) is used to manage and orchestra Semgrep Code, Semgrep Supply Chain, and Semgrep Secrets.
+
+With SCP, you can:
+
+* Add a repository to automate the continuous scanning of your code
```
s-santillan, line 16: More of a fact-checking note: "Adding a repository" is currently "setting up a CI job in your CI provider" because Semgrep cannot do this for you for any non GitHub Actions CI provider. Meaning to say that it only looks like you're "adding" repositories to Semgrep when you're using GHA, because everyone else has to commit a CI file (it would be more accurate to say that you're adding Semgrep to their CI workflow).

s-santillan: (I am aware that the "setting up a CI job" is an implementation detail we're trying to abstract from the user, and that "adding repos to Semgrep" is much more elegant, it's just that we aren't quite there yet where Semgrep just DOES IT for more than 1 CI provider)

s-santillan:
I've put as much of this info as I can into the new deployment docs - are we ok with closing this? @khorne3

khorne3:
Yup, all good!

PR #1303, opened 2023-12-18
Add information about how to provide feedback

armchairlinguist:
In many cases, we provide hints for how to give feedback to Assistant, but it's useful to have it documented also, especially for the cases where we don't give hints (primarily the Assistant-thinks-it's-a-true-positive case).

### Please ensure

### DIFF
--- a/docs/semgrep-code/semgrep-assistant-code.md
+++ b/docs/semgrep-code/semgrep-assistant-code.md
@@ -204,9 +204,9 @@ A low confidence level is recommended as even incorrect suggestions may be usefu
 
 ### Providing feedback to Semgrep Assistant
 
-If Semgrep Assistant suggests that a finding is a false positive, you will be prompted to leave feedback. Your feedback helps Semgrep refine the assistant feature and is very much appreciated.
+If Semgrep Assistant suggests that a finding is a false positive, you will be prompted to leave feedback. Your feedback helps Semgrep refine the Assistant feature and is very much appreciated.
 
-* In Slack notifications, **Agree** and **Disagree** buttons appear under the Assistant recommendation message, as shown in "Semgrep Assistant in a Slack notification".
+* In Slack notifications, **Agree** and **Disagree** buttons appear under the Assistant recommendation message, as shown in [Semgrep Assistant in a Slack notification](/img/semgrep-assistant-slack.png).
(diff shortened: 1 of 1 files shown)


docs/semgrep-code/semgrep-assistant-code.md
```diff
@@ -202,4 +202,15 @@ A low confidence level is recommended as even incorrect suggestions may be usefu
...
 *Figure*. Semgrep Assistant generating a potential fix.
 
+### Providing feedback to Semgrep Assistant
+
+If Semgrep Assistant suggests that a finding is a false positive, you will be prompted to leave feedback. 
+
+* In Slack notifications, **Agree** and **Disagree** buttons appear under the Assistant recommendation message, as shown in "Semgrep Assistant in a Slack notification".
+* In the Semgrep Cloud Platform UI, the Assistant recommendation appears under **Activity** for a finding, along with the buttons **Agree and ignore** or **Disagree**.
+
+You can also leave feedback on any Assistant comment in GitHub pull requests with `/semgrep assistant agree|disagree`.
```
Sly1029, line 212: Not every Assistant comment can take feedback. One example would be autofix comments do not take feedback

khorne3 (approved):
Thank you!

armchairlinguist:
@khorne3 Thanks for taking a look. When I checked after the changes, the punctuation on the list items looks inconsistent to me. I usually try to make sure that either all elements are complete sentences with periods, or none of them are. Do you have a thought on the best approach for these specific items, considering you added the screenshot in and that entails a colon?

Maybe you could also weigh in on the [thread](https://semgrepinc.slack.com/archives/C02JT6YA5S7/p1702941943565869) about how to handle repeating the screenshot, since you changed how I did it :)

khorne3:
@armchairlinguist The extra punctuation was my mistake -- I think we try to omit completely, per Google's guidelines. I've since removed.

Thanks for pointing out the screenshot thing -- I had just looked at the edited section at the bottom. I thought linking to the image was a bit confusing, so I think repeating the screenshot is a better option. That said, perhaps we consider could removing it, since the text isn't confusing necessarily without it?

armchairlinguist:
OK, I think we're in a happy place for now with the list! Please don't hesitate to modify in the future for further clarity.

PR #1316, opened 2024-01-03
add notes regarding GHA scan behavior

khorne3:
For https://linear.app/semgrep/issue/MKT-640/make-explicit-scan-paths-behavior-in-gha


s-santillan (approved):
Thank you!

PR #1320, opened 2024-01-09
corrected table based on feedback

s-santillan:

### Please ensure

- [ ] A technical writer reviews the content or PR **(nano-sized pr)**

Feedback:
> ...language support graph for SCA is confusing. At first glance it feels like we are saying Kotlin has reachability support because the :x: is only for License Detection and there is no :x: for Time Period of Reachability Rule Coverage.

Agreed, it erroneously says "Since May 2022" for all package managers, including those which don't have any reachability rules. Reachability rules for these package managers should be n/a.

> the :x: feels like a very loud "no" where a blank communicates the same, but leaves room for "soon" 
Also agreed, and considering that the language table uses `--`, I went with those for consistency.


s-santillan:
Snag_3ffcbbc

PR #1341, opened 2024-01-31
Explain permissions required for private repositories when using GHAS

pjzcodes:
The previous version of this doc indicated that it wasn't possible to run this job in private repositories. This update clarifies that it is possible if the permissions are set appropriately.

### Please ensure

### DIFF
--- a/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
+++ b/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
@@ -72,7 +72,9 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@v3
-      - run: semgrep scan --sarif > semgrep.sarif
+      - run: semgrep ci --sarif > semgrep.sarif
+        env:
+          SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
       - name: Upload SARIF file for GitHub Advanced Security Dashboard
         uses: github/codeql-action/upload-sarif@v2
         with:
--- a/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
+++ b/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
@@ -13,7 +13,7 @@ If you run the alternate job and it fails with a "resource not accessible by int
 
 ## The workflow or job does not have sufficient permissions
 
-Semgrep findings in a private repository can only populate in the GitHub Advanced Security Dashboard if the correct permissions are set using the `permissions` key. See an example below.
+Semgrep findings in a private repository can only populate in the GitHub Advanced Security Dashboard if the correct permissions are set using the `permissions` key. See the following example.
 
 ## The workflow permissions in your repository's Actions settings are set to read-only
 
--- a/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
+++ b/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
@@ -11,7 +11,7 @@ When scanning with Semgrep in CI, findings automatically populate in Semgrep Clo
 
 If you run the alternate job and it fails with a "resource not accessible by integration" error, there are two possible causes.
 
-## Your repository's default workflow permissions are set to read-only
+## Your repository's workflow permissions are set to read-only
 
 Repository-level workflow permissions are set to `read-only` (default) unless they've previously been changed. Use of the `permissions` key within the workflow file does not override this setting.
 
@@ -45,7 +45,6 @@ on:
     branches: ["master", "main"]
   schedule:
     - cron: '20 17 * * *' # Sets Semgrep to scan every day at 17:20 UTC.
-    # It is recommended to change the schedule to a random time.
 
 jobs:
   semgrep:
@@ -56,7 +55,6 @@ jobs:
       # A Docker image with Semgrep installed. Do not change this.
       image: returntocorp/semgrep
 
-    # Skip any PR created by dependabot to avoid permission issues:
     if: (github.actor != 'dependabot[bot]')
     permissions:
       # required for all workflows
--- a/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
+++ b/docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
@@ -23,7 +23,7 @@ To update this setting:
 Target permissions
 
 :::info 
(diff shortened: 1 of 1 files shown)


docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -11,9 +11,9 @@ When scanning with Semgrep in CI, findings automatically populate in Semgrep Clo
 
 If you run the alternate job and it fails with a "resource not accessible by integration" error, there are two possible causes.
 
-## Your repository is private
+## The workflow or job does not have sufficient permissions
 
-Findings from Semgrep can only populate in the GitHub Advanced Security Dashboard if the repository is public. Otherwise, the job does not have sufficient permissions to write to the dashboard regardless of any permissions set at the workflow or job level.
+Semgrep findings in a private repository can only populate in the GitHub Advanced Security Dashboard if the correct permissions are set using the `permissions` key. See an example below.
```
s-santillan, line 16:
```suggestion
Semgrep findings in a private repository can only populate in the GitHub Advanced Security Dashboard if the correct permissions are set using the `permissions` key. See the following example.
```

pjzcodes:
@armchairlinguist sure! I'll ping you.

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -35,35 +35,44 @@ Changing the repository's default workflow permissions changes the permissions f
...
+      # required for all workflows
+      security-events: write
+      # for workflows in private repos
+      actions: read
+      contents: read
     steps:
       - uses: actions/checkout@v3
-      - run: semgrep ci --sarif > semgrep.sarif
-        env:
-          SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
```
armchairlinguist, line 65: I don't see this elsewhere in the file, so I don't think this line should be deleted. Pretty important to have the token. :)

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -35,35 +35,44 @@ Changing the repository's default workflow permissions changes the permissions f
 This job only requires `write` permissions for `security-events`.
 
 ```yml
+# Name of this GitHub Actions workflow.
```
armchairlinguist, line 38: I'd suggest not adding these comments in this example - they're helpful in our default configs to explain what's occurring, but I don't think they add a lot for customizing the config, and they distract from the important elements that are different.

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -11,9 +11,9 @@ When scanning with Semgrep in CI, findings automatically populate in Semgrep Clo
 
 If you run the alternate job and it fails with a "resource not accessible by integration" error, there are two possible causes.
 
-## Your repository is private
```
armchairlinguist, line 14: With these changes, I think it would make sense to reverse the order of the possible issues, since one applies to all repos, and one only applies to private repos. Since this one has greater need the config file as a concrete example, that also would allow it to be adjacent to the config file without needing to repeat the file contents, while the other can perhaps have an anchor link or something in the info box.

This is the change that's a little harder to picture/share an exact outcome for since it requires a re-jig of the structure, but hopefully my description makes sense.

As part of that, I think it would make sense to change the title of this section to "The workflow or job does not have read permissions in a private repository" or something more specific like that - otherwise, both titles reference permissions so it's not entirely clear why they're separate.

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -11,15 +11,11 @@ When scanning with Semgrep in CI, findings automatically populate in Semgrep Clo
...
+## Your repository's default workflow permissions are set to read-only
```
armchairlinguist, line 14: A tad confused about this change - I think I'd stick with "repository's workflow permissions", mainly because they could be read-only by being the default (very likely) or for some other reason (were set that way intentionally after having been different), and I don't think that distinction is relevant here, unless I'm missing something (possible).

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -11,15 +11,11 @@ When scanning with Semgrep in CI, findings automatically populate in Semgrep Clo
...
 
-Findings from Semgrep can only populate in the GitHub Advanced Security Dashboard if the repository is public. Otherwise, the job does not have sufficient permissions to write to the dashboard regardless of any permissions set at the workflow or job level.
+Repository-level workflow permissions are set to `read-only` (default) unless they've previously been changed. Use of the `permissions` key within the workflow file does not override this setting.
```
armchairlinguist, line 16: I'm a little bit confused about this statement as well. Below you say that for more granular permissions you can set the permissions at the workflow or job level, but here you're saying you can't do that. Which one is true, or are these not conflicting but I'm misunderstanding?

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -44,9 +45,8 @@ on:
     branches: ["master", "main"]
   schedule:
     - cron: '20 17 * * *' # Sets Semgrep to scan every day at 17:20 UTC.
-    
-permissions: 
-    security-events: write
+    # It is recommended to change the schedule to a random time.
```
armchairlinguist, line 48: This comment really doesn't belong here - if you're going to retain it, it should be up with the cron line.

docs/kb/semgrep-ci/github-upload-findings-in-security-dashboard.md
```diff
@@ -11,30 +11,31 @@ When scanning with Semgrep in CI, findings automatically populate in Semgrep Clo
...
+Changing the repository's default workflow permissions changes the permissions for all workflows in that repository. Use of the `permissions` key will not override this setting, so updating it is a required step. Learn more about the `permissions` key at [Assigning permissions to jobs](https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#setting-the-github_token-permissions-for-all-jobs-in-a-workflow), or review the example workflow-level permissions below.
 :::
 
+## The workflow or job does not have the correct permissions in a private repository
+
+In order for Semgrep findings in a private repository to appear on the GitHub Advanced Security Dashboard, you must ensure that the appropriate permissions are configured at the workflow level using the permissions key. See the following example.
```
armchairlinguist, line 31:
```suggestion
In order for Semgrep findings in a private repository to appear on the GitHub Advanced Security Dashboard, you must ensure that the appropriate permissions are configured at the workflow level using the `permissions` key. See the following example.
```

armchairlinguist (approved):
Thanks for working through this! I left just one minor suggestion that I recommend for clarity, but otherwise looks good.

PR #1344, opened 2024-02-02
add suggested CLI-only path

khorne3:
@s-santillan pushing to share this content. I would be happy to move the file, reformat, etc., based on what works best.

### Please ensure

### DIFF
--- a/docs/getting-started/cli.md
+++ b/docs/getting-started/cli.md
@@ -1,0 +1,6 @@
+# CLI guide 
+
+Learn how to set up Semgrep, scan your first project for security issues, and view your findings in the CLI.
+
+## Prerequisites
+You must have Python 3.8 or later installed on the machine where the Semgrep CLI is running.
--- a/docs/getting-started/cli.md
+++ b/docs/getting-started/cli.md
@@ -1,6 +1,177 @@
-# CLI guide 
+---
+slug: cli
+title: CLI-only workflow 
+hide_title: true
+description: Learn how to set up Semgrep, scan your first project for security issues, and view your findings in the CLI.
+tags:
+  - quickstart
+  - cli
+---
+
+import MoreHelp from "/src/components/MoreHelp"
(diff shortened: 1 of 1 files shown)


docs/getting-started/cli.md
```diff
@@ -0,0 +1,177 @@
...
+
+### Scan without sending results to Semgrep
+
+To scan your project **without** sending data to Semgrep, use:
+
+```console
+semgrep ci --dry-run
+```
+
+## Set log levels
```
s-santillan, line 150: Very useful section, thank you.

s-santillan:
I know I'm not supposed to review this yet, but have some notes that may be useful to you as you build on this thing.

First thing - I suggest closing this branch, then opening a new branch that branches out of `sara/sidebar-v3.02/deployment` so that you can use the anchors from `docs/prerequisites/` (or componentize & import the Python 3.8 requirement if you prefer the user not have to click the link).

- We can also componentize the installation of Semgrep CLI, because it appears in other docs anyway; we can also use a link but if the intent is to make this a fairly standalone doc, I support that too - the choice is yours. If you choose the standalone way, The componentizing is useful so we can import to a bunch of different docs.
- One of the hardest things about the CLI workflow is dealing with the OSS vs Pro workflows. To help clarify (for us as well) the differences, maybe it's worth it to also include OSS workflows in the PR, then we can decide if we want OSS **and** Pro to be in 1 whole doc, or if we can separate them. We can include 1 file in 2 different sidebars, so this "CLI workflow doc" you started can be in both our Pro and OSS sidebars. 
    - Something that might help you write: https://semgrepinc.slack.com/archives/C8HRFBPSQ/p1699643585940479 - a discussion about the intentions behind the two commands.
    -  There's also this somewhat unmaintained doc: https://semgrep.dev/docs/running-rules/ that you can check out.

Let me know if these suggestions make sense :vulcan_salute:

PR #1345, opened 2024-02-05
add info on custom dependency exceptions

khorne3:
for https://github.com/semgrep/semgrep-app/pull/12250

### Please ensure

### DIFF
--- a/docs/semgrep-supply-chain/license-compliance.md
+++ b/docs/semgrep-supply-chain/license-compliance.md
@@ -135,4 +135,22 @@ Exempted packages appear in the Supply Chain > **Settings** tab.
 
 Package exemptions are currently version-specific. Each version used must be exempted individually.
 
+### Custom dependency exceptions
+
+Custom dependency exceptions allow you to prevent Semgrep from blocking a pull
+request or merge request due to licensing issues. For example, if
+`bitwarden/cli@2023.9.0`, which has a GPL-3.0 license, is on the allowlist,
+setting a custom dependency exception means that the exclusion won't fail when
+upgrading to `bitwarden/cli@2023.9.1`.
+
+To set a custom dependency exception:
+
+1. Log into Semgrep Cloud Platform and navigate to **Supply Chain** > <i class="fa-solid fa-gear"></i> **Settings**.
+2. Under **Custom Dependency Exceptions**, click **Add custom exception**.
+3. In the **Add custom dependency exception** window that appears:
+   1. Select the **Ecosystem** where this dependency applies
+   2. Provide the **Package name**, e.g., `bitwarden/cli`
+   3. Provide the **Version** information for the package. The major, minor, and patch version information is required; pre-release and build metadata are optional.
+4. Click **Add** to save and add the exception.
+
 <MoreHelp />


s-santillan:
@khorne3 just checking, is this tracked in linear?

docs/semgrep-supply-chain/license-compliance.md
```diff
@@ -135,4 +135,22 @@ Exempted packages appear in the Supply Chain > **Settings** tab.
 
 Package exemptions are currently version-specific. Each version used must be exempted individually.
 
+### Custom dependency exceptions
```
s-santillan, line 138: Just curious, how does "custom dependency exceptions" interact with "exempting packages"?

Also, do we want to establish a preferred term for dependency / package? When I read the two sections (pretending that I am jumping in from Google) my first thought is "why is one "package" and the other "dependency"?

docs/semgrep-supply-chain/license-compliance.md
```diff
@@ -135,4 +135,22 @@ Exempted packages appear in the Supply Chain > **Settings** tab.
...
+setting a custom dependency exception means that the exclusion won't fail when
+upgrading to `bitwarden/cli@2023.9.1`.
+
+To set a custom dependency exception:
+
+1. Log into Semgrep Cloud Platform and navigate to **Supply Chain** > <i class="fa-solid fa-gear"></i> **Settings**.
+2. Under **Custom Dependency Exceptions**, click **Add custom exception**.
+3. In the **Add custom dependency exception** window that appears:
+   1. Select the **Ecosystem** where this dependency applies
+   2. Provide the **Package name**, e.g., `bitwarden/cli`
```
s-santillan, line 152:
```suggestion
   2. Provide the **Package name**, for example, `bitwarden/cli`.
```
https://developers.google.com/style/word-list#eg

docs/semgrep-supply-chain/license-compliance.md
khorne3: I asked a clarifying question about this above; I think one is a subset (superset) of the other -- either way, I'll update with an explanation.

Good catch on package/dependency; I think I'll update all instances of "package" to "dependencies" since that's what's in the UI.

khorne3:
https://linear.app/semgrep/issue/MKT-804/add-info-on-custom-dependency-exceptions-1345

docs/semgrep-supply-chain/license-compliance.md
```diff
@@ -135,4 +135,22 @@ Exempted packages appear in the Supply Chain > **Settings** tab.
...
+Custom dependency exceptions allow you to prevent Semgrep from blocking a pull
+request or merge request due to licensing issues. For example, if
+`bitwarden/cli@2023.9.0`, which has a GPL-3.0 license, is on the allowlist,
+setting a custom dependency exception means that the exclusion won't fail when
+upgrading to `bitwarden/cli@2023.9.1`.
+
+To set a custom dependency exception:
+
+1. Log into Semgrep Cloud Platform and navigate to **Supply Chain** > <i class="fa-solid fa-gear"></i> **Settings**.
+2. Under **Custom Dependency Exceptions**, click **Add custom exception**.
```
s-santillan, line 149: The style guide recommends not to use location-specific terms: https://developers.google.com/style/word-list#under

Very minor, so feel free to break the rule on this one. But do recall our UI tends to change, so what's "under" now may be "next to" in another version.

docs/semgrep-supply-chain/license-compliance.md
```diff
@@ -135,4 +135,22 @@ Exempted packages appear in the Supply Chain > **Settings** tab.
...
+3. In the **Add custom dependency exception** window that appears:
+   1. Select the **Ecosystem** where this dependency applies
```
s-santillan, line 151:
```suggestion
   1. Select the **Ecosystem** where this dependency applies.
```

docs/semgrep-supply-chain/license-compliance.md
gautambhat: @khorne3 Thanks for asking the clarifying question - curious if you tested the behaviour described here yourself? 

I expect this workflow to do the same thing under the hood as clicking on the "allow license exception button" from the Dependencies page - the reason we added this was because there would be dependencies that aren't present on that list in UI beforehand, and hence there is no way to create an exception for it with manually adding an entry in the DB.

With this workflow, if they are currently on `bitwarden/cli@2023.9.0` and want to upgrade to `bitwarden/cli@2023.9.1` without Semgrep's license policy blocking the PR, they will need to create a custom dependency exception for the package they are upgrading **to**, i.e. `bitwarden/cli@2023.9.1`.

I know it's not a perfect workflow, but it was done as a stop gap for the time-being, while we flush out a better, longer term license policy framework. Happy to provide more context if needed :)

khorne3: When I tested the new functionality, it behaved exactly as the license exception button, so if that's what's expected, we're good to go here (tbh, I think that was part of my confusion -- I was expecting differing behavior). I'll update the text to make it clearer how these two methods are the same/different. Thank you!

gautambhat: No problem! I'm happy to review the updated docs when they're ready!

khorne3: @gautambhat please let me know what you think of the updates. thanks!

gautambhat (approved):

docs/semgrep-supply-chain/license-compliance.md
```diff
@@ -121,18 +121,51 @@ Some packages allow multiple licenses. Semgrep treats packages with multiple lic
...
+### Create custom dependency exceptions
+
+Custom dependency exceptions allow you to manually allowlist a dependency to
+prevent Semgrep from blocking a pull request or merge request due to licensing
+issues.
+
+For example, if `bitwarden/cli@2023.9.0`, which has a GPL-3.0 license, is on the
+allowlist, you must add an additional exception when upgrading to
+`bitwarden/cli@2023.9.1`. However, the dependency to which you're upgrading
+isn't yet listed in **Dependencies**; they appear only *after* you've scanned
```
s-santillan, line 151:
```suggestion
isn't yet listed in **Dependencies**; they appear only **after** you've scanned
```

PR #1352, opened 2024-02-09
Add note about mixed content

armchairlinguist:
The intro on this page suggested generic mode is mostly for unsupported languages or structured data, but it can also be used in other cases where parsing is more challenging, such as mixed content.

### Please ensure

### DIFF
--- a/docs/writing-rules/generic-pattern-matching.md
+++ b/docs/writing-rules/generic-pattern-matching.md
@@ -11,7 +11,7 @@ description: "Semgrep can match generic patterns in languages that it doesn’t 
 
 ## Introduction
 
-Semgrep can match generic patterns in languages that it does **not** yet support. Use generic pattern matching for languages that do not have a parser, configuration files, or other structured data such as XML. Generic pattern matching can also be useful in files with mixed content even if the languages are supported, such as HTML with embedded JavaScript or PHP code. In those cases you can also consider [Extract mode (experimental)](/docs/writing-rules/experiments/extract-mode/), but generic patterns may be simpler and still effective.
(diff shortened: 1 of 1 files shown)


docs/writing-rules/generic-pattern-matching.md
```diff
@@ -11,9 +11,9 @@ description: "Semgrep can match generic patterns in languages that it doesn’t
 
 ## Introduction
 
-Semgrep can match generic patterns in languages that it does **not** yet support. Use generic pattern matching for languages that do not have a parser, configuration files, or other structured data such as XML.
+Semgrep can match generic patterns in languages that it does **not** yet support. Use generic pattern matching for languages that do not have a parser, configuration files, or other structured data such as XML. Generic pattern matching can also be useful in files with mixed content even if the languages are supported, such as HTML with embedded JavaScript or PHP code. In those cases you can also consider [Extract mode (experimental)](/docs/writing-rules/experiments/extract-mode/), but generic patterns may be simpler and still effective.
```
khorne3, line 14: "Mixed content" means a file with multiple languages used in the file, right? (Perhaps it would be helpful to add a brief definition here)

khorne3 (approved):

docs/writing-rules/generic-pattern-matching.md
armchairlinguist: Yeah, good point - I'm not really trying to use or define a term, so I was hoping the examples would work. Maybe I should use something like "in files containing code in more than one supported language".

khorne3: Yeah, I like that!

armchairlinguist: Perfect, I updated. Appreciate the review! I'm interested in having someone from PA or SR review as well, so will wait a bit on merge.

PR #1358, opened 2024-02-15
Various updates to "Customization"

armchairlinguist:
We were missing some updates to the features we have (PR blocking) and also said some things that aren't quite right.

### Please ensure

### DIFF
--- a/docs/semgrep-supply-chain/getting-started.md
+++ b/docs/semgrep-supply-chain/getting-started.md
@@ -118,7 +118,7 @@ Semgrep Supply Chain supports the scanning of monorepos. As outlined in [Project
 
 ## Block pull or merge requests
 
-Semgrep Supply Chain with Semgrep versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but more recent versions of Semgrep do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings with high or critical severity.
+Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings with high or critical severity.
 
 1. Log in to Semgrep Cloud Platform.
 2. Click **Supply Chain** > **Settings** on the header menu.
--- a/docs/semgrep-supply-chain/getting-started.md
+++ b/docs/semgrep-supply-chain/getting-started.md
@@ -118,7 +118,7 @@ Semgrep Supply Chain supports the scanning of monorepos. As outlined in [Project
 
 ## Block pull or merge requests
 
-Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings with high or critical severity.
+Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings in direct dependencies with high or critical severity. 
 
 1. Log in to Semgrep Cloud Platform.
 2. Click **Supply Chain** > **Settings** on the header menu.
--- a/docs/semgrep-supply-chain/getting-started.md
+++ b/docs/semgrep-supply-chain/getting-started.md
@@ -60,9 +60,9 @@ Semgrep Supply Chain frequently receives rule updates. To take advantage of thes
 
 <CiScheduling />
 
-:::note Rules updates
+:::note Rule updates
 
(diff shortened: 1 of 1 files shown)


s-santillan:
@khorne3 I have done some reviews from this batch, but I wanted to give you first choice on reviewing this because this is your work :)

docs/semgrep-supply-chain/getting-started.md
```diff
@@ -114,11 +114,17 @@ Additionally, you can view your results in Semgrep Cloud Platform (SCP). SCP dis
...
+Semgrep Supply Chain supports the scanning of monorepos. As outlined in [Project directory structure](#project-directory-structure), findings are grouped by directory based on the [lockfiles](/semgrep-supply-chain/glossary/#lockfile) or manifest files present in the monorepo.
 
 ## Block pull or merge requests
 
-Though Semgrep Supply Chain **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, it no longer does this. You can, however, configure your version control system to prevent merging if Semgrep Supply Chain identifies reachable findings.
+Semgrep Supply Chain with Semgrep versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but more recent versions of Semgrep do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings with high or critical severity.
```
khorne3, line 121: Would this make sense? 

```suggestion
Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings with high or critical severity.
```

khorne3 (approved):
one minor thing, otherwise LGTM!

armchairlinguist:
@mmcqd if you have a moment, I'd love your review here to make sure what I'm saying makes sense, particularly on the monorepo and PR blocking parts.

docs/semgrep-supply-chain/getting-started.md
```diff
@@ -52,17 +52,17 @@ If you have code files in `my-project/biking`, Semgrep Supply Chain does not ass
...
 
-Semgrep Supply Chain frequently receives rule updates. To take advantage of these updates and increase the frequency with which Semgrep Supply Chain scans your codebase.
+Semgrep Supply Chain frequently receives rule updates. To take advantage of these updates, adjust the frequency with which Semgrep Supply Chain scans your codebase.
 
 <CiScheduling />
 
 :::note Rules updates
 
-If a rule is updated, findings generated against the updated rule are considered **new findings**, even if the previous version of the rule generated a finding. Furthermore, because the finding is new, you'll receive notifications through the channels you've set up (e.g., Slack, email).
+If a rule is updated, findings generated against the updated rule are considered **new findings**, even if the previous version of the rule generated a finding. Because the finding is new, you'll receive notifications through the channels you've set up (e.g., Slack, email).
```
armchairlinguist, line 65: @andy-r2c This line had stuck with me since I made a wording change unrelated to the content, but based on separate discussion it sounds like it might not even be correct? Could you confirm?

docs/semgrep-supply-chain/getting-started.md
```diff
@@ -114,11 +114,17 @@ Additionally, you can view your results in Semgrep Cloud Platform (SCP). SCP dis
...
+Semgrep Supply Chain supports the scanning of monorepos. As outlined in [Project directory structure](#project-directory-structure), findings are grouped by directory based on the [lockfiles](/semgrep-supply-chain/glossary/#lockfile) or manifest files present in the monorepo.
 
 ## Block pull or merge requests
 
-Though Semgrep Supply Chain **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, it no longer does this. You can, however, configure your version control system to prevent merging if Semgrep Supply Chain identifies reachable findings.
+Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings with high or critical severity.
```
mmcqd, line 121: Slight subtlety: we will block on any reachable code finding with these severities, but we will only block on a lockfile-only "always reachable" finding if it is high/critical and it's a _direct_ dependency. Always reachable findings on transitive dependencies won't block, because we thought it would be too noisy.

mmcqd (approved):

docs/semgrep-supply-chain/getting-started.md
mmcqd: It depends whether we consider it worth it to bring up the "directly importing a transitive dependency" edge case here. The blocking situation is this:

 |  | Reachable in code      | Always reachable in lockfile |
| ----------- | ----------- | ----------- |
| Direct dep      | Block       | Block |
| Transitive dep  | Block*        | No Block |

\* Only happens in edge case that should be avoided

Feels hard to keep it snappy 😅

armchairlinguist: Yep - I'm not sure we need to cover the weird edge case in the glossary, but good point that we might want to consider covering it elsewhere - it might be a good fit for the KB.

PR #1359, opened 2024-02-15
Clarify why we parse manifest files

armchairlinguist:
For some reason I thought this info already existed, but basically we should be clear about what the presence of a manifest file does (determine transitivity).

### Please ensure

### DIFF
--- a/docs/supported-languages.md
+++ b/docs/supported-languages.md
@@ -87,7 +87,7 @@ Visit the Semgrep public language dashboard to see the parse rates for each lang
 
 <SscIntro/>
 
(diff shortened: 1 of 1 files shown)


s-santillan (approved):
A mention of the manifest being used for transitivity does exist ([glossary](https://semgrep.dev/docs/semgrep-supply-chain/glossary/#manifest-file)), but we could always mention it again :) 

thank you, lgtm!

armchairlinguist:
Yep - I figured I'd mention it here for clarification and link to that. I'm not sure anyone normally reads the glossary. :)

PR #1360, opened 2024-02-15
Clarify some wording in SSC Glossary

armchairlinguist:
Some of the wording here was less than clear or a bit awkward - this should clear things up, particularly for the false positive rate and transitive dependency.

### Please ensure


docs/semgrep-supply-chain/glossary.md
```diff
@@ -25,19 +25,19 @@ Exploitability is the practical assessment of a vulnerability's threat, typicall
 
 ## Lockfile
 
-A lockfile describes a dependency tree to ensure that deployments and organizations install the same **dependencies and exact versions **for their codebase. Lockfile information includes versions of the dependency and any transitive (indirect) dependencies. Lockfiles are automatically generated by a package manager such as `pip` or `npm`.
+A lockfile describes a dependency tree to ensure that deployments and organizations install the same **dependencies and exact versions** for their codebase. Lockfile information includes versions of the dependency and any transitive (indirect) dependencies. Lockfiles are automatically generated by a package manager such as `pip` or `npm`.
 
-Semgrep Supply Chain uses lockfiles as part of its reachability analysis by determining the exact version of a dependency that a codebase is using.
+Semgrep Supply Chain uses lockfiles as part of its analysis to determine the exact version of a dependency that a codebase is using.
```
s-santillan, line 30: I see that we are removing the "reachability analysis" wording? Previously, the reason for specifying "reachability analysis" is because part of an SSC rule's makeup is to specify the vulnerability's affected versions (the `version` key). Something is only reachable if they're using those affected versions. The lockfile tells us this information. Hence we use the lockfile to determine reachability.

docs/semgrep-supply-chain/glossary.md
```diff
@@ -53,19 +53,19 @@ A package registry stores dependencies and provides a means to upload or downloa
...
 
-Continuous integration scans with Semgrep Supply Chain rules block any merge or pull requests upon detecting any reachable findings.
+Continuous integration scans with Semgrep Supply Chain rules can block pull or merge requests upon detecting any reachable findings.
 
 See also [Reachability](#reachability).
 
 ## Reachability
 
-Reachability refers to whether or not a vulnerable piece of code from a dependency is used in the codebase itself. In Semgrep Supply Chain, both a dependency's vulnerable version and code pattern must match for a vulnerability to be considered reachable.
+Reachability refers to whether or not a vulnerable piece of code from a dependency is used in the related codebase. In Semgrep Supply Chain, both a dependency's vulnerable version and code pattern must match for a vulnerability to be considered reachable.
```
s-santillan, line 62:
```suggestion
Reachability refers to whether or not a vulnerable piece of code from a dependency is used in the codebase importing it. In Semgrep Supply Chain, both a dependency's vulnerable version and code pattern must match for a vulnerability to be considered reachable.
```

Not sure if my suggestion is any better, tbh.

docs/semgrep-supply-chain/glossary.md
```diff
@@ -53,19 +53,19 @@ A package registry stores dependencies and provides a means to upload or downloa
...
 
 See [Overview of Semgrep Supply Chain](/semgrep-supply-chain/overview) to learn how Semgrep leverages its code-scanning and rule syntax capabilities to provide high-signal rules that determine a finding's reachability. This assists security engineers in remediation and triage processes.
 
 ## Reachability rules
 
-A type of Semgrep Supply Chain rule that performs reachability analysis. A reachability rule can determine if the vulnerable piece of code from a dependency is used in the codebase itself.
+A type of Semgrep Supply Chain rule that performs reachability analysis. A reachability rule can determine if the vulnerable piece of code from a dependency is used in the related codebase.
```
s-santillan, line 68:
```suggestion
A type of Semgrep Supply Chain rule that performs reachability analysis. A reachability rule can determine if the vulnerable piece of code from a dependency is used in the codebase that imports it.
```

s-santillan (approved):
lgtm, minor comments only!

docs/semgrep-supply-chain/glossary.md
armchairlinguist: @s-santillan that's a bit of a strange way of looking at it from my perspective. The dependency is affected at specific versions even if the rule is lockfile-only. If the rule also has reachability, once we determine an affected version is present we can look for code usages. But the version determination is not really related to reachability in my mind. I'm happy for @andy-r2c to disagree though, that's one reason I wanted his review!

docs/semgrep-supply-chain/glossary.md
armchairlinguist: I think it's good, but should we go with "importing it" or "that imports it"? Probably want to use the same here and below.

s-santillan: I went with "that imports it" and updated the suggestion accordingly. Thank you!

docs/semgrep-supply-chain/glossary.md
s-santillan: I stand corrected! - let's remove "reachability", since it works for both analyses.

docs/semgrep-supply-chain/glossary.md
```diff
@@ -53,19 +53,19 @@ A package registry stores dependencies and provides a means to upload or downloa
...
+Reachability refers to whether or not a vulnerable piece of code from a dependency is used in the codebase that imports it. In Semgrep Supply Chain, both a dependency's vulnerable version and code pattern must match for a vulnerability to be considered reachable.
 
 See [Overview of Semgrep Supply Chain](/semgrep-supply-chain/overview) to learn how Semgrep leverages its code-scanning and rule syntax capabilities to provide high-signal rules that determine a finding's reachability. This assists security engineers in remediation and triage processes.
 
 ## Reachability rules
 
-A type of Semgrep Supply Chain rule that performs reachability analysis. A reachability rule can determine if the vulnerable piece of code from a dependency is used in the codebase itself.
+A type of Semgrep Supply Chain rule that performs reachability analysis. A reachability rule can determine if the vulnerable piece of code from a dependency is used in the codebase that imports it.
```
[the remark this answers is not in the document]
armchairlinguist: I'm open to that, but let's do it in a separate PR - currently the definition of reachability here doesn't really touch on that either, so the page would need a broader reworking to be consistent.

PR #1375, opened 2024-02-23
Provide simplified `semgrep publish` GitHub Actions job

armchairlinguist:
I'm not sure why the config on this job was invoking Docker explicitly; it might just be older syntax. This config is simpler and better matches our current sample CI config for regular Semgrep runs.

### Please ensure


s-santillan (approved):
Wow, this is one ooooold doc. Thank you for updating it!

PR #1379, opened 2024-02-23
Add GitLab workflow example for semgrep publish

armchairlinguist:
Adds a basic GitLab workflow file allowing publishing of private semgrep rules from GitLab.

### Please ensure


swati-sharma (approved):

khorne3 (approved):
I indented the addition so it lined up properly and edited the text a bit

armchairlinguist:
Ah I see, I wasn't thinking of it as part of step 2, but it probably does make more sense that way. Thanks!

PR #1389, opened 2024-02-29
update with lockfile only entry for kotlin gradle

s-santillan:

### Please ensure


khorne3 (approved):
Unless there's a reason why Maven is listed before Gradle, I think we should alphabetize these?

s-santillan:
Thank you for the reminder to alphabetize. It's fixed in the latest.

image

andy-r2c (approved):

PR #1394, opened 2024-03-01
Use specified versions of @markprompt/*

nickrttn:
👋 

Hi @s-santillan,

This PR updates the `package.json` and `yarn.lock` files to make your build work. 

For now, we unfortunately can't support newer versions of `@markprompt/{core,css,react}` because they depend on packages that require `react@18`. Docusaurus 2 is still on React 17 and so you would need an upgrade to Docusaurus 3 to be able to use the newer packages.

That said, I went through upgrading a few of our Docusaurus demo's to Docusaurus 3 and the process is really not as painful as it looks from reading the migration guide!

Hope this helps :)

### DIFF
--- a/docs/extensions/overview.md
+++ b/docs/extensions/overview.md
@@ -36,7 +36,7 @@ The [pre-commit framework](https://pre-commit.com/) can run `semgrep` at commit-
 ```yaml
 repos:
 - repo: https://github.com/semgrep/semgrep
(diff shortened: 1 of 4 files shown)


s-santillan:
I had to make a separate PR to bump up the node version in our Build and Deploy workflow to 20. However, we did add:
```
"@markprompt/docusaurus-theme-search": "~0.15.0",
```
And removed the resolutions, and it did build. Closing this one.

PR #1446, opened 2024-03-22
Update taint docs

IagoAbal:
Closes FLOW-51
Closes FLOW-52
Closes FLOW-53
Closes FLOW-54
Closes FLOW-55
Closes FLOW-60
Closes FLOW-64

### Please ensure


s-santillan:
Hi @IagoAbal,

Thanks for this huge update. Your writing based on what I saw during the presentation looks solid and well-presented.

This is the first time I'll be reviewing a rule-writing doc. It may be helpful to review the hows and whys of a writing review so everyone knows the basis behind our notes and can tell us where the notes are helpful or not, given how specialized the topic is.

- The TW team uses the [Google style guide](https://developers.google.com/style) to keep the tone, tense, and other parts of speech consistent.
    -  The goal of consistency in the docs is to keep the user focused on the content and meaning. When the tone in one doc sounds different from another, it jolts the user out of the experience and makes them focus on the style rather than the substance.
    - The rules sometimes seem excessive or silly (for example, "via" has to be replaced with "through") but they come from a place of accessibility for English speakers with a wide range of experience.
- We try to make the text look and feel more digestible through the ff:
    - Using less words overall where we can.
    - Breaking up long paragraphs.
    
I hope this helps explain some of the notes here. Some notes will be nits for sure, so feel free to push back or give feedback whenever.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -2,12 +2,12 @@
 slug: taint-mode
 append_help_link: true
 description: >-
-  Taint mode allows you to write simple rules that catch complex injection bugs.
+  Taint mode allows you to write simple rules that catch complex injection bugs thanks to taint analysis (aka taint tracking).
```
s-santillan, line 5:
```suggestion
  Taint mode allows you to write simple rules that catch complex injection bugs thanks to taint analysis.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -2,12 +2,12 @@
...
+# Taint analysis
 
-Semgrep supports intra-procedural [taint analysis](https://en.wikipedia.org/wiki/Taint_checking). This data-flow analysis feature tracks the flow of untrusted (**tainted**) data throughout the body of a function or method. Tainted data originate from tainted **sources**. If tainted data are not transformed or checked accordingly (**sanitized**), taint analysis reports a finding whenever tainted data reach a vulnerable function (called **sink**). Tainted data flow from sources to sinks through **propagators** (such as assignments, or function calls).
+Semgrep supports [taint analysis](https://en.wikipedia.org/wiki/Taint_checking) (or taint tracking) via so-called taint rules (`mode: taint`). Taint analysis is a data-flow analysis that tracks the flow of untrusted (**tainted**) data throughout the body of a function or method. Tainted data originate from tainted **sources**. If tainted data are not transformed or checked accordingly (**sanitized**), taint analysis reports a finding whenever tainted data reach a vulnerable function (called **sink**). Tainted data flow from sources to sinks through **propagators** (such as assignments, or function calls).
```
s-santillan, line 10:
```suggestion
Semgrep supports [taint analysis](https://en.wikipedia.org/wiki/Taint_checking) (or taint tracking) through taint rules (specified by adding `mode: taint` to your rule). Taint analysis is a data-flow analysis that tracks the flow of untrusted, or **tainted** data throughout the body of a function or method. Tainted data originate from tainted **sources**. If tainted data are not transformed or checked accordingly (**sanitized**), taint analysis reports a finding whenever tainted data reach a vulnerable function, called a **sink**. Tainted data flow from sources to sinks through **propagators**, such as assignments, or function calls.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
-  - pattern: make_tainted($X)
-  - focus-metavariable: $X
+  - patterns:
+      - pattern: make_tainted($X)
+      - focus-metavariable: $X
+    by-side-effect: true
 ```
 
-**Deprecated workaround**: The previously recommended workaround was to write a source specification such as the one specified below. This instructs Semgrep that **every** occurrence of `$X` after `make_tainted($X)` must be considered a source.
+When this option is enabled, and the source specification matches a variable (or in general, an l-value) exactly, then Semgrep assumes that the variable (or l-value) becomes tainted by side-effect at the precise places where the source specification produces a match.
```
s-santillan, line 113: For my own knowledge, what is an `l-value`?

s-santillan, line 113:
```suggestion
When this option is enabled, and the source specification matches a variable (or in general, an [l-value](https://en.wikipedia.org/wiki/Value_(computer_science)#lrvalue)) exactly, then Semgrep assumes that the variable (or l-value) becomes tainted by side-effect at the precise places where the source specification produces a match.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=5r400" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+The matched occurrences themselves are considered tainted too, that is the occurrence of `x` in `make_tainted(x)` is itself tainted too. If you do not want this to be the case, then set `by-side-effect: only` instead.
```
s-santillan, line 117:
```suggestion
The matched occurrences themselves are considered tainted; that is, the occurrence of `x` in `make_tainted(x)` is itself tainted too. If you do not want this to be the case, then set `by-side-effect: only` instead.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
+
+:::note
+It is important to use `focus-metavariable: $X` to focus the match on the l-value that we want to taint, otherwise `by-side-effect` does not work as expected.
```
s-santillan, line 120:
```suggestion
You must use `focus-metavariable: $X` to focus the match on the l-value that you want to taint, otherwise `by-side-effect` does not work.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
+:::
+
+If the source does not set `by-side-efect`, then only the very occurrence of `x` in `make_tainted(x)` will be tainted, but not the occurrence of `x` in `sink(x)`. The source specification matches only the first occurrence and, without `by-side-effect: true`, Semgrep does not know that `make_tainted` is updating the variable `x` by side-effect. Thus, a taint rule using such specification does not produce any finding.
```
s-santillan, line 123:
```suggestion
If the source does not set `by-side-efect`, then only the very occurrence of `x` in `make_tainted(x)` will be tainted, but not the occurrence of `x` in `sink(x)`. The source specification matches only the first occurrence and, without `by-side-effect: true`, Semgrep does not know that `make_tainted` is updating the variable `x` by side-effect. Thus, a taint rule using such a specification does not produce any finding.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
+
+:::info
+One can be tempted to write a source specification as the one below (and this was the official workaround before `by-side-effect`):
```
s-santillan, line 126:
```suggestion
One can be tempted to write a source specification as the following example (and this was the official workaround before `by-side-effect`):
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=Zq5ow" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+For many rules this distinction is not very meaningful because it does not always make sense that a sink occurs inside the arguments of a source function.
+
+:::note
+If one of your rules relies on non-exact matching of sources, we advice you to make it explicit with `exact: false`, even if it is the current default, so that your rule does not break if the default changes.
+:::
+
 ### Sources by side-effect
```
s-santillan, line 94: I'm just trying to wrap my head around the idea of a tainted side-effect. Given:
```
function add10(x) {
x = x + 10
return x
}
```
I assume the above is a pure function (no side effects), deterministic output.

So a side-effect would be the value of y in the ff example
```
function addy(y, x) {
y = getUserInputNumber()
x = x + y
return x
```
So if we wanted to catch `y`, we'd say:
```
 - patterns:
      - pattern: addy($Y, $X)
      - focus-metavariable: $Y
    by-side-effect: true
```
Is this correct?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -31,90 +31,116 @@ Here Semgrep tracks the data returned by `get_user_input()`, which is the source
...
+  - pattern-inside: |
       make_tainted($X)
       ...
   - pattern: $X
 ```
 
-This approach had two main limitations. First, it overrides any sanitization that can be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` is reported as tainted despite `x` was re-assigned to a fresh set!
+This tells Semgrep that **every** occurrence of `$X` after `make_tainted($X)` must be considered a source.
+
+This approach has two main limitations. First, it overrides any sanitization that can be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` is reported as tainted despite `x` has been sanitized!
```
s-santillan, line 139:
```suggestion
This approach has two main limitations. First, it overrides any sanitization that can be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` is reported as tainted despite the fact that `x` has been sanitized!
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+```yaml
+pattern-sources:
+  - patterns:
+    - pattern-inside: |
+        def foo($X, ...):
+          ...
+    - focus-metavariable: $X
+```
+
+Note that the use of `focus-metavariable: $X` is very important, and using `pattern: $X` is **not** equivalent. With `focus-metavariable: $X` we match exactly the formal parameter. Click "Open in Playground" below and use "Inspect Rule" to visualize what the source is matching.
```
s-santillan, line 171:
```suggestion
Note that the use of `focus-metavariable: $X` is very important, and using `pattern: $X` is **not** equivalent. With `focus-metavariable: $X`, Semgrep matches the formal parameter exactly. Click "Open in Playground" below and use "Inspect Rule" to visualize what the source is matching.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=L1vJ6" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+Now do the same with this other taint rule that uses `pattern: $X`. The `pattern: $X` does not match the formal parameter itself, but matches all its uses inside the function definition. Even if `x` is sanitized via `x = sanitize(x)`, the occurrence of `x` inside `sink(x)` is a taint source itself (due to `pattern: $X`) and so `sink(x)` is tainted!
```
s-santillan, line 175:
```suggestion
The following example does the same with this other taint rule that uses `pattern: $X`. The `pattern: $X` does not match the formal parameter itself, but matches all its uses inside the function definition. Even if `x` is sanitized via `x = sanitize(x)`, the occurrence of `x` inside `sink(x)` is a taint source itself (due to `pattern: $X`) and so `sink(x)` is tainted!
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+pattern-sources:
+- pattern: source(...)
+  control: true
+```
+
+This is useful for checking _reachability_, that is to check if from a given code location the control-flow can reach another code location, regardless of whether there is any flow of data between them. In the following example we check whether `foo()` could be followed by `bar()`:
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=yyjrx" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+In essence, using a control source we can define a context, from which Semgrep will detect if a call to some other code (i.e., the sink) can be reached.
```
s-santillan, line 195:
```suggestion
By using a control source, you can define a context from which Semgrep detects if a call to some other code, such as a sink, can be reached.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+
+:::note
+Using [taint labels](#taint-labels-pro-) it is possible to combine both data and control sources in the same rule.
```
s-santillan, line 198:
```suggestion
Use [taint labels](#taint-labels-pro-) to combine both data and control sources in the same rule.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+```
+
+Once the source is "exact", Semgrep will no longer consider subexpressions as sanitized, and `sink("taint")` inside `sanitize(sink("taint"))` will be reported as a tainted sink.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=Zqz8o" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+For many rules this distinction is not very meaningful because it does not always make sense that a sink occurs inside the arguments of a sanitizer function.
+
+:::note
+If one of your rules relies on non-exact matching of sanitizers, we advice you to make it explicit with `exact: false`, even if it is the current default, so that your rule does not break if the default changes.
```
s-santillan, line 249:
```suggestion
If one of your rules relies on non-exact matching of sanitizers, We at Semgrep advise you to make it explicit with `exact: false`, even if it is the current default, so that your rule does not break if the default changes.
```
https://developers.google.com/style/word-list#we

Just making it clear who "we" are, at least at this point in the doc.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+<iframe src="https://semgrep.dev/embed/editor?snippet=4bvGz" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+:::note
+It is important to use `focus-metavariable: $X` to focus the match on the l-value that we want to sanitize, otherwise `by-side-effect` does not work as expected.
+:::
+
+If the sanitizer does not set `by-side-efect`, then only the very occurrence of `x` in `check_if_safe(x)` will be sanitized, but not the occurrence of `x` in `sink(x)`. The sanitizer specification matches only the first occurrence and, without `by-side-effect: true`, Semgrep does not know that `check_if_safe` is updating/sanitizing the variable `x` by side-effect. Thus, a taint rule using such specification does produce a finding for `sink(x)` in the example above.
+
+:::info
+One can be tempted to write a sanitizer specification as the one below (and this was the official workaround before `by-side-effect`):
```
s-santillan, line 282:
```suggestion
You can be tempted to write a sanitizer specification as the one below (and this was the official workaround before `by-side-effect`):
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+- patterns:
+  - pattern-inside: |
+      check_if_safe($X)
+      ...
+  - pattern: $X
+```
+
+This tells Semgrep that **every** occurrence of `$X` after `check_if_safe($X)` must be considered santized.
+
+This approach has two main limitations. First, it overrides any further tainting that can be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` is  **not** reported as tainted despite `x` has been tainted!
```
s-santillan, line 295:
```suggestion
This approach has two main limitations. First, it overrides any further tainting that can be performed on the code matched by `$X`.  In the example code below, the call `sink(x)` is  **not** reported as tainted despite `x` having been tainted!
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+
+```python
+check_if_safe(x)
+x = source()
+sink(x) # false negative
+```
+
+Note also that [`...` ellipses operator](/writing-rules/pattern-syntax/#ellipses-and-statement-blocks) has limitations. For example, in the code below Semgrep does still match despite `x` has been sanitized in both branches:
```
s-santillan, line 303:
```suggestion
Note also that [`...` ellipses operator](/writing-rules/pattern-syntax/#ellipses-and-statement-blocks) has limitations. For example, in the code below Semgrep still matches despite `x` having been sanitized in both branches:
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+    check_if_safe(x)
+sink(x) # false positive
+```
+
+The `by-side-effect` option was added precisely [to address those limitations](https://semgrep.dev/playground/s/PeB3W). However, that kind of workaround can still be useful in other situations!
+:::
+
+## Sinks
+
+A taint sink is specified by a pattern. Like in a search-mode rule, you can start this pattern with one of the following keys: `pattern`, `patterns`, `pattern-either`, `pattern-regex`. Unlike for sources and sanitizers, by default Semgrep does not consider the subexpressions of the matched expressions as sinks.
```
s-santillan, line 318:
```suggestion
A taint sink is specified by a pattern. Like in a search-mode rule, you can start this pattern with one of the following keys: `pattern`, `patterns`, `pattern-either`, `pattern-regex`. Unlike sources and sanitizers, by default Semgrep does not consider the subexpressions of the matched expressions as sinks.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+    - pattern: sink($SINK, ...)
+    - focus-metavariable: $SINK
+```
 
-## Propagators
+This rule causes Semgrep to only annotate the first parameter passed to `sink` as the sink, rather than the function `sink` itself. If taint goes into any other parameter of `sink`, then that is not considered a problem.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=v83Nl" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+Remember, anything that you can matchn with Semgrep can be made into a sink, like the index in an array access:
```
s-santillan, line 374:
```suggestion
Anything that you can match with Semgrep can be made into a sink, like the index in an array access:
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+If you specify a sink such as `sink(...)` then any tainted data passed to `sink`, through any of its arguments, results in a finding.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=OrAAe" border="0" frameBorder="0" width="100%" height="432"></iframe>
+:::
+
+### At-exit sinks (Pro) 🧪
+
+**At-exit taint sinks is a Semgrep Pro feature.**
+
+At-exit sinks are meant to facilitate writing leak-detection rules using taint-mode. By setting `at-exit: true` you can restrict a sink specification to only match at "exit" statements, that is statements after which the control-flow will exit the function being analyzed.
```
s-santillan, line 393:
```suggestion
At-exit sinks are meant to facilitate writing leak-detection rules using taint mode. By setting `at-exit: true` you can restrict a sink specification to only match at "exit" statements, that is statements after which the control-flow will exit the function being analyzed.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+  - pattern: return ...
+  - pattern: $F(...)
+  at-exit: true
+```
+
+The above sink pattern matches either `return` statements (which are always "exit" statements), or function calls occuring as "exit" statements.
+
+Unlike regular sinks, at-exit sinks trigger a finding if any tainted l-value reaches the location of the sink. For example, the at-exit sink specification above will trigger a finding at a `return 0` statement if some tainted l-value reaches the `return`, even if `return 0` itself is not tainted. The location itself is the sink rather than the code that is at that location.
+
+We can use this, for example, to check that file descriptors are being closed within the same function where they were opened.
```
s-santillan, line 407:
```suggestion
You can use this, for example, to check that file descriptors are being closed within the same function where they were opened.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=OrAzB" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+The `print(content)` statement is reported because the control-flow exits the function at that point, and the file has not been closed.
```
s-santillan, line 411:
```suggestion
The `print(content)` statement is reported because the control-flow exits the function at that point, and the file has not been closed.
```
```suggestion
The `print(content)` statement is reported because the control flow exits the function at that point, and the file has not been closed.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -126,18 +152,300 @@ if cond:
...
+
+```yaml
+pattern-propagators:
+- pattern: strcpy($DST, $SRC)
+  from: $SRC
+  to: $DST
+```
+
+:::info
+Taint propagators only work intra-procedurally, that is, within a function or method. You cannot use taint propagators to propagate taint across different functions/methods, that is the job of [inter-procedural analysis](#inter-procedural-analysis-pro).
```
s-santillan, line 444:
```suggestion
Taint propagators only work intra-procedurally, that is, within a function or method. You cannot use taint propagators to propagate taint across different functions/methods. Use [inter-procedural analysis](#inter-procedural-analysis-pro).
```
Just confirming: does this mean to say that if a user flips on Semgrep Pro Engine, their taint rules that use propagators will all be able to detect interfile propagations? Can we say something like: turn on Semgrep Pro Engine so that your rules with propagators can perform interfile analysis.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-2. The `to` and `from` metavariables. These metavariables should match a **variable** or an **expression**.
-    - The `from` metavariable specifies the entry point of the taint.
-    - The `to` metavariable specifies where the tainted data is propagated to, typically an object or data structure.
+### Propagation without side-effect
 
-:::note
-When the `to` metavariable matches a variable (or in general, an l-value), the taint is propagated _by side-effect_.
-See [Sources by side effect](#sources-by-side-effect) for an explanation of what this means.
-:::
+Taint propagators can be used in very imaginative ways, and in some cases we may not want taint to propagate by side-effect. This can be achieved by disabling `by-side-effect` (enabled be default).
```
s-santillan, line 505:
```suggestion
Taint propagators can be used in very imaginative ways, and in some cases you may not want taint to propagate by side-effect. This can be achieved by disabling `by-side-effect`, which is enabled be default.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-Given the sanitizer specification below, and a piece of code such as `sanitize(sink(source))`, the call `sink(source)` is **not** reported.
-The reason is that the pattern `sanitize(...)` matches all of `sanitize(sink(source))`, and that makes Semgrep label every subexpression in that piece of code as being sanitized. In particular, `source` is considered to be sanitized!
+<iframe src="https://semgrep.dev/embed/editor?snippet=4bv6x" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
-```yaml
-pattern-sanitizers:
-- pattern: sanitize(...)
-```
-:::
+Because we are disabling `by-side-effect`, the `sink` occurence that is inside the `if` block is tainted, but this does not affect the `sink` occurrence outside the `if` block.
```
s-santillan, line 526:
```suggestion
Because the rule disables `by-side-effect`, the `sink` occurrence that is inside the `if` block is tainted, but this does not affect the `sink` occurrence outside the `if` block.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-    raise Exception("unsafe")
+<!-- <iframe src="https://semgrep.dev/embed/editor?snippet=KxJRL" border="0" frameBorder="0" width="100%" height="432"></iframe> -->
 
-x = tainted
-check_if_safe(x)
-sink(x)
-```
+### Deduplication of findings
+
+Semgrep tracks all the possible ways that taint can reach a sink, but at present it only reports one taint trace among the possible ones. Click "Open in Playground" in the example below, run the example to get one finding, and then ask the Playground to visualize the dataflow of the finding. Even though `sink` can be tainted via `x` or via `y`, the trace will only show you one of these possibiities. If you e.g. replace `x = user_input` with `x = "safe"`, then Semgrep will then report the taint trace via `y`.
```
s-santillan, line 536:
```suggestion
Semgrep tracks all the possible ways that taint can reach a sink, but at present it only reports one taint trace among the possible ones. Click "Open in Playground" in the example below, run the example to get one finding, and then ask the Playground to visualize the dataflow of the finding. Even though `sink` can be tainted via `x` or via `y`, the trace will only show you one of these possibiities. If you replace `x = user_input` with `x = "safe"`, then Semgrep will then report the taint trace via `y`.
```
> Ask the playground to visualize...
Do you mean clicking on Inspect rule in the playground?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=WAYzL" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
-**Deprecated workaround**: The previously recommended workaround was to write a sanitizer specification such as the one below. This instructs Semgrep that **every** occurrence of `$X` after `check_if_safe($X)` must be considered safe.
+### Report findings on the sources (Pro)
+
+**Reporting findings on the source of taint is a Semgrep Pro feature.**
+
+By default Semgrep reports taint findings at the location of the sink being matched, and one must look at the taint trace to identify where the taint is coming from. It is also possible to make Semgrep report the findings at the location of the taint sources, by setting the [rule-level option](/writing-rules/rule-syntax/#options) `taint_focus_on` to `source`. Then
```
s-santillan, line 544:
```suggestion
By default Semgrep reports taint findings at the location of the sink being matched. You must look at the taint trace to identify where the taint is coming from. It is also possible to make Semgrep report the findings at the location of the taint sources, by setting the [rule-level option](/writing-rules/rule-syntax/#options) `taint_focus_on` to `source`. Then
```

s-santillan: Is something supposed to follow "then..."?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-pattern-sinks:
-  - patterns:
-    - pattern-inside: $S = new Sandbox(); ...
-    - pattern: $S.run($SINK, ...)
-    - focus-metavariable: $SINK
-```
+### Restrict taint by type (Pro)
 
-This rule causes Semgrep to only annotate the first parameter passed to `$S.run` as the sink, rather than the method `$S.run` itself. If taint goes into any other parameter of `$S.run`, then that is not considered a problem.
+By enabling `taint_assume_safe_booleans` Semgrep will automatically sanitize Boolean expressions (when it can infer that).
```
s-santillan, line 569:
```suggestion
By enabling `taint_assume_safe_booleans` Semgrep automatically sanitizes Boolean expressions when it can infer that the expression resolves to Boolean.
```

s-santillan: Correct me if I'm mistaken, but the ability to infer is one of those things that Semgrep does at different levels depending on the language. For example, we do this for Java's stdlib, but I'm not aware of if we do this for other stdlibs?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-```javascript
-pattern-sinks:
-  - patterns:
-    - pattern-inside: $ARRAY[$SINK]
-    - focus-metavariable: $SINK
-```
+<iframe src="https://semgrep.dev/embed/editor?snippet=6JvzK" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
-This way we tell Semgrep that we do not want arrays to be accessed with tainted indexes.
+Similarly, enabling `taint_assume_safe_numbers` Semgrep will automatically sanitize numeric expressions (when it can infer that).
```
s-santillan, line 575:
```suggestion
Similarly, enabling `taint_assume_safe_numbers` Semgrep will automatically sanitize numeric expressions when it can infer that the expression is numeric.
```

docs/writing-rules/data-flow/taint-mode.md
s-santillan: Stylistically, the reason why I replaced "when it can infer that" with a much longer phrase is because "that" (pronoun) can be a little vague. I just wanted to make the text clearer.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
+sink(x); // undesired finding here
 ```
 
-In some codebases, this conservative assumption can generate too many false positives. If that is the case, consider enumerating such functions as sanitizers:
-
-```yaml
-pattern-sanitizers:
-- pattern: some_safe_function(...)
-```
+This can of course generate false positives, and for certain rules on certain codebases it can produce a high amount of noise.
```
s-santillan, line 608:
```suggestion
This can generate false positives, and for certain rules on certain codebases it can produce a high amount of noise.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-For example, in the code below, `some_safe_function` receives tainted data as input, and to be on the safe side, Semgrep assumes that it also returns tainted data as output. As a result, a finding is produced.
+:::note
+Semgrep Pro's ability to infer types for expressions varies depending on the language. For example, in Python type annotations are not always present, and the `+` operator can also be used to concatenate strings. Semgrep also ignores the types of functions and classes coming from third-party libraries.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=zdjnn" border="0" frameBorder="0" width="100%" height="432"></iframe>
+:::
+
+### Assume tainted indexes are safe
+
+By default, Semgrep assumes that accessing an array-like object with a tainted index (that is, `obj[tainted]`) is itself tainted expression, even if the object itself is not tainted. Setting `taint_assume_safe_indexes: true` makes Semgrep assume that these expressions are safe.
```
s-santillan, line 589:
```suggestion
By default, Semgrep assumes that accessing an array-like object with a tainted index (that is, `obj[tainted]`) is itself a tainted **expression**, even if the **object** itself is not tainted. Setting `taint_assume_safe_indexes: true` makes Semgrep assume that these expressions are safe.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=X56pj" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+### Assume function calls are safe
+
+:::note
+We refer to a function call as _opaque_ when Semgrep does not have access to its definition, in order to examine it and determine its "taint behavior" (e.g., whether the function call propagates or not any taint that comes through its inputs). In Semgrep OSS, where taint analysis is intra-procedural, all function calls are opaque. In Semgrep Pro, with [inter-procedural taint analysis](#inter-procedural-analysis-pro), an opaque function could be one coming from a third-party library.
```
s-santillan, line 596:
```suggestion
This documentation refers to a function call as **opaque** when Semgrep does not have access to its definition, in order to examine it and determine its "taint behavior", such as whether the function call propagates or not any taint that comes through its inputs.

In Semgrep OSS, where taint analysis is intra-procedural, all function calls are opaque.

In Semgrep Pro, with [inter-procedural taint analysis](#inter-procedural-analysis-pro), an opaque function could be one coming from a third-party library.
```

s-santillan: We don't have any references at all wrt to level of "magic" that Semgrep supports per language, no? Didn't we have a reference table somewhere?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
+:::
+
+By default Semgrep considers that an _opaque_ function call propagates any taint passed via any of its arguments to its output.
```
s-santillan, line 599:
```suggestion
By default Semgrep considers that an _opaque_ function call propagates any taint passed through any of its arguments to its output.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-```
+This can of course generate false positives, and for certain rules on certain codebases it can produce a high amount of noise.
 
-If this is too cumbersome, then set Semgrep to assume that functions are safe by default. This is done by setting `taint_assume_safe_functions: true` under the rule [`options`](/writing-rules/rule-syntax/#options) key. If you use this approach, you have to enumerate your taint propagators:
+Setting `taint_assume_safe_functions: true` makes Semgrep assume that opaque function calls are safe and do not propagate any taint. If it is desired that specific functions do propagate taint, then that can be achieved via custom propagators:
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=gBD0" border="0" frameBorder="0" width="100%" height="432"></iframe>
 
-Similarly, by default, Semgrep assumes that indexing an array with a tainted index (that is, `array[tainted]`) is a tainted expression, even if the array itself is not tainted. Set `taint_assume_safe_indexes: true` to instead assume that these expressions are safe.
+### Propagate only via assignments 🧪
```
s-santillan, line 614:
```suggestion
### Propagate only through assignments 🧪
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
 The following example demonstrates the use of source and sink metavariable unification:
 
 <iframe src="https://semgrep.dev/embed/editor?snippet=G652" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+## Inter-procedural analysis (Pro)
```
s-santillan, line 634:
```suggestion
## Inter-procedural analysis (Pro)
```
Should we call this "cross-function"?

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+
+**Inter-procedural taint analysis is a Semgrep Pro feature.**
+
+[Semgrep Pro](/semgrep-pro-vs-oss/) can perform inter-procedural taint analysis, that is, to track taint across multiple functions.
+
+In the example below, `user_input` is passed to `foo` as input and, from there, it flows to the sink at line 3, trough a call-chain involving three functions. Semgrep is able to track this and report the sink as tainted. Semgrep also provides an inter-procedural taint trace that explains how exactly `user_input` reaches the `sink(z)` statement (click "Open in Playground", then click on "dataflow" in the "Matches" panel).
```
s-santillan, line 640:
```suggestion
In the example below, `user_input` is passed to `foo` as input and, from there, flows to the sink at line 3, through a call chain involving three functions. Semgrep is able to track this and report the sink as tainted. Semgrep also provides an inter-procedural taint trace that explains how exactly `user_input` reaches the `sink(z)` statement (click "Open in Playground", then click on "dataflow" in the "Matches" panel).
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+Using the CLI option `--pro-intrafile`, Semgrep will perform inter-procedural (across functions) _intra_-file (within one file) analysis. That is, it will track taint across functions, but it will not cross file boundaries. This is supported for essentially every language, and performance is very close to that of intra-procedural taint analysis.
+
+Using the CLI option `--pro`, Semgrep will perform inter-procedural (across functions) as well as *inter*-file (across files) analysis. Inter-file analysis is only supported for [a subset of languages](/supported-languages/#semgrep-code-language-support). For a rule to run inter-file it also needs to set `interfile: true`:
+
+```yaml
+options:
+  interfile: true
+```
+
+**Memory requirements for inter-file analysis:**
```
s-santillan, line 653: Note for myself to add this to the top-level Semgrep Pro Engine docs too.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+While inter-file analysis is more powerful, be aware that it also demands more memory resources. We advise a minimum of 4 GB of memory per core, but _recommend 8 GB per core or more_. The amount of memory needed depends on the codebase and on the number of inter-file rules being run.
```
s-santillan, line 654:
```suggestion
While interfile analysis is more powerful, it also demands more memory resources. The Semgrep team advises a minimum of 4 GB of memory per core, but **recommend 8 GB per core or more**. The amount of memory needed depends on the codebase and on the number of interfile rules being run.
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+
+## How sensitive is taint-mode ?
```
s-santillan, line 656:
```suggestion
## Taint mode sensitivity
```
For consistency among headings.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+
+### Field sensitivity
+
+The taint engine provides basic field sensitivity support, it can:
```
s-santillan, line 660:
```suggestion
The taint engine provides basic field sensitivity support. It can:
```

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+**Index sensitivity is a Semgrep Pro feature.**
+
+Semgrep Pro has basic index sensitivity support:
+- Only for accesses using the built-in `a[E]` syntax.
+- Works for _statically constant_ indexes that may be either integers (e.g. `a[42]`) or strings (e.g. `a["foo"]`).
+- If an arbitrary index `a[i]` is sanitized, then every index becomes clean of taint.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=GdoK6" border="0" frameBorder="0" width="100%" height="432"></iframe>
+
+## Taint labels (Pro) 🧪
```
s-santillan, line 682: This was originally its own document, right? Why did we move it out?

s-santillan (commented):
Thanks for this @IagoAbal! Most of the notes should be easy to resolve; I had some about the layout (such as why we removed taint labels as its own doc?) LMK if you have any concerns.

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+
+Taint labels increase the expressiveness of taint analysis by allowing you to specify and track different kinds of tainted data in one rule using labels. This functionality has various uses, for example, when data becomes dangerous in several steps that are hard to specify through single pair of source and sink.
+
+<iframe class="yt_embed" width="100%" height="432px" src="https://www.youtube.com/embed/lAbJdzMUR4k" frameborder="0" allowfullscreen></iframe>
```
s-santillan, line 686: kyle
 [2 hours ago](https://semgrepinc.slack.com/archives/C02JT6YA5S7/p1711423039363669)
Yello!  I think there is an error on this page:
https://semgrep.dev/docs/writing-rules/experiments/taint-labels/
I believe the test code should look like:
def bad(user_input):
    x = user_input
    # ok: taint-labels (no evil!)
    sink(x)
rather than the current code, which shows sink(y) :slightly_smiling_face:

observation from @jkylekelly

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: The idea seems correct although the example code not be the best one. If you change `addy(y, x)` to `addy(ys, x)`, `y = getUserInputNumber()` to `ys.add(getUserInputNumber())`, and `x = x + y` to `x = x + ys.pop()`, then yeah it's clearly correct. I'm no Python expert but I'm just stressing that `y` would be a mutable object that is being mutated inside `addy`. Numbers in principle are not mutable in Python, I think, and if you `y = getUserInputNumber()` then you're just replacing the value of `y` without altering what has been passed as argument.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: An `l-value` essentially means "an expression that denotes an object in memory", a memory location, something that you can use in the LHS of an assignment. Maybe we should put the definition in the text too.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Nope, clicking on the list of matches (on the right bottom), then click "dataflow" for the match in question.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Whoops, I think no. I guess I thought about writing something more but then decided not to, but forgot to remove that.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: What do you mean bye "magic"?

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: I'd prefer we just add a clarification in the text like "also referred as cross-function" or something. I see "cross-function" as a more marketing friendly term targeting those who will actually buy, whereas I see these docs as oriented towards rule writers or those who will actually use the tool. If you google both terms you see that "interprocedural analysis" gives you very relevant results, whereas "cross-function analysis" gives you mostly unrelated results.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Prefer to have everything under one page, for better visibility. Also, despite taint labels are kinda "unstable", we use them a lot ourselves and they're not going away.

docs/writing-rules/data-flow/taint-mode.md
IagoAbal: Is this related to Pieter's video?

docs/writing-rules/data-flow/taint-mode.md
s-santillan: Sorry, my question about magic was not directly related to opacity, but more of "what does Semgrep understand" wrt to stdlib. For example, a beginner rule-writer would not be able to infer from just reading the rule that Semgrep is able to discern that `endsWith()` returns a boolean. It's kinda subtle, but knowing that Semgrep can do that alters the way we'd write a Java rule, because we can take Semgrep's understanding of standard library functions into consideration. 

Ref: https://en.wikipedia.org/wiki/Magic_(programming)

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -306,3 +630,104 @@ Semgrep can also treat metavariables with the same name as the _same_ metavariab
...
+### Index sensitivity (Pro)
+
+**Index sensitivity is a Semgrep Pro feature.**
+
+Semgrep Pro has basic index sensitivity support:
+- Only for accesses using the built-in `a[E]` syntax.
+- Works for _statically constant_ indexes that may be either integers (e.g. `a[42]`) or strings (e.g. `a["foo"]`).
+- If an arbitrary index `a[i]` is sanitized, then every index becomes clean of taint.
+
+<iframe src="https://semgrep.dev/embed/editor?snippet=GdoK6" border="0" frameBorder="0" width="100%" height="432"></iframe>
```
s-santillan, line 680: Sorry @IagoAbal it was meant to go here:

kyle
2 hours ago
Yello! I think there is an error on this page:
https://semgrep.dev/docs/writing-rules/experiments/taint-labels/
I believe the test code should look like:

def bad(user_input):
    x = user_input
    # ok: taint-labels (no evil!)
    sink(x)


rather than the current code, which shows `sink(y)` 🙂

observation from @jkylekelly

docs/writing-rules/data-flow/taint-mode.md
s-santillan: Found the right line. Resolving this one.

docs/writing-rules/data-flow/taint-mode.md
s-santillan: I've added this definition to https://github.com/semgrep/semgrep-docs/pull/1155

docs/writing-rules/data-flow/taint-mode.md
```diff
@@ -182,118 +500,124 @@ pattern-propagators:
...
-2. The `to` and `from` metavariables. These metavariables should match a **variable** or an **expression**.
-    - The `from` metavariable specifies the entry point of the taint.
-    - The `to` metavariable specifies where the tainted data is propagated to, typically an object or data structure.
+### Propagation without side-effect
 
-:::note
-When the `to` metavariable matches a variable (or in general, an l-value), the taint is propagated _by side-effect_.
-See [Sources by side effect](#sources-by-side-effect) for an explanation of what this means.
-:::
+Taint propagators can be used in very imaginative ways, and in some cases you may not want taint to propagate by side-effect. This can be achieved by disabling `by-side-effect`, which is enabled be default.
```
s-santillan, line 505:
```suggestion
Taint propagators can be used in very imaginative ways, and in some cases you may not want taint to propagate by side-effect. This can be achieved by disabling `by-side-effect`, which is enabled by default.
```

s-santillan (approved):
Looking good! Thank you.

docs/writing-rules/data-flow/taint-mode.md
s-santillan: Got it. Yeah, it's hard to think of examples (at least for me) sometimes.

s-santillan:
Will take the liberty of merging this. Iago is OOO, and I think the docs are in a good state for merging. We can make subsequent follow-up PRs, most of the notes are small.

PR #1479, opened 2024-04-10
add info to distinguish between fixed and removed findings (reliable fixed status)

khorne3:

### Please ensure


s-santillan (approved):
LGTM!

vivekkhimani:
Thanks for updating the docs. I can leave a comment on the PR too but I don’t think our “removed” status is ever exposed to the user. It is something that we use to “archive” the findings instead of permanently deleting them from the database. I am unsure though if users can get “removed” issues via our public API. That might be a good question for @daniel-bryant !

Except that, the fixed part looks great.

khorne3:
@daniel-bryant do you know if users would be able to obtain information regarding removed findings from the API? If so, we'll want to add a note regarding this

daniel-bryant:
Hey @khorne3. Hmm okay it's kind of interesting. So we [silently filter out `removed` findings from the list endpoint](https://github.com/semgrep/semgrep-app/blob/develop/server/semgrep_app/core_exp/findings/handlers/issue.py#L521), which means theoretically customers would never have `removed` findings returned. But we do list `removed` as a possible `state` https://semgrep.dev/api/v1/docs/#tag/Finding/operation/semgrep_app.core_exp.findings.handlers.issue.openapi_list_recent_issues.

PR #1498, opened 2024-04-23
remove beta flags from Secrets

khorne3:


docs/usage-and-billing.mdx
```diff
@@ -38,7 +38,7 @@ All Semgrep Pro products are free under the **Team** tier for **10 monthly contr
 - Semgrep Cloud Platform
 - Semgrep Code
 - Semgrep Supply Chain
-- Semgrep Secrets (beta)
+- Semgrep Secrets
```
s-santillan, line 41:
```suggestion
```
Let's remove the entire line for now while we're in this weird in-between state!

s-santillan (changes requested):
:pray:  While it isn't self-serve, let's keep Secrets off the "free" callout box

s-santillan (approved):

PR #1501, opened 2024-04-24
Sara/cherry pick trailingslash false

s-santillan:

### Please ensure

- [ ] Pablo reviews the content or PR

I went through the cherry pick commits and resolved 'em to the best of my cognition, whatever state it was in.


s-santillan:
Converting to draft for now because I see some easy items to fix.

pabloest (approved):
Thank you for this change, @s-santillan! Overall it looks good to me. There is one oddity to point out that doesn’t seem to be a concern but is worth mentioning and monitoring.

The redirects won’t all work when testing locally (using `build && serve`). Specifically the redirects containing a `from:` value that includes a trailing slash will cause a redirect to the root `/docs/` directory. The probably because the local server behaves differently than Netlify. 

However I merged your branch into `develop` and pushed `develop` onto the Netlify staging server. I tested the redirects and they are working as expected. You can see an example here: [https://semgrep-docs-staging.netlify.app/docs/semgrep-ci/overview/](https://semgrep-docs-staging.netlify.app/docs/semgrep-ci/overview/).

I expect the Netlify production server to behave in the same way. If for some reason it doesn’t, then we can use [Netlify’s own redirect system](https://docs.netlify.com/routing/redirects/).

pabloest:
This merged and pushed to production. Doing some spot checking now and everything seems good so far 🎉

PR #1521, opened 2024-04-30
Update Semgrep Editor docs + add information re: structure mode

khorne3:

### Please ensure


src/components/procedure/_enable-turbo-mode.md
```diff
@@ -1,9 +1,9 @@
-:::tip Try Turbo mode (beta)
-Semgrep can automatically scan for findings using the Semgrep OSS Engine, on any change in the Rule or Sample Code panes.
+The **Turbo** option runs your rule automatically against the Semgrep OSS Engine after every keystroke or change to the rule.
 
-Click the <i class="fa-solid fa-toggle-large-on"></i> **Turbo mode** toggle to try it out.
+Click the <i class="fa-solid fa-toggle-large-on"></i> **Turbo mode** toggle to enable.
 
-In this mode, the **Run** button is hidden because Semgrep scans upon detecting any change to the text. Turbo mode works in both Simple mode and Advanced mode. Turbo mode cannot be enabled for Semgrep Pro Engine.
+The **Run** button is hidden in Turbo mode because Semgrep scans after any changes to the rule. Turbo mode works with the editor in either simple or advanced mode.
```
brandonspark, line 5:
```suggestion
The **Run** button is hidden in Turbo mode because Semgrep scans after any changes to the rule. Turbo mode works with the editor in either structure or advanced mode.
```

brandonspark: (simple mode won't actually exist anymore)

brandonspark (approved):
looks good! it's worth noting that these screenshots are very subtly out of date (after I merged https://github.com/semgrep/semgrep-app/pull/13519), because `simple` mode will no longer exist, `structure` takes its place. but, it probably doesn't matter that this is 100% faithful. its great we're deprecating the really old playground screenshots :)

s-santillan:
Howdy!
1. Requesting that Pablo review these docs.
2. I think we'll want to remove the playground markdown file? Did we miss any information in the Playground that isn't in the Editor?

khorne3:
@s-santillan Sounds good!

I don't think there's much difference in terms of the two, so I think we can delete playground and redirect?

pabloest:
@khorne3 Can you share some of the reasoning behind removing the Playground file? AFAICT structure mode is available in the Playground. Are we trying to instruct users to instead go to Semgrep Code > Editor for structure mode?

khorne3:
@pabloest The playground doc has been hidden for some time now. I've moved info from the Playground doc to the Editor doc if necessary, and since people aren't supposed to see the former, I deleted it and added a redirect. Additionally, they're both so similar that one doc should be sufficient and make it easier to maintain

pabloest:
OK great! Thanks for the context, @khorne3.

pabloest (approved):

PR #1534, opened 2024-05-06
update credit card payment information

khorne3:

### Please ensure

### DIFF
--- a/docs/usage-and-billing.mdx
+++ b/docs/usage-and-billing.mdx
@@ -149,7 +149,8 @@ Pay through the following methods:
 
 <dl>
   <dt>Pay using your credit card.</dt>
-  <dd>The payment will be processed through Stripe.</dd>
+  <dd>The payment will be processed through Stripe. If you need to change the credit card on file, please reach out to{" "}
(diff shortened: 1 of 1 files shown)


docs/usage-and-billing.mdx
```diff
@@ -149,7 +149,8 @@ Pay through the following methods:
 
 <dl>
   <dt>Pay using your credit card.</dt>
-  <dd>The payment will be processed through Stripe.</dd>
+  <dd>The payment will be processed through Stripe. If you need to change the credit card on file, please reach out to{" "}
```
s-santillan, line 152: What's the {" "} for, just curious?

s-santillan (approved):

docs/usage-and-billing.mdx
khorne3: I thought it was necessary for formatting, but I just tried without and it doesn't seem to be the case.

khorne3: Never mind, it was; it seems to keep the text and the link spaced properly if there is a line break

PR #1537, opened 2024-05-07
SBOM limitation: latest full scan on default branch

armchairlinguist:
Currently, SBOM vulns always come from the default branch and dependency lists from the latest full scan - so those need to match.

### Please ensure

### DIFF
--- a/docs/semgrep-supply-chain/sbom.md
+++ b/docs/semgrep-supply-chain/sbom.md
@@ -23,7 +23,7 @@ Semgrep Supply Chain supports the following:
 
 :::info Prerequisites
 - **SBOM generation** can be performed only through Semgrep AppSec Platform.
-- You need at least one successful Supply Chain scan on the **trunk** branch of each repository you want to generate an SBOM for. See [<i class="fa-regular fa-file-lines"></i> Core deployment](/deployment/core-deployment) to set up your Semgrep account and Supply Chain scans.
+- The latest successful full scan running Supply Chain must have run on the **trunk** (default) branch of each repository you want to generate an SBOM for, or there can be a mismatch between dependencies and vulnerabilities. See [<i class="fa-regular fa-file-lines"></i> Core deployment](/deployment/core-deployment) to set up your Semgrep account and Supply Chain scans.
 :::
 
 1. In Semgrep AppSec Platform, click **Supply Chain > Dependencies**.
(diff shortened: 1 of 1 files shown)


docs/semgrep-supply-chain/sbom.md
```diff
@@ -23,7 +23,7 @@ Semgrep Supply Chain supports the following:
 
 :::info Prerequisites
 - **SBOM generation** can be performed only through Semgrep AppSec Platform.
-- You need at least one successful Supply Chain scan on the **trunk** branch of each repository you want to generate an SBOM for. See [<i class="fa-regular fa-file-lines"></i> Core deployment](/deployment/core-deployment) to set up your Semgrep account and Supply Chain scans.
+- The latest successful full scan running Supply Chain must have run on the **trunk** (default) branch of each repository you want to generate an SBOM for, or there can be a mismatch between dependencies and vulnerabilities. See [<i class="fa-regular fa-file-lines"></i> Core deployment](/deployment/core-deployment) to set up your Semgrep account and Supply Chain scans.
```
khorne3, line 26: Not sure if I got the details correct, but I wanted to reword a bit since the first sentence has a few criteria that must be met; I'm also not sure what the specifics of the mismatch would be. WDYT of something like this?

> You need at least one successful Supply Chain scan on the **trunk** (default) branch of each repository you want to generate an SBOM for, and the most recent scan must be a full scan. Otherwise, there can be issues associating the vulnerabilities with the correct dependencies.

armchairlinguist: No, that's not quite right unfortunately. The most recent full scan for the repo must be on the default branch. Otherwise, when we create the SBOM, the vulnerabilities are drawn from the default branch, but the dependencies will be drawn from the latest full scan, whatever branch that is. So it has to be the default for the two to match.

armchairlinguist: if we need to pull this out of the callout bc it's more complex, totally open to that!

khorne3: Yeah, I think it would be helpful to pull it out into its own section and then list the reqs out with bullet points. WDYT?

armchairlinguist: I'll give it a shot!

gautambhat (approved):
Your PR description says the opposite, "Currently, SBOM dependency lists always come from the default branch and vulns from the latest full scan - so those need to match.", but the actual docs look good to me!

armchairlinguist:
@gautambhat whoops, good catch! fixed for posterity, glad I got it right in the doc

PR #1538, opened 2024-05-07
Fix reverted change to focus on Code, add note about confidence and limited runs

armchairlinguist:
Redux of correcting the over-general compatibility (probably reverted accidentally) + some other updates for clarification on confidence and per-scan max. Kept it a little vague because otherwise the max is likely to get out of date.

### Please ensure


armchairlinguist:
It looks like the link workflow failed because https://openai.com/policies/api-data-usage-policies redirects to https://openai.com/enterprise-privacy now. I don't know what the page looked like before, but the new page looks hard to navigate and pretty general, so I'm not entirely sure what's best to do here.

docs/semgrep-assistant/overview.md
```diff
@@ -23,7 +23,7 @@ Semgrep Assistant provides GPT-4-powered security recommendations to help you re
 
 ## Support and availability
 
-Semgrep Assistant supports findings generated by Semgrep Code, Semgrep Supply Chain, and Semgrep Secrets. It requires Semgrep AppSec Platform and is not available on the Semgrep CLI.
+Semgrep Assistant primarily supports findings generated by Semgrep Code. It requires Semgrep AppSec Platform.
```
Sly1029, line 26:
```suggestion
Semgrep Assistant primarily supports findings generated by Semgrep Code. It requires the Semgrep AppSec Platform.
```

Sly1029 (approved):

docs/semgrep-assistant/overview.md
```diff
@@ -69,7 +69,7 @@ Assistant's suggestions to ignore findings are also surfaced in PR or MR comment
 
 Semgrep Assistant can suggest [autofix](/writing-rules/autofix/) code snippets for Semgrep Code findings when it identifies a true positive. Assistant only suggests an autofix if the rule doesn't have a human-written autofix. Assistant customizes the code snippets it provides based on previous feedback, if any, and your rule customizations. For example, if you have a custom rule that recommends a specific sanitizer, Assistant can recommend its use in the autofix suggestion for the issue in your code.
 
-You can set the minimum autofix confidence level required to display autofix suggestions from Semgrep Assistant in Semgrep AppSec Platform's **Settings** page.
+You can set the minimum autofix confidence level required to display autofix suggestions from Semgrep Assistant in Semgrep AppSec Platform's **Settings** page. To receive as many Assistant suggestions as are available, set the minimum to low confidence.
```
khorne3, line 72:
```suggestion
You can set the minimum autofix confidence level required to display autofix suggestions from Semgrep Assistant in Semgrep AppSec Platform's **Settings** page. To receive as many Assistant suggestions as are available, set the minimum to **low confidence**.
```

docs/semgrep-assistant/overview.md
```diff
@@ -83,6 +83,10 @@ Autofixes are also available in Semgrep AppSec Platform's **Findings page** unde
 
 The finding's details also include a link to the PR or MR with the autofix, so you can go directly to the PR or MR to commit the autofix.
 
+:::info 
+If there are a large number of new issues found in a given scan, Assistant auto-triage and autofix may not run on every issue.
```
khorne3, line 87: This is just my curiosity: if this were to happen, what would be the next steps for the user? Manually triggering auto-triage and autofix?

khorne3 (approved):

khorne3:
@armchairlinguist I think we should update the link to where the redirect goes; I dug through OpenAI's docs and they seem to be funneling all requests for  API data usage policies to that page specifically. WDYT?

docs/semgrep-assistant/overview.md
armchairlinguist: @Sly1029 might need to confirm, but I'm not exactly sure. They wouldn't get the PR comments AFAIK if it was a PR scan (which is where this most often applies).

armchairlinguist:
@khorne3 Makes sense. It's unfortunate that it's a bit hard to navigate, but if that's where they're sending people then we'll send them too.

docs/semgrep-assistant/overview.md
armchairlinguist: @khorne3 Do you think we should add a CTA of sorts here? I'm open to that if there is a good action item.

armchairlinguist:
Also, it seems like maybe OpenAI is refusing permission to bots (how rude of them), so the check is still failing. :(

docs/semgrep-assistant/overview.md
khorne3: I think it is fine for now; we can change if it becomes a problem (or if we think it will)

PR #1548, opened 2024-05-14
Focus more on triggering event which is the key thing

armchairlinguist:

### Please ensure


s-santillan (approved):

s-santillan:
We honestly do not know if the original case was a runner using a webhook, or just plain lost in the telephone game so any investigation is appreciated!

PR #1550, opened 2024-05-15
add morehelp to docitem/content

s-santillan:

### Please ensure

### Previously, on semgrep-docs...

- We had to import `<MoreHelp />` for every doc that we made.
- We had to put the `<ul>` to render tags

### Now, on semgrep-docs...

- `<MoreHelp />` is part of the document template and is baked into every document automatically.
- So that it doesn't get repeated, I have removed all the imports of it.
- Same with the opening tag `<ul>`

### How to test

- Pull this PR and try out some of the pages. It's programmatic find-and-replace so if some work they should all work.


s-santillan:
FYI, the dead link is due to a link from `extensions/overview`, which is pulled from an external repo. We won't address it for now (nor will we run an ignore directive, because it will be overwritten from the source).

khorne3 (approved):
LGTM!

PR #1551, opened 2024-05-15
[KB] Add monorepo example config

sinat101:

### Please ensure


docs/kb/semgrep-ci/scan-monorepo-in-parts.md
```diff
@@ -21,24 +21,24 @@ When scanning a repo with Semgrep in CI, the base command is `semgrep ci`. To un
...
-    /src/moduleA
-    /src/moduleB
-    /src/moduleC
-    /src/moduleD
+    src/moduleA
+    src/moduleB
+    src/moduleC
+    src/moduleD
 
 Then splitting its scans into four separate scans, one for each module, would provide a logical separation for findings. In general, we recommend modules to not exceed ~100,000 lines of code in order to keep optimal scan time and efficiency.
```
armchairlinguist, line 31:
```suggestion
Then splitting its scans into four separate scans, one for each module, would provide a logical separation for findings. In general, we recommend that modules not exceed ~100,000 lines of code in order to maintain optimal scan time and efficiency.
```

docs/kb/semgrep-ci/scan-monorepo-in-parts.md
```diff
@@ -21,24 +21,24 @@ When scanning a repo with Semgrep in CI, the base command is `semgrep ci`. To un
...
 
 After choosing a logical split, use the `--include` flag ([see CLI reference](/docs/cli-reference)) with the relevant path to only scan files in that module's code path:
 
-    semgrep ci --include=/src/moduleA/*
+    semgrep ci --include=src/moduleA/**
```
armchairlinguist, line 35: Are the ** actually necessary? My understanding of our includes is that we don't really need any wildcards because of how we parse them. To test this just now, I ran a scan on juice-shop with `--include=frontend` and it scanned all the files underneath.

armchairlinguist (approved):
I'd recommend jettisoning the **, but other than that it's fine, and they don't break anything, so re-approving.

docs/kb/semgrep-ci/scan-monorepo-in-parts.md
sinat101: to my understanding * means all files but won't go deeper than the current level. ** means all files and folders underneath and including the current path level

armchairlinguist: That's normally true, but I don't believe it applies to the parser that we use - I dug into this a while back, and I also checked just now with the logs from the scan I tried.

sinat101: Based on [this convo](https://semgrepinc.slack.com/archives/C048LGPK46L/p1716418818578379?thread_ts=1716414918.816559&cid=C048LGPK46L) I'm inclined to write the include in the proper gitignore spec format in the hopes that it's fixed when we move to osemgrep, so we don't have to come back here and update the KB doc.

armchairlinguist: Fair enough!

PR #1561, opened 2024-05-20
update SBOM sidebar location

khorne3:
- Moves SBOM out of Reporting and into SCA (Supply Chain) in the sidebar. 
- Adds a link on the Dashboard reports page to SBOM doc.

### DIFF
--- a/docs/semgrep-appsec-platform/dashboard.md
+++ b/docs/semgrep-appsec-platform/dashboard.md
@@ -98,6 +98,7 @@ Filters mentioned in previous sections [Filtering findings by time](#filtering-f
 
 ## See also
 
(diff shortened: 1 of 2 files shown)


armchairlinguist:
I checked this on the preview and something a bit odd is going on there. The SBOM link has moved, but when I click "Reports" to review that, the "Ticketing" part of the sidebar expands, and Reports doesn't.

I also would suggest we might put SBOM after dependency search, since it uses both vulnerability and dependency features.

s-santillan (approved):

khorne3:
@armchairlinguist I've fixed both things, thank you!

PR #1567, opened 2024-05-22
update carets

s-santillan:

### Please ensure

- [x] You review the content

### This PR...

1. Makes the carets smaller, they seem huge.
2. Adds carets to the top-level categories in the sidebar so that people know they are clickable sections. See the [original request](https://semgrepinc.slack.com/archives/C02JT6YA5S7/p1715756250788169) to understand the origins of this bug.

### Why am I tagged for this random PR?

Because I value your opinion. 

### What do I do?

Check the Netlify deploy preview. **Don't forget to append `/docs` at the end of it!** We all have different 2nd monitors, browser default font sizes, laptop screens, etc, so a variety would be good.


khorne3 (approved):
Much improved!

pabloest:
I don’t see the carets on mobile:

CleanShot 2024-05-22 at 18 55 36

Also I think we might want to rethink the phrase/label “Back to main menu” as I think it may be commonly understood to be about the main docs menu, not the main Semgrep menu. But that can be a separate thing.

adamboros:
> I don’t see the carets on mobile:
> 
This only happens on the root level for some reason, one level down the carets are there: 
image

> Also I think we might want to rethink the phrase/label “Back to main menu” as I think it may be commonly understood to be about the main docs menu, not the main Semgrep menu. But that can be a separate thing.

Maybe a simple 'Back' button would be enough. 

The new caret size looks right to me! ✨

s-santillan:
Fixed the missing carets. Almost there: 

image

pabloest (approved):

PR #1574, opened 2024-05-30
fix formatting in diff-aware scan snippet

khorne3:
Removed a bit of extra whitespace:

image


armchairlinguist (approved):
Perfect! I think this is a bit better than I did in my attempt :)

PR #1583, opened 2024-06-07
Callout the requirement for tenants to be pre-enabled

armchairlinguist:
Authentication from tenant requires pre-enablement, make sure to call that out as a pre-requisite.

### Please ensure


armchairlinguist:
This was taken care of through another process.

PR #1589, opened 2024-06-11
Reflect change of toggle name for SCA blocking

armchairlinguist:
We updated the name of the blocking toggle since it also causes full scans to exit with code 1. I changed the name and lightly modified the verbiage - since we still generally focus on PR/MRs for blocking, I didn't remove that aspect of the content.

### Please ensure

### DIFF
--- a/docs/semgrep-supply-chain/getting-started.md
+++ b/docs/semgrep-supply-chain/getting-started.md
@@ -111,7 +111,7 @@ Semgrep Supply Chain supports the scanning of monorepos. As outlined in [Project
 
 ## Block pull or merge requests
 
-Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge request scans if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on scans that detect reachable findings in direct dependencies with high or critical severity.
+Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge request scans if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on scans that detect reachable findings in direct dependencies with high or critical severity. Findings only block when there is an upgrade available, to prevent blocking when there is no resolution for the vulnerability.
 
 1. Log in to Semgrep AppSec Platform.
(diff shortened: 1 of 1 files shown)


docs/semgrep-supply-chain/getting-started.md
```diff
@@ -111,11 +111,11 @@ Semgrep Supply Chain supports the scanning of monorepos. As outlined in [Project
 
 ## Block pull or merge requests
 
-Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on pull request scans that detect reachable findings in direct dependencies with high or critical severity.
+Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge request scans if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on scans that detect reachable findings in direct dependencies with high or critical severity.
```
khorne3, line 114:
```suggestion
Semgrep Supply Chain versions **v0.122.0** and earlier automatically blocked pull/merge requests and scans if it discovered reachable findings in the code, but later versions do not do this. You can, however, configure Semgrep Supply Chain to block on scans that detect reachable findings in direct dependencies with high or critical severity and where there is an upgrade available.
```

I don't know if the first change is helpful or not (I think it is since it blocks both?), but I think the [second caveat is important to know](https://github.com/semgrep/semgrep-app/pull/14645).

armchairlinguist: So for the first bit - I didn't necessarily want to fully change the topic of this over to scan blocking, since the section is called "Block pull or merge requests". I'm open to whether we want to change that title too, but since we still generally focus on PR/MRs for blocking, I hadn't done it myself. What do you think? Should we change it entirely?

For the second bit, absolutely! Not sure why it wasn't mentioned here, I think this even came up when I talked to Swati (who changed the toggle name) about this and we thought it could be useful to add. Good callout!

khorne3 (approved):

docs/semgrep-supply-chain/getting-started.md
armchairlinguist: Resolved in Slack :)

PR #1601, opened 2024-06-20
Update both webhooks with current keys/format

armchairlinguist:
Our webhook samples for both scans and findings were out of date.

I believe I've anonymized/fakeified potentially sensitive data, so this change should not have security implications, but feel free to request another review if you disagree!

### Please ensure


s-santillan (approved):
LGTM!
Also checked on my end, always good to have 2 sets of eyes, looks consistent with my testing as well. Thank you.

PR #1607, opened 2024-06-25
update content on constant propagation

khorne3:
* minor edits to the doc
* added info on constant propagation that would be helpful for troubleshooting

### Please ensure

### DIFF
--- a/docs/writing-rules/data-flow/constant-propagation.md
+++ b/docs/writing-rules/data-flow/constant-propagation.md
@@ -81,10 +81,14 @@ public class Test2 {
 }
 ```
 
-Because `REGEX` is public in the first code snippet, Semgrep doesn't propagate its value to other classes on the assumption that it could have mutated. However, in the second example, `REGEX` is private, and Semgrep assumes it to be immutable. The rule would also work with:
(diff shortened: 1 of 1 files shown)


docs/writing-rules/data-flow/constant-propagation.md
```diff
@@ -7,37 +7,88 @@ tags:
...
+private String REGEX = "(a+)+$";
+
+public class Test2 {
+    public static void main(String[] args) {
+        Pattern pattern = Pattern.compile(REGEX);
+    }
+}
+```
+
+Because `REGEX` is public in the first code snippet, Semgrep doesn't propagate its value to other classes on the assumption that it could have mutated. However, in the second example, `REGEX` is private, and Semgrep assumes it to be immutable. The rule would also work with:
```
brandonspark, line 84: i think in this case, it's actually a two step process. semgrep sees that it's `private`, _and also_ knows its only assigned to once, and so assumes it to be immutable. this code doesn't work if you add another assignment into the target:
```
import java.util.regex.Pattern;

private String REGEX = "(a+)+$";
REGEX = "(a+)+$";

public class Test2 {
    public static void main(String[] args) {
        Pattern pattern = Pattern.compile(REGEX);
    }
}
```
however, if it were `final`, then it would work always

brandonspark (approved):
just one comment!

PR #1609, opened 2024-06-25
Update Semgrep Pro rules documentation

s-santillan:

### Please ensure

### The what and why of this PR

It looks like this doc has not been updated in a while. The TLDR is that this update cleans up the doc and removes implementation details.

- It seems to assume that Pro rules are not included in the Default ruleset and have to be manually added (whereas now it is automatically available for all new accounts).
- It uses the term "removing rulesets", when our preferred operation is "disabling rulesets", and disabled rulesets are (now) under **Rule modes**.
- It provides several ways of doing something (i.e. disabling from various places), when we should be recommending a best path or path of least resistance.
- And a section about snippets was cut off, which was the original bug.
- In the original doc it classifies rules with the section "Kinds of rules", which is basically "types of rules" -- but doesn't explain by what facet? Rule by language? Rule by vuln class? I clarify that the types here refer to authorship.
- Removed the rule updates link since SR no longer updates that document.
- Removed the bit about losing access to Semgrep Pro rules, because these rules are provided by default. If a user has "lost access" to these rules, they have already lost access to run scans at all from Semgrep AppSec Platform (i.e. they are over 10 and not subscribed).
- The Filtering behavior header was moved to the Policies page (it is the primary source of info for Registry) and Pro rules links to it instead.


khorne3 (approved):
Just some very minor things!

docs/semgrep-code/pro-rules.md
```diff
@@ -8,95 +8,67 @@ tags:
...
 
+The goal of Pro rules is to provide a set of well-supported rules with improved coverage across languages and vulnerability types. Semgrep Pro rules are written using Semgrep’s latest features and, in general, target users who are looking to produce highly accurate, actionable findings.
 
-## Kinds of rules in the Semgrep Registry
+## Types of rules in the Semgrep Registry by author
 
-The Semgrep Registry includes the following kinds of rules:
 * **Community rules** - reviewed by the Semgrep team, these rules consist of contributions from Semgrep’s community. Community rules encompass a wide array of rules, including many that are made for security auditors.
 * **Third-party rules** - created directly by external contributors such as Trail of Bits, GitLab and many more.
 * **Private rules** - rules that can be authored and published by your own organization and for use only by your organization.
```
armchairlinguist, line 24:
```suggestion
* **Private rules** - rules authored and published by your own organization, for use only by your organization.

```

docs/semgrep-code/pro-rules.md
```diff
@@ -8,95 +8,67 @@ tags:
...
-
-The goal of Pro rules is to provide a set of well-supported rules with improved coverage across languages and vulnerability types. Semgrep Pro rules are written using Semgrep’s latest features and, in general, target users who are looking to produce highly accurate, actionable findings.
-
 ## Semgrep Pro rules content
 
 Semgrep Pro rules provide improved coverage for many languages, including Java, JavaScript, TypeScript, Python, PHP, Ruby, C#, Swift, and Go.
 
-After using Semgrep Pro rules, you will see improved findings across many languages on specific classes of vulnerabilities (such as hard-coded secrets, injection vulnerabilities, deserialization, XXE, and many others) as well as increased support for frameworks and technologies (such as Express, Spring, Java Servlets, Laravel, Go net/http, React, Next.js, Angular).
+Semgrep Pro rules provide improved findings across many languages on specific classes of vulnerabilities, such as hard-coded secrets, injection vulnerabilities, deserialization, XXE, and many others, as well as increased support for frameworks and technologies such as Express, Spring, Java Servlets, Laravel, Go net/http, React, Next.js, and Angular.
```
armchairlinguist, line 31: I'm not sure if we should include secrets here, since the Secrets product rules are not visible in the Registry, and those are where we're putting effort.

docs/semgrep-code/pro-rules.md
```diff
@@ -8,95 +8,67 @@ tags:
...
-:::
-
-## Using Semgrep Pro rules
-
-You can use Semgrep Pro rules in both Semgrep AppSec Platform and your own CLI or CI environment.
+## Scan with Semgrep Pro rules
 
-* To use Pro Rules with Semgrep AppSec Platform, add them to your Policies page. The rules are then included in your next scan.
-* To use Pro Rules in your own CI or CLI environment, add the rules or rulesets through the `--config` flag.
+Your Semgrep AppSec Platform account already includes Pro rules, as they are included in the **Default** ruleset. These Pro rules run on all your scans.
```
armchairlinguist, line 37:
```suggestion
Your Semgrep AppSec Platform account already includes Pro rules that are likely to be widely useful, as they are included in the **Default** ruleset. These Pro rules run on all your scans.

```

armchairlinguist: Is it worth mentioning that for them to run interfile, the interfile analysis toggle needs to be enabled also?

docs/semgrep-code/pro-rules.md
```diff
@@ -8,95 +8,67 @@ tags:
...
-
-1. Go to [Semgrep Registry](https://semgrep.dev/r).
-2. Click on **Visibility > Pro rules**.
-3. Optional: Apply additional filters by entering search terms in the search box or selecting filters from drop-down boxes.
-4. For a single rule, click the rule's card > Run locally. For rulesets, click the card.
-5. Copy and paste the snippet to your CLI or CI configuration file. You can add several rulesets. Refer to the following sample snippet:
+### Change rule modes or disable Pro rules in Semgrep AppSec Platform
 
-### Adding Pro rules to your Semgrep org through the Registry
+Like any other rule or ruleset, you can disable Pro rules or change their rule mode to leave comments to developers or potentially block a PR.
```
armchairlinguist, line 45:
```suggestion
Like any other rule or ruleset, you can disable Pro rules or change their rule mode to leave comments for developers or potentially block a PR.

```

docs/semgrep-code/pro-rules.md
```diff
@@ -8,95 +8,67 @@ tags:
...
-<!-- Future feature: Add Semgrep Pro rules through your Policy -->
+1. Navigate to **Rules > Policies**.
+1. Under **Source**, click **Pro <i class="fa-solid fa-gem"></i>** to view all the Semgrep Pro rules currently in your Policies.
+1. Find and select the rules you want to disable or change.
+1. Click **Change modes** and select one of the provided options.
 
-<!-- Future feature: Receiving updates on Semgrep Pro rules -->
-
-## Removing or disabling Semgrep Pro rules
+You can find all Semgrep Pro rules in your Policies page, so if you want to re-enable Pro rules, you can simply use the **Source > Pro <i class="fa-solid fa-gem"></i>** filter.
```
armchairlinguist, line 53:
```suggestion
You can find all previously-added Semgrep Pro rules in your Policies page, so if you want to re-enable Pro rules or adjust the mode again in the future, use the **Source > Pro <i class="fa-solid fa-gem"></i>** filter as described previously.
```
exact wording here I'm not picky about, but saying "all" seems confusing to me, as it's not all, just all the ones you had originally

docs/semgrep-code/pro-rules.md
```diff
@@ -8,95 +8,67 @@ tags:
...
 
-<RemoveRuleset />
+1. Go to [Semgrep Registry](https://semgrep.dev/r).
+2. Click **Visibility > Pro rules**.
+3. Optional: Apply additional filters by entering search terms in the search box or selecting filters from drop-down boxes.
+4. For a single rule, click the **Rule's card > Run locally**. For rulesets, click the card.
+5. Copy and paste the snippet to your CLI or CI configuration file. You can add several rulesets. Snippets can look like these:
+    ```bash
+    semgrep --config "p/c"
+    semgrep --config "p/xss"
```
armchairlinguist, line 70: I don't think these are good examples here - most of the rulesets you'd add will have a mix of OSS and Pro rules, and I don't think even filtering the Registry to Pro before you browse would change that. I'm not actually sure what a good example would be, unfortunately, but it might be worth checking with the Code folks.

s-santillan: Opted to use a screenshot instead to make sure the user doesn't get lost.

armchairlinguist (approved):

PR #1612, opened 2024-06-26
No active voice for removed, since it's not a user action

armchairlinguist:
A finding changing to a removed status is not something a user can do directly, it just happens, so we shouldn't really use the active voice here.

### Please ensure


khorne3 (approved):
Thank you!

PR #1626, opened 2024-07-03
add pattern syntax new sections

s-santillan:

### Please ensure

### Preview

image

### DIFF
--- a/docs/writing-rules/pattern-syntax.mdx
+++ b/docs/writing-rules/pattern-syntax.mdx
@@ -515,7 +515,15 @@ When referencing an ellipsis metavariable in a rule message or [metavariable-pat
 - message: Call to foo($...ARGS)
 ```
 
-### Displaying matched metavariables in rule messages
+### Anonymous metavariables
+
+tk add content
+
+### Metavariable unification
+
+tk add content
+
+### Display matched metavariables in rule messages
 
 Display values of matched metavariables in rule messages. Add a metavariable to the rule message (for example `Found $X`) and Semgrep replaces it with the value of the detected metavariable.
 
--- a/docs/writing-rules/pattern-syntax.mdx
+++ b/docs/writing-rules/pattern-syntax.mdx
@@ -527,13 +527,23 @@ For example, if you want to specify that a function should **always** have 3 arg
 - pattern: def function($_, $_, $_)
 ```
 
+An anonymous metavariable does not produce any binding to the code it matched. This means it does not store the code that it matched. The pattern:
+
+```yaml
+- pattern: def function($A, $B, $C)
+```
+
+is not equivalent to the former example, as `$A`, `$B`, and `$C` bind to the code that matched the pattern. You can then use `$A` or any other metavariable in your rule definition to specify that specific code. Anonymous metavariables cannot be used this way.
+
 Anonymous metavariables also communicate to the reader that their values are not relevant, but rather their occurrence in the pattern.
 
 ### Metavariable unification
 
-For search mode rules, metavariables with the same name are treated as the same metavariable across and within pattern operators **except for `pattern-either`**.
+For search mode rules, metavariables with the same name are treated as the same metavariable across and within pattern operators **except for `pattern-either`**. This is called metavariable unification.
 
-For taint mode rules, metavariable unification is **not** enabled by default. To enforce unification, set `taint_unify_mvars: true` under the rule `options` key. When `taint_unify_mvars: true` is set, a metavariable defined in `pattern-sinks` and `pattern-sources` with the same name is treated as the same metavariable. See [Metavariables, rule message, and unification](/writing-rules/data-flow/taint-mode#metavariables-rule-message-and-unification) for more information.
+For taint mode rules, patterns defined **within** `pattern-sinks` and `pattern-sources` still unify. However, metavariable unification **between** `pattern-sinks` and `pattern-sources` is **not** enabled by default.
+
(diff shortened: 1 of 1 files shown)


docs/writing-rules/pattern-syntax.mdx
```diff
@@ -515,7 +515,27 @@ When referencing an ellipsis metavariable in a rule message or [metavariable-pat
...
 ```
 
-### Displaying matched metavariables in rule messages
+### Anonymous metavariables
+
+Anonymous metavariables are used to specify that there a metavariable exists in the pattern you want to capture.
+
+An anonymous metavariable always takes the form `$_`. Variables such as `$_1` or `$_2` are **not** anonymous. You can use more than one anonymous metavariable in a rule definition.
+
+For example, if you want to specify that a function should **always** have 3 arguments, then you can use anonymous metavariables:
```
brandonspark, line 524: could we stress why this is different than `function($A, $B, $C)`? id like if we explicitly said that anonymous metavars _do not produce bindings_, and that is why there is no unification that occurs between them

docs/writing-rules/pattern-syntax.mdx
```diff
@@ -515,7 +515,27 @@ When referencing an ellipsis metavariable in a rule message or [metavariable-pat
...
+- pattern: def function($_, $_, $_)
+```
+
+Anonymous metavariables also communicate to the reader that their values are not relevant, but rather their occurrence in the pattern.
+
+### Metavariable unification
+
+For search mode rules, metavariables with the same name are treated as the same metavariable across and within pattern operators **except for `pattern-either`**.
+
+For taint mode rules, metavariable unification is **not** enabled by default. To enforce unification, set `taint_unify_mvars: true` under the rule `options` key. When `taint_unify_mvars: true` is set, a metavariable defined in `pattern-sinks` and `pattern-sources` with the same name is treated as the same metavariable. See [Metavariables, rule message, and unification](/writing-rules/data-flow/taint-mode#metavariables-rule-message-and-unification) for more information.
```
brandonspark, line 536: technically, unification is only not enabled between the source and sink. the patterns within the source will still unify with each other, maybe worth clarifying

brandonspark (approved):
looks good overall! just one concern

s-santillan:
@brandonspark Thank you!!

docs/writing-rules/pattern-syntax.mdx
s-santillan: Here is the addition:
image

docs/writing-rules/pattern-syntax.mdx
```diff
@@ -515,7 +515,37 @@ When referencing an ellipsis metavariable in a rule message or [metavariable-pat
 - message: Call to foo($...ARGS)
 ```
 
-### Displaying matched metavariables in rule messages
+### Anonymous metavariables
+
+Anonymous metavariables are used to specify that there a metavariable exists in the pattern you want to capture.
```
khorne3, line 520:
```suggestion
Anonymous metavariables are used to specify that a metavariable exists in the pattern you want to capture.
```

khorne3 (approved):

docs/writing-rules/pattern-syntax.mdx
```diff
@@ -515,7 +515,37 @@ When referencing an ellipsis metavariable in a rule message or [metavariable-pat
...
+- pattern: def function($A, $B, $C)
+```
+
+is not equivalent to the former example, as `$A`, `$B`, and `$C` bind to the code that matched the pattern. You can then use `$A` or any other metavariable in your rule definition to specify that specific code. Anonymous metavariables cannot be used this way.
+
+Anonymous metavariables also communicate to the reader that their values are not relevant, but rather their occurrence in the pattern.
+
+### Metavariable unification
+
+For search mode rules, metavariables with the same name are treated as the same metavariable across and within pattern operators **except for `pattern-either`**. This is called metavariable unification.
```
brandonspark, line 542: hm, i might also say that it's more that this only happens across patterns combined with `patterns`

of the six fundamental pattern operators, we have `pattern`, `pattern-not`, `pattern-inside`, `pattern-regex`, `pattern-either`, and `patterns`. the first four only take in, at maximum, one pattern as an argument, so they're out. of the latter two, only `pattern-either` and `patterns` can ingest numerous patterns, and only `patterns` does unification.

i would sooner view it as `patterns` being the weird one, than `pattern-either`, personally

docs/writing-rules/pattern-syntax.mdx
```diff
@@ -515,7 +515,37 @@ When referencing an ellipsis metavariable in a rule message or [metavariable-pat
...
+
+An anonymous metavariable always takes the form `$_`. Variables such as `$_1` or `$_2` are **not** anonymous. You can use more than one anonymous metavariable in a rule definition.
+
+For example, if you want to specify that a function should **always** have 3 arguments, then you can use anonymous metavariables:
+
+```yaml
+- pattern: def function($_, $_, $_)
+```
+
+An anonymous metavariable does not produce any binding to the code it matched. This means it does not store the code that it matched. The pattern:
```
brandonspark, line 530: i might hesitate to say "store" in case people conflate that with the notion of Semgrep Inc storing code.

"An anonymous metavariable does not produce any binding to the code it matched. This means it does not enforce that it matches the same code at each place it is used."

brandonspark (approved):
just a few more things!

docs/writing-rules/pattern-syntax.mdx
s-santillan: Thanks for the catch!

PR #1630, opened 2024-07-09
Remove won't share identifiable information bullet for Assistant code sharing

Sly1029:

Discussed with Jonathan and verified that this ok to do. This removal should go out after messaging is finished about this change

### Please ensure

- [X] A subject matter expert (SME) reviews the content
- [X] This change has no security implications or else you have pinged the security team


docs/semgrep-assistant/overview.md
```diff
@@ -124,7 +124,6 @@ Semgrep uses API permissions to access code on your pre-selected GitHub or GitLa
 * Semgrep, Inc. is a paying customer of OpenAI and has a Data Protection Agreement signed with them (provided upon request by [contacting support](/docs/support). The code snippets we upload are persisted by OpenAI temporarily, following their data usage policies at [Enterprise privacy at OpenAI](https://openai.com/enterprise-privacy).
 <!-- markdown-link-check-enable -->
 * Semgrep, Inc. takes the following steps to protect data that is processed by AI since Assistant requires the sharing of code snippets with a third party:
-  * Semgrep shares code snippets with OpenAI without identifying the customer or repository name.
```
s-santillan, line 127: @jsmoxon What exactly has changed here - do we share code snippets with openai at all? is it possible to at least keep part of this bullet (let's say we don't identify the customer, but we do send the repository name)?

jsmoxon: We do share code snippets with OpenAI. What has changed is that we now sometimes share the repository name. And sometimes it is possible to determine the customer name from the repository name. E.g. if the repo is called 'acme-foo-service`. 

Previously we didn't share the repo name. @Sly1029 correct me if I have this wrong.

Sly1029: Yeah now we have some desire to start sharing this information. But I wanted to leave it up to Jack to do the comms first

Sly1029: Ah yeah that second option sounds good

s-santillan (approved):
Approving for now, we'll make a follow up PR if there are any changes.

PR #1634, opened 2024-07-10
Remove PAT option for GHE, no longer available

armchairlinguist:
We removed the PAT connection option for GHE as the app is the recommended connection method.

### Please ensure


vivekkhimani (approved):
Stamping and confirming that the PAT section is not required anymore. Will let @khorne3 review the actual changes.

khorne3 (approved):

PR #1635, opened 2024-07-10
Clarify token terminology a bit

armchairlinguist:
The way "private rule and ruleset" was written was a bit unclear. This should be more clear. And we were using hyphens in a way that I don't think is correct?

### Please ensure

~- [ ] A subject matter expert (SME) reviews the content~ xs PR


s-santillan (approved):
Thank you for the catch!

PR #1640, opened 2024-07-11
add notes on using as-metavariable

khorne3:
adds `as-metavariable` info to experimental rules writing page.

### Please ensure

### DIFF
--- a/docs/writing-rules/experiments/pattern-syntax.md
+++ b/docs/writing-rules/experiments/pattern-syntax.md
@@ -210,17 +210,19 @@ where:
 
 ## <i class="fa-solid fa-exclamation"></i> `as-metavariable`
 
+> `as-metavariable` is only available in the new syntax.
+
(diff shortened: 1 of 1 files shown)


docs/writing-rules/experiments/pattern-syntax.md
```diff
@@ -208,6 +208,37 @@ where:
   - focus: $A
 ```
 
+## <i class="fa-solid fa-exclamation"></i> `as-metavariable`
+
+`as-metavariable` is a rule-writing feature that bridges the gap between metavariables and matches. Metavariables get access to things like `metavariable-comparison`, `metavariable-regex`, and `metavariable-pattern`, but you can’t use them on arbitrary matches. However, the `as` operator lets you embed arbitrary matches into metavariables, or bind arbitrary matches to a name.
+
+The syntax is as follows:
```
brandonspark, line 215: i might also make a note that it is _only_ valid in the new syntax

docs/writing-rules/experiments/pattern-syntax.md
```diff
@@ -208,6 +208,37 @@ where:
...
+
+```yaml
+all:
+ - pattern: |
+ @decorator
+ def $FUNC(...):
+ ...
+    as: $DECORATED_FUNC
```
brandonspark, line 223: this `as` is one character too far up, it should be vertically aligned with the `p` in `pattern`

brandonspark (approved):
thank you!

s-santillan (approved):

PR #1651, opened 2024-07-18
Note that multiple broker instances can cause log chatter

armchairlinguist:
This works but can cause log chatter which can be alarming if users aren't aware of that.

### Please ensure


docs/semgrep-ci/network-broker.md
```diff
@@ -127,4 +127,6 @@ You can check the logs for Semgrep Network Broker by running:
 
 ## Run multiple instances of the Semgrep Network Broker
 
-If necessary, you can run multiple instances of the Semgrep Network Broker. Semgrep handles its requests accordingly, preventing issues like duplicate PR or MR comments.
+If necessary, you can run multiple instances of the Semgrep Network Broker. Semgrep handles its requests accordingly, preventing issues like duplicate PR or MR comments. 
+
+You may see some contention in logs, as the broker hasn't yet been optimized for this use case.
```
khorne3, line 132: Would it be helpful to be more explicit on what users would see? For example, do the logs show duplicate entries or are they just looking at 3x the info in the event they have 3 brokers?

khorne3: Hey @armchairlinguist! I just wanted to follow up on this ☝️

armchairlinguist: Sorry @khorne3 , I didn't answer because I don't know for sure and I was hoping @tpetr might review/comment. I only know from discussions internally that it generates unexpected chatter in the logs (it made one user think it wasn't working to have multiple instances).

andy-r2c (approved):

PR #1657, opened 2024-07-24
Add size suggestion for bitbucket

relsqui:
This is a suggestion from a customer who ran into this problem on several projects. Feel free to rephrase it or decide it would be better elsewhere, but I wanted to pass it along.

### Please ensure

### DIFF
--- a/docs/semgrep-ci/sample-ci-configs.md
+++ b/docs/semgrep-ci/sample-ci-configs.md
@@ -261,6 +261,8 @@ These steps can also be performed through Bitbucket's UI wizard. This UI wizard 
 
 ### Sample Bitbucket Pipelines configuration snippet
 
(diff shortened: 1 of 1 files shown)


s-santillan:
@relsqui Thank you!

Sara to add
```
options:
  size: 2x
```

code snippet when her electricity comes back.

khorne3 (approved):

khorne3:
@s-santillan Sorry, I didn't see that you had looked at this already!

Do you think it's better to include the example for adding memory to the pipeline as a whole or just the step?

```
pipelines:
  default:
    - step:
        size: 2x
        script:
          - echo "This step gets double the memory!"
```

relsqui:
Not Sara, but: IMO the step, because this has a billing impact (2x the runner size will consume 2x the minutes).

s-santillan:
(LGTM, or nothing more for me to add, thank you @khorne3 and @relsqui )

PR #1666, opened 2024-07-30
Wording and command fix for historical secrets

armchairlinguist:
The provided `semgrep scan` command doesn't work anymore/right now. We might be able to restore it, but the ci + dry-run option seems most consistent for now.

And there was some weird wording.

TW reviewer: I'm uncertain how we should approach "historical scanning" vs "historical scans" in this article. The toggle in the app is currently called "Historical scanning" (though we could change that, in theory), but usage in the doc was mixed between "scanning" and "scans". I understand we're generally moving toward plain verbs rather than gerunds, but sometimes the use of plain verbs causes weird sentence structure, particularly in the sentence about diff-aware and historical scans. I welcome your feedback and suggestions to find the best option here.

### Please ensure

### DIFF
--- a/docs/semgrep-secrets/historical-scanning.md
+++ b/docs/semgrep-secrets/historical-scanning.md
@@ -13,7 +13,7 @@ tags:
 
 Detect valid, leaked secrets in previous Git commits through a **historical scan**.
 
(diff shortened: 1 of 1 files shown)


salolivares (approved):
lgtm!

PR #1669, opened 2024-07-31
Update CLI reference to use tabs

sinat101:
Using tabs for the `scan` vs `ci` output makes a lot more sense to me and looks pretty clear. I know there was discussion around why we didn't use tabs to begin with, but I'm hoping to spark that up again here now that we can also see what it looks like with tabs vs one long scrolling page.

Additionally, CTRL/CMD+F on the webpage is better when using tabs, so you don't get duplicate results.

### Please ensure

### DIFF
--- a/docs/cli-reference.md
+++ b/docs/cli-reference.md
@@ -9,6 +9,8 @@ hide_title: true
 title: CLI reference
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
 import CLIHelpOutput from '/src/components/reference/_cli-help-output.md'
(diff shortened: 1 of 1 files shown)


s-santillan:
@khorne3 Do you mind checking this out too? It's specifically: https://deploy-preview-1669--semgrep-docs-prod.netlify.app/docs/cli-reference#semgrep-scan-and-semgrep-ci-command-options

I am approving it, but your opinion (should you have one) is valued.

I do think it's ok as it's close to an anchor link.

s-santillan (approved):

khorne3 (approved):
Think this looks great!

PR #1681, opened 2024-08-02
Update SCM connection instructions to include GitLab Cloud, and add details for GitLab group name

maggieyu-semgrep:
Changes:
- We now support adding a SCM config for GitLab Cloud. Added the instructions for GitLab Cloud (copied the instructions for GitLab Self-Managed and updated some words).
- Added some details about GitLab group name

### Please ensure


static/img/source-code-manager.png
maggieyu-semgrep, line 1: Does the new image need to have the same width and height as the old one?

s-santillan: Nope! Also, wow that's a lot of SCM connections :tada:

s-santillan:
@khorne3 This is a note I should have mentioned this earlier with the Azure and BB docs (which I believe Maggie used as a base) - see the following screenshot: 
image

It looks like the instructions sort of jump the gun on Adding a CI job, which is a separate article. I was thinking maybe we can remove the "Ensure the SCM works" steps for 2 reasons:

1. It's redundant.
2. In the future these SCMs have the option to be managed by SMS so we wouldn't want to advise the extra step of testing them through a CI job connection as it wouldn't be used by the customer.

I am also setting you as reviewer as this was your domain recently and don't mean to step on your toes. I did check the text and it LGTM.

khorne3 (approved):

khorne3:
thanks @s-santillan -- I agree with your suggestion. I'll merge in this PR and make that change

maggieyu-semgrep:
Sorry, was OOO the past two days. Thank you for the edits and shipping this change! 🚢

PR #1682, opened 2024-08-05
Network broker: log verbosity howto and gentle warning

armchairlinguist:
Add note on how to adjust logging on the broker, and warn that it can lead to high memory usage if the req/res are large.

### Please ensure

### DIFF
--- a/docs/semgrep-ci/network-broker.md
+++ b/docs/semgrep-ci/network-broker.md
@@ -125,6 +125,34 @@ You can check the logs for Semgrep Network Broker by running:
 
 <pre class="language-console"><code>sudo docker logs <span className="placeholder">CONTAINER_ID</span></code></pre>
 
+### Adjusting log verbosity
+
+The Semgrep Network broker can log details of the proxied requests and responses for troubleshooting. Normally, these should only be collected when working to identify an issue. If requests and responses are large, they can occupy significant memory in the tunnel.
+
+In your broker configuration, that looks like:
+
+```yaml
+inbound:
+  logging:
+    logRequestBody: true
+    logResponseBody: true
+```
+
+In the logs, this leads to entries for `proxy.request` and `proxy.response`.
+
+These values can also be set on a per-allowlist basis:
+
+```
+inbound:
+  allowlist:
+    - url: https://httpbin.org/*
+      methods: [GET, POST]
+      logRequestBody: true
+      logResponseBody: true
+```
+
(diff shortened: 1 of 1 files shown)


vivekkhimani (approved):

docs/semgrep-ci/network-broker.md
```diff
@@ -125,6 +125,34 @@ You can check the logs for Semgrep Network Broker by running:
 
 <pre class="language-console"><code>sudo docker logs <span className="placeholder">CONTAINER_ID</span></code></pre>
 
+### Adjusting log verbosity
+
+The Semgrep Network broker can log details of the proxied requests and responses for troubleshooting. Normally, these should only be collected when working to identify an issue. If requests and responses are large, they can occupy significant memory in the tunnel.
+
+In your broker configuration, that looks like:
```
s-santillan, line 132: Is this an example we want people to copy?

s-santillan (approved):

docs/semgrep-ci/network-broker.md
armchairlinguist: yep, they can safely copy this example. the one below is also syntactically fine although wouldn't be very useful

s-santillan: Got it. It sounded a little hypothetical (or rather I wasn't sure if I would use it) - maybe we can be more explicit that the reader can use the snippet? Minor suggestion only.

armchairlinguist: How would you phrase that? I don't have great idea of how to make it more clear, especially given it follows a caution only to use it for debugging. :)

s-santillan, line 132: Maybe:

> To log additional details, add this snippet to your broker configuration:

armchairlinguist: Valid, but I don't know how well it works with the warning above 🤔 
I'd prefer to merge it as is, but I don't feel strongly - if you prefer this, I'm happy to accept it as a suggestion / push and then merge.

s-santillan: Yeap, The specific bit I was thinking about was "that looks like" - it seemed ambiguous. There's no indication of "if i want to troubleshoot, I should copy this thing" - if you are troubleshooting something, additional logging is a valid step to take even if it doesn't always lead to the solution.

I would have suggested "To log additional details for troubleshooting, add this snippet..." However, the text says "normally" used for troubleshooting - so there are other cases we may not know why a user wants to add logs. I will make the suggestion, thanks for being open to it.

s-santillan:
```suggestion
To log additional details, add this snippet to your broker configuration:
```

PR #1683, opened 2024-08-06
Add a caution about using literal metavariables with metavariable-pattern

relsqui:
This [came up in the community slack](https://semgrep.slack.com/archives/CK86BJ5DW/p1722930455768219) and it took a little digging to figure out [why my first attempt wasn't working](https://github.com/semgrep/semgrep/issues/6299#issuecomment-1287082436).

It might also be worth clarifying why the pattern here has two sets of quotes (the outer one gets stripped when the yaml file is interpreted -- it's only really necessary because the literal metavariable is the entire pattern) but I wasn't sure how to do that succinctly or if this is where it belonged, so I left it for a future update.

### Please ensure

### DIFF
--- a/docs/writing-rules/pattern-syntax.mdx
+++ b/docs/writing-rules/pattern-syntax.mdx
@@ -364,6 +364,25 @@ You can also use `/$X/` and `:$X` to respectively match
 any regular expressions or atoms (in languages that support
 those constructs, e.g., Ruby).
 
+:::caution
+Because literal metavariables bind to strings that may not be valid code, if you want to match them in more detail with a [`metavariable-pattern`](/writing-rules/rule-syntax/#metavariable-pattern), you'll need to [specify `generic` language](/writing-rules/rule-syntax#metavariable-pattern-with-nested-language) inside the `metavariable-pattern`. For example:
+
+```
+rules:
+  - id: match-literal-string
+    languages:
+      - python
+    severity: INFO
+    message: Found "$STRING"
+    patterns:
+      - pattern: '"$STRING"'
+      - metavariable-pattern:
+          language: generic
+          metavariable: $STRING
+          pattern: "literal string contents"
+```
+:::
+
 ### Typed metavariables
 
 #### Syntax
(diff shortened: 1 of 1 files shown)


docs/writing-rules/pattern-syntax.mdx
```diff
@@ -364,6 +364,25 @@ You can also use `/$X/` and `:$X` to respectively match
 any regular expressions or atoms (in languages that support
 those constructs, e.g., Ruby).
 
+:::info
+Because literal metavariables bind to strings that may not be valid code, if you want to match them in more detail with a [`metavariable-pattern`](/writing-rules/rule-syntax/#metavariable-pattern), you'll need to [specify `generic` language](/writing-rules/rule-syntax#metavariable-pattern-with-nested-language) inside the `metavariable-pattern`. For example:
```
s-santillan, line 368:
```suggestion
Because literal metavariables bind to strings that may not be valid code, if you want to match them in more detail with a [`metavariable-pattern`](/writing-rules/rule-syntax/#metavariable-pattern), you must [specify `generic` language](/writing-rules/rule-syntax#metavariable-pattern-with-nested-language) inside the `metavariable-pattern`. For example:
```

s-santillan (approved):
@relsqui Thank you!

docs/writing-rules/pattern-syntax.mdx
relsqui: Open to this change, but I'm interested in your reasoning -- does our brand voice lean more formal than I'm used to?

s-santillan: Hi, it's totally reasonable to ask that question! We use the google style guide. 

`Will` is not banned as much as it is in google, but it's still avoided where possible: 

- https://developers.google.com/style/word-list#will
- https://developers.google.com/style/tense

`You (will) need to` is longer than `You must` and where possible we try to shorten and be prescriptive, especially for instructions where the user _has_ to do the thing, there's no other option.

PR #1684, opened 2024-08-06
Add Component as a supported Jira field type

zdhickman:

### Please ensure

### DIFF
--- a/docs/semgrep-appsec-platform/jira.md
+++ b/docs/semgrep-appsec-platform/jira.md
@@ -88,6 +88,7 @@ Optionally, you can customize the Jira field mappings and indicate which Semgrep
 * Drop-down menus
 * Checkboxes
 * Labels
+* Components (including Compass components)
 
 ![Jira configuration screen for field data mappings](/img/jira-configure-defaults.png#md-width)
 _**Figure.** The Jira configuration screen for field data mappings._
--- a/docs/semgrep-appsec-platform/jira.md
+++ b/docs/semgrep-appsec-platform/jira.md
@@ -132,7 +132,7 @@ After setting up your Jira integration, you're now ready to create Jira tickets.
 
 If you've created a custom field mapping for a component field type, be aware that if you choose to create tickets in a project other than your default project as configured in your integration settings, you'll need to ensure that project has access to the component you selected for your mapping.
 
-If your default project uses [**Jira Components**](https://support.atlassian.com/jira-software-cloud/docs/what-are-jira-components/) and you create a component field mapping in your integration settings, **you'll be unable to create tickets in another project** because Jira components are always project-scoped.
+If your default project uses [**Jira Components**](https://support.atlassian.com/jira-software-cloud/docs/what-are-jira-components/) and you create a component field mapping in your integration settings, **you cannot create tickets in another project** because Jira components are always project-scoped.
 
 If your default project uses [**Compass Components**](https://support.atlassian.com/jira-software-cloud/docs/what-are-compass-components/) and you create a component field mapping in your integration settings, you can create tickets in another project **only if your selected component is available in that project**. You can configure which components are available in each project in your Compass settings.
 
--- a/docs/semgrep-appsec-platform/jira.md
+++ b/docs/semgrep-appsec-platform/jira.md
@@ -81,7 +81,7 @@ Issue types may have the same name, but a different Issue type ID. If you can't 
 
 ### Create mappings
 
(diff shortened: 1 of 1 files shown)


zdhickman:
@s-santillan two thoughts:

- I've added `Components` under the supported field types, but it's a bit inaccurate to the language above that list which specifically talks about mapping to Semgrep fields. Should this belong in its own section?
- I want to add a disclaimer about the edge case around creating tickets in another project via the project dropdown in the triage flow, but not sure where that may fit best in these docs

s-santillan:
Hi @zdhickman, thank you for opening this PR!

WRT your statements: 

1
> _but it's a bit inaccurate to the language above that list which specifically talks about mapping to Semgrep fields. Should this belong in its own section?_

When I took a look at your demo, the mapping from Jira field (**Components** in the drop-down) to Semgrep fields still uses the same form but with a new header (**Semgrep or static fields**) - was your demo mapping from a Jira field **Components** to a **static field**? I assume that means the values for the static field (such as `zach-test-component`) are actually Jira things and not Semgrep things? Does that mean that the mapping is some Jira-to-Jira map? I apologize if the questions sound silly.

2
> _I want to add a disclaimer about the edge case around creating tickets in another project via the project dropdown in the triage flow, but not sure where that may fit best in these docs_

This is quite important and I think worth making a separate header rather than a colored callout (i.e. you would use `###` so that our CSEs can create links to it when they help users troubleshoot. In this case, I would suggest putting it in the [Create tickets](https://semgrep.dev/docs/semgrep-appsec-platform/jira#create-tickets) section, above Code.

zdhickman:
>  was your demo mapping from a Jira field Components to a static field? I assume that means the values for the static field (such as zach-test-component) are actually Jira things and not Semgrep things? Does that mean that the mapping is some Jira-to-Jira map?

Definitely not a silly question - yes there are static values provided from Jira! We opted to not try and special-case where and how Components are mapped in the integration setup, and instead modified the labels (to Semgrep or Static) to be more accurate.

s-santillan:
@zdhickman In that case:

- The workflow is the same, so we can keep the **Create mappings** heading. We could add, to the paragraph before the list, "Semgrep can map from Jira fields to either Semgrep fields or static Jira fields." - or something similar.

zdhickman:
@s-santillan got it! I made my best attempt at a first pass, let me know what you think.

docs/semgrep-appsec-platform/jira.md
```diff
@@ -142,7 +151,7 @@ To create tickets:
 1. Set the status to **Open**, **Fixing**, or **Reviewing**. Select **Fixing** if it is a known issue that needs to be fixed or **Reviewing** if the finding needs more investigation.
 1. Select the **Create tickets...** checkbox.
     1. Optional: Click the first drop-down list to choose between making a ticket for **groups of findings** or an individual ticket for **each finding**.
-    1. Optional: Click the **JIRA project** drop-down list to select which Jira project to add the findings to.
+    1. Optional: Click the **JIRA project** drop-down list to select which Jira project to add the findings to. You can choose any project that is associated with the issue type configured in your integration settings.
```
zdhickman, line 154: Not related to components but think this would be good to mention here

s-santillan: SGTM!

docs/semgrep-appsec-platform/jira.md
```diff
@@ -127,6 +128,14 @@ If you opt for a drop-down or a checkbox issue type field, verify that:
 
 After setting up your Jira integration, you're now ready to create Jira tickets. Jira tickets can be created from findings in Semgrep Code, Supply Chain, and Secrets.
 
+### Component Mappings
+
+If you've created a custom field mapping for a component field type, be aware that if you choose to create tickets in a project other than your default project as configured in your integration settings, you'll need to ensure that project has access to the component you selected for your mapping.
+
+If your default project uses [**Jira Components**](https://support.atlassian.com/jira-software-cloud/docs/what-are-jira-components/) and you create a component field mapping in your integration settings, **you'll be unable to create tickets in another project** because Jira components are always project-scoped.
```
s-santillan, line 135:
```suggestion
If your default project uses [**Jira Components**](https://support.atlassian.com/jira-software-cloud/docs/what-are-jira-components/) and you create a component field mapping in your integration settings, **you cannot create tickets in another project** because Jira components are always project-scoped.
```

s-santillan (approved):
@zdhickman LGTM! I just suggested to add 'Jira' before some projects to make it clear we're talking about Jira projects and not Semgrep projects. Thank you!

PR #1689, opened 2024-08-08
Update metavariable-regex docs about anchoring

relsqui:
This addresses the documentation side of [#8577](https://github.com/semgrep/semgrep/issues/8577) (but not the question of how the regex parser should behave, only accurately representing how it does).

### Please ensure

- [ ] A subject matter expert (SME) reviews the content -- I did verify that all my examples work in the playground as described, if that counts

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -49,7 +49,7 @@ The below optional fields must reside underneath a `patterns` field.
 
 | Field            | Type     | Description           |
 | :--------------- | :------- | :-------------------- |
-| [`metavariable-regex`](#metavariable-regex)         | `map` | Search metavariables for [Python `re`](https://docs.python.org/3/library/re.html#re.match) compatible expressions; regex matching is **unanchored** |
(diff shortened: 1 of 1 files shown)


docs/writing-rules/rule-syntax.md
```diff
@@ -416,7 +416,23 @@ rules:
...
           metavariable: $METHOD
-          regex: (^insecure$)
+          regex: (insecure$)
+    message: module using insecure method call
+    languages:
+      - python
+    severity: ERROR
+```
+
+This one matches all of the function calls in the example, returning a false positive on the `module.secure` call:
```
khorne3, line 426:
```suggestion
The following example matches all of the function calls in the example, returning a false positive on the `module.secure` call:
```

khorne3 (approved):
lgtm but will also ask @brandonspark to take a look!

docs/writing-rules/rule-syntax.md
relsqui: Good call -- I'm going to make a separate commit so I can rephrase the second instance of "example" at the same time.

brandonspark (approved):

PR #1690, opened 2024-08-08
Point out `-j` in the Bitbucket memory tip

relsqui:
This came from a followup with the same user who suggested the original tip -- it turned out both were necessary in their case. I don't want to make a specific suggestion here because it would require tracking Bitbucket's infrastructure settings, so instead I'm pointing out both tuning tools.

### Please ensure

### DIFF
--- a/docs/semgrep-ci/sample-ci-configs.md
+++ b/docs/semgrep-ci/sample-ci-configs.md
@@ -291,7 +291,7 @@ You can customize the scan by entering custom rules or other rulesets to scan wi
 </Tabs>
 
 :::tip
-If the pipeline's default runner runs out of memory, you can [add the `size` directive](https://support.atlassian.com/bitbucket-cloud/docs/global-options/#Size) to the Semgrep step to increase the memory available:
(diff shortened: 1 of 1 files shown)


docs/semgrep-ci/sample-ci-configs.md
```diff
@@ -291,7 +291,7 @@ You can customize the scan by entering custom rules or other rulesets to scan wi
 </Tabs>
 
 :::tip
-If the pipeline's default runner runs out of memory, you can [add the `size` directive](https://support.atlassian.com/bitbucket-cloud/docs/global-options/#Size) to the Semgrep step to increase the memory available:
+If the pipeline's default runner runs out of memory, you can limit the number of subprocesses Semgrep uses with the [`-j` switch](/cli-reference), or [add the `size` directive](https://support.atlassian.com/bitbucket-cloud/docs/global-options/#Size) to the Semgrep step to increase the memory available:
```
khorne3, line 294:
```suggestion
If the pipeline's default runner runs out of memory, you can limit the number of subprocesses Semgrep uses with the [`-j` flag](/cli-reference), or [add the `size` directive](https://support.atlassian.com/bitbucket-cloud/docs/global-options/#Size) to the Semgrep step to increase the memory available:
```

khorne3 (approved):

relsqui:
... it's not worth fixing that commit message for how backwards I read it, confusing myself is punishment enough. :|

PR #1693, opened 2024-08-09
Update Jira component disclaimer

zdhickman:
We made a behavior change and this is section no longer accurate. Users can freely assign component mappings and can expect ticket creation in non-default projects to work if a component with the same name is available in that project.

### Please ensure


s-santillan (approved):
@zdhickman Thank you for the clarification!

PR #1699, opened 2024-08-12
delist ticketing from sidebars - may return one day

s-santillan:

### Please ensure

- [x] A living human being **Approves**

https://linear.app/semgrep/issue/TEC-86/ticketing-page-no-longer-private-beta-do-we-even-need-this-page


khorne3 (approved):
small nit: the Jira item isn't highlighted in the sidebar when you're on the page, but I can't quite figure out why

s-santillan:
cry, i will look into why

PR #1709, opened 2024-08-15
update samples to indicate that the token should be provided with the…

khorne3:
… base url

### DIFF
--- a/docs/semgrep-ci/network-broker.md
+++ b/docs/semgrep-ci/network-broker.md
@@ -89,17 +89,17 @@ Update the `config.yaml` by replacing the SCM information containing `YOUR_BASE_
   # for GitLab
   gitlab:
     baseUrl: <https://gitlab.exampleCo.net/api/v4>
-    token: gitlab-pat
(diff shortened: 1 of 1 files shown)


s-santillan:
I checked the preview and saw two things I think I can fix (since they were my fault)

(1) Use of sample values (`<https://github.com/exampleco.com/api/v3>`) mixed with PLACEHOLDER (GITHUB_TOKEN). - this  is because it can be hard for the user to know "what should the value here look like". I actually okayed this idea and so falls on me to figure out how to be consistent and show both bits of information. See [Slack thread](https://semgrepinc.slack.com/archives/C02JT6YA5S7/p1720627278174069) for context.

image

(2) The `pre` / `code` / `span` wrap here is broken:
image


**Action item: I'll make a branch (out of this PR specifically) that you can choose to merge back into this PR.**.

s-santillan:
OMG haha that's better than what I was going to do (I was going to have tabs with a template snippet in one tab and sample values in the other)

s-santillan (approved):

PR #1718, opened 2024-08-22
Semgrep Editor: update defaults/tiering, discuss rule organization

armchairlinguist:
This doc had gotten a bit out of date with some of the current defaults of the editor. It also needed a few minor fixes. I added a short section on organizing rules because we often get questions about creating folders for custom rules and it isn't possible.

I changed one header tag, but found no instances of that link to update.

I don't think this change has security implications. I am updating the doc to describe a newer behavior that is more secure by default.

### Please ensure

### DIFF
--- a/docs/semgrep-code/editor.md
+++ b/docs/semgrep-code/editor.md
@@ -217,7 +217,7 @@ To run your rule against selected repositories or projects:
 
 Upon saving, a rule’s visibility is **private** by default. A private rule is visible only to members within an organization.
 
-- To share a rule, click **Share > Public > Confirm**. If you want to share this specific version of the rule, you can also toggle Permalink. This provides a shortlink to this version of the rule, which will not change if the rule is modified.
+- To share a rule outside your organization, click **Share > Public > Confirm**. If you want to share this specific version of the rule, you can also toggle Permalink. This provides a shortlink to this version of the rule, which will not change if the rule is modified.
 - To share a private rule with those who can access it, click **Share** and copy the **URL link**.
 
 Some older rules in Semgrep AppSec Platform may be **unlisted** rather than private. These rules are marked with an icon without a lock, and can be shared with anyone, including those who cannot access Semgrep AppSec Platform.
--- a/docs/semgrep-code/editor.md
+++ b/docs/semgrep-code/editor.md
@@ -20,7 +20,7 @@ import InstallPrivateGitHubApp from "/src/components/procedure/_install-private-
 
 **Semgrep Editor** allows you to write rules, verify their performance through tests, and add them to your organization’s [Policies page](/semgrep-code/policies) to enforce code standards and increase code security.
 
-The Editor is free to use on all subscription tiers, but saving a rule for **private use** is a **Team/Enterprise tier feature**.
+The Editor is free to use on all subscription tiers.
 
 ## Access Semgrep Editor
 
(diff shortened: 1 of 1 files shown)


docs/semgrep-code/editor.md
```diff
@@ -215,12 +215,14 @@ To run your rule against selected repositories or projects:
 
 ## Set a rule’s visibility and share a rule
 
-Upon saving, a rule’s visibility is **unlisted** by default. This rule can be shared with anyone through an identifier, even to non-Semgrep AppSec Platform users.
+Upon saving, a rule’s visibility is **private** by default. A private rule is visible only to members within an organization.
 
-A rule can be saved as a **private rule**, which is visible only to members within an organization. You can still share a private rule, but only members of the organization can see it. Private rules are a **Team/Enterprise tier feature**.
+- To share a rule, click **Share > Public > Confirm**. If you want to share this specific version of the rule, you can also toggle Permalink. This provides a shortlink to this version of the rule, which will not change if the rule is modified.
```
relsqui, line 220:
```suggestion
- To share a rule outside your organization, click **Share > Public > Confirm**. If you want to share this specific version of the rule, you can also toggle Permalink. This provides a shortlink to this version of the rule, which will not change if the rule is modified.
```

relsqui: this is covered in the next line, but never rely on someone reading two lines of docs 😂

docs/semgrep-code/editor.md
```diff
@@ -244,6 +246,10 @@ To add a rule to the **Policies** page:
 
 If successful, you'll see a pop-up window indicating that your rule has been added.
 
+## Organize private rules
+
+All private rules for an organization are saved to the organization's folder. To further organize rules, consider organizational naming conventions to keep related rules near each other, such as by internal team, rule language, or vulnerability category.
```
relsqui, line 251: I regret not thinking of this response when I got a ticket about it today

relsqui (approved):
one non-blocking suggestion, thank you for the update 🙏

docs/semgrep-code/editor.md
relsqui: actually, wait ... can you sort by name in the editor? I don't see a way to do that, and without it, namespacing doesn't help

armchairlinguist: oh...good call. let me check

armchairlinguist: yeah no, I was making that up. Darn. Maybe we can put in a request to make it work that way...

armchairlinguist: I think I'll revise it to "to make it easier to find/search for rules", how does that sound?

relsqui: I think that's fine, although including this section at all still feels a little like advertising what the product can't do yet. it hasn't come up enough for me to have strong feelings though.

s-santillan: I have no strong feelings about keeping or removing, but I do think giving the user a tip to follow a naming convention (as we have nothing else) helps them use the feature; an organizational tip is better than nothing :)

docs/semgrep-code/editor.md
```diff
@@ -215,12 +215,14 @@ To run your rule against selected repositories or projects:
 
 ## Set a rule’s visibility and share a rule
 
-Upon saving, a rule’s visibility is **unlisted** by default. This rule can be shared with anyone through an identifier, even to non-Semgrep AppSec Platform users.
+Upon saving, a rule’s visibility is **private** by default. A private rule is visible only to members within an organization.
```
s-santillan, line 218: Thank you for updating this bit. I hadn't realized it had gotten so out of date.

s-santillan (approved):

PR #1755, opened 2024-09-16
Sara/tec 46 revamp and redesign docs homepagelanding page

s-santillan:

### Please ensure


s-santillan:
Current progress:

image

Remaining tasks:

- [ ] Check that we didn't impact other parts of the site
- [ ] Check that dark mode wasn't impacted negatively
- [ ] Check across all breakpoints
- [ ] Maybe add Supported languages link
- [ ] Maybe 2-col the bottom "help" footer?

s-santillan:
Dark mode issues (note that bg color is #1b1b1d): 
image

s-santillan:
Dark mode for now
image

khorne3:
- Both modes look good to me as-is!
- I think the help footer is fine as single-column (does it look better as two-column? We do have two lines of text, so maybe a simpler fix is to clean up the language)
- Our most popular pages are the homepage, the quickstart, the CLI reference, and then supported languages, so adding the link would be very useful.
- WDYT about making the text that starts with "Deploy static application security testing (SAST)..." a bit smaller, which would help move the "Enhance your Semgrep experience" section a bit?
- I didn't think the lack of text under "Enhance your Semgrep experience" was strange -- I think I lean toward less text on the homepage

s-santillan:
@khorne3 Thank you! Some replies:

> Our most popular pages are the homepage, the quickstart, the CLI reference, and then supported languages, so adding the link would be very useful.

Maybe we should add a **Most popular resources** section? And since they're the most popular pages, bump them up above  **Enhance your Semgrep experience**?

> WDYT about making the text that starts with "Deploy static application security testing (SAST)..." a bit smaller, which would help move the "Enhance your Semgrep experience" section a bit?

I can do that!

> I think the help footer is fine as single-column (does it look better as two-column? We do have two lines of text, so maybe a simpler fix is to clean up the language)

The design I made looked like this:

image

Can probably do it in a separate PR.

khorne3:
> Maybe we should add a Most popular resources section? And since they're the most popular pages, bump them up above Enhance your Semgrep experience?

What about adding CLI reference and Supported languages to just above the Knowledge base link in the sidebar (or is that weird?)

Re: Footer -- I think single-column looks a bit better

s-santillan:
> What about adding CLI reference and Supported languages to just above the Knowledge base link in the sidebar (or is that weird?)

I'd like to preserve the categorization & content hierarchy of our top-level sidebar. All of the links take them to a top-level doc and 2nd-level sidebar. It'd be weird if the "tree" went from trunk to leaf before passing through a branch/stem, if that analogy helps. I'll make a separate section.

> I think single-column looks a bit better

Alright, we'll shelve the design for now. Convenience link if you'd like to be reminded of what we're working out of https://www.figma.com/design/JIJCiYT1MC2hTtotpjOfsJ/Docs-v4-landing-page?node-id=0-1&node-type=canvas&t=2CCialbXyhypHWBp-0

khorne3 (approved):

PR #1758, opened 2024-09-17
Be clear that removing the workflow file is typical but optional

armchairlinguist:
Some folks are not converting fully to Managed Scans for every repo they migrate, so this explains that you can retain the workflow file if it's appropriate.

### Please ensure


bandarisantosh (approved):
Looks good to me!

khorne3 (approved):

PR #1764, opened 2024-09-19
Changing rule severity or other metadata, componentized

sinat101:

### Please ensure


docs/kb/rules/changing-rule-severity-and-other-metadata.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Changing the severity
+
+Once you have forked the rule, you can change the [severity or other metadata](/docs/writing-rules/rule-syntax#required) to your liking. 
+
+Then, save this custom version of the rule to your organization's rules, making it available to use within your policy as defined in Semgrep AppSec Platform.
+
+![Save a rule in the Editor](/img/kb/save_rule_editor.png)
+
+By default, saving the rule also includes it in the Semgrep Registry, with visibility limited to your organization.
```
s-santillan, line 31:
```suggestion
By default, saving the rule also enables you to search for it in the [Semgrep Registry](https://semgrep.dev/r), with visibility limited to your organization.
```
So - in other parts of the Semgrep ecosystem/product, we say stuff like **Contribute to Registry** or that the registry is the source of community rules. If we say "includes it in the Registry" that may imply that the rule is uploaded to the semgrep-rules repo (which users are aware of and fork). I am aware that the Semgrep Registry is more than community rules, but it is a bit tricky because we have to not contradict wording elsewhere.

Just to sidestep that bit, I reworded your text a little.

src/components/reference/_fork_existing_rule.md
```diff
@@ -0,0 +1,20 @@
+One way to create new rules is to fork an existing rule in Semgrep Registry and modify it to meet your software and business requirements.
```
s-santillan, line 1:
```suggestion
One way to create new rules is to fork an existing rule in the Semgrep Registry and modify it to meet your software and business requirements.
```

These are very murky waters but the Semgrep Registry is not a T1 product so I think we can use articles for it as though it were a common noun.

(For context, T1 products are treated like Proper Nouns even though they might use common nouns; a good example is Semgrep Assistant. For non-Semgrep examples, see the [GitHub Copilot docs](https://docs.github.com/en/copilot/using-github-copilot/getting-code-suggestions-in-your-ide-with-github-copilot) - you'd normally say "the copilot has taken over the flight" in regular English, but no such a/an/the article appears for GitHub's specific branded Copilot.)

s-santillan: cc @khorne3 if you have any issues with how English is getting bent out of shape with all the common-noun trademarking in tech.

s-santillan (approved):

PR #1777, opened 2024-09-27
Update azure-pr-comments.md

tembalii:

### Please ensure

### DIFF
--- a/docs/semgrep-appsec-platform/azure-pr-comments.md
+++ b/docs/semgrep-appsec-platform/azure-pr-comments.md
@@ -59,43 +59,51 @@ PR comments are enabled by default for users who have connected their Azure DevO
 Only rules set to the **Comment** and **Block** rule modes in the [Policies page](https://semgrep.dev/orgs/-/policies) create PR comments.
 :::
 
-> **_Please Note: In the workflow file, it's crucial to export the SEMGREP_REPO_URL variable to enable PR comments. Ensure that the namespace follows the format {organization}/{project}. For example: `export SEMGREP_REPO_URL="https://dev.azure.com/{organization}/{project}/_git/{project}"`_**
-
-Sample Workflow File below: 
+In the Azure Pipelines configuration file, you must export the `SEMGREP_REPO_URL` variable to enable PR comments. Ensure that the namespace as part of the variable's value follows the format `{organization}/{project}`:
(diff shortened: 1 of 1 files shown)


zyannes (approved):

khorne3 (approved):

armchairlinguist:
@zyannes I see you approved, but I'm sort of confused why people using a supported SCM **have** to set the repo name; that doesn't seem great. Shouldn't it be that if the SCM is connected, it can determine the right URL?

docs/semgrep-appsec-platform/azure-pr-comments.md
```diff
@@ -59,6 +59,52 @@ PR comments are enabled by default for users who have connected their Azure DevO
...
+
+      pip install semgrep
+
+      if [ $(Build.SourceBranchName) = "main" ]; then
+          echo "Semgrep full scan"
+          semgrep ci
+      elif [ $(System.PullRequest.PullRequestId) -ge 0 ]; then
+          echo "Semgrep diff scan"
+          export SEMGREP_PR_ID=$(System.PullRequest.PullRequestId)
+          export SEMGREP_REPO_URL="https://dev.azure.com/{organization}/{project}/_git/{project}"
```
zyannes, line 93: We also recommend that they set the `SEMGREP_REPO_NAME` environment variable. It can be set via 
```
export SEMGREP_REPO_NAME="{organization}/${SYSTEM_TEAMPROJECT}/${BUILD_REPOSITORY_NAME}"
```
Similarly the `SEMGREP_REPO_URL` can be simplified to 
```
export SEMGREP_REPO_URL="https://dev.azure.com/{organization}/${SYSTEM_TEAMPROJECT}/_git/${BUILD_REPOSITORY_NAME}"
```

khorne3:
> Shouldn't it be that if the SCM is connected, it can determine the right URL?

@armchairlinguist I spoke to Zach, and it turns out that sometimes we get a project GUID instead of the project name or similar when building the URL. This works most of the time, but it can be confusing to see a GUID instead of a project name in Platform. So, the vars are needed at this time, at least until [we update the CLI to handle this behavior](https://linear.app/semgrep/issue/FS-1513/update-azurepipelinesmeta-repo-url-logic-in-cli).

PR #1783, opened 2024-10-10
Correct API URLs for GH/GL

armchairlinguist:
When the formatting for these URLs was redone, they were also all set to the same path value after the base URL, but that's not right - the API path for each SCM is different and needs to be correct or the customer's config won't work.

### Please ensure


maggieyu-semgrep (approved):
thank you!!

PR #1793, opened 2024-10-18
Add check runs whitelist for NB with SMS

armchairlinguist:
Check run updates specifically require a PATCH, which we didn't have in the allowlist we suggest for SMS. This adds it.

### Please ensure

I don't think this change has security implications, inasmuch as we're already advising whitelisting the whole instance for GET/POST, but feel free to request security review if you disagree!

### DIFF
--- a/docs/semgrep-ci/network-broker.md
+++ b/docs/semgrep-ci/network-broker.md
@@ -180,7 +180,9 @@ inbound:
 
 ## Use Semgrep Network Broker with Managed Scans
 
-Semgrep Managed Scans uses Semgrep Network Broker to connect to your internal source code management instance. To clone repositories for scanning from any organization or group, the URL allowlist must include the base URL of your instance. For example, if your source code manager is at `https://git.example.com/`, the following allowlist will permit cloning repositories:
+Semgrep Managed Scans uses Semgrep Network Broker to connect to your internal source code management instance. 
+
+To clone repositories for scanning from any organization or group, the URL allowlist must include the base URL of your instance. For example, if your source code manager is at `https://git.example.com/`, the following allowlist will permit cloning repositories:
 
 ```yaml
 inbound:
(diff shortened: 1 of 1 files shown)


docs/semgrep-ci/network-broker.md
```diff
@@ -189,6 +191,17 @@ inbound:
...
 ```
+
+Semgrep also creates and updates [GitHub Checks](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/collaborating-on-repositories-with-code-quality-features/about-status-checks#checks) when performing Managed Scans on pull requests. To ensure checks can be both created and updated, add the `PATCH` method to the preceding allowlist example, or add a separate entry to allowlist check updates:
+
+```
+inbound:
+  allowlist:
+    # allow PATCH requests to update checks
+    - url: https://git.example.com/api/v3/repos/:owner/:repo/check-runs/:id
+      methods: [GET, POST, PATCH]
```
chmccreery, line 202: Can we just recommend including PATCH from the get-go? I'm worried people will see the first thing and stop reading.

zyannes (approved):

docs/semgrep-ci/network-broker.md
armchairlinguist: We definitely can; I went this way first because it's more "least-privilege" oriented, though. What's your thought on that? Not worth it?

s-santillan (approved):
lgtm for the technical writing side of things! thank you for keeping this complicated doc up-to-date.

docs/semgrep-ci/network-broker.md
armchairlinguist: @chmccreery it looks like other folks have approved - are you open to me merging now, and we can consider the change if we find this version isn't helpful?

PR #1802, opened 2024-10-24
Update ignores table: change order, refine descriptions

armchairlinguist:
Update table to make the order work and get the details just right, remove stale known issue.
In this case I'm asserting myself as the SME since I tested this yesterday. :)

### Please ensure


s-santillan (approved):
:sonic:

PR #1828, opened 2024-11-04
update Broker info

khorne3:
- added info on multiple instances and using a single config per deployment
- token scopes info has been added to SCM docs at an earlier date

### DIFF
--- a/docs/semgrep-ci/network-broker.md
+++ b/docs/semgrep-ci/network-broker.md
@@ -178,30 +178,32 @@ inbound:
     verbose: true
 ```
 
-## Use Semgrep Network Broker with Managed Scans
-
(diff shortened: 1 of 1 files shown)


armchairlinguist (changes requested):
Hey @khorne3 - I'm kind of confused by these changes. It seems like you've removed information that we still need in order to add these details. This doesn't replace the additional information about using the broker with SMS, it's a separate case where you're potentially adding multiple SCMs to the broker.

khorne3:
@armchairlinguist I'm really sorry, I accidentally deleted the SMS section when making these additions 🤦‍♀️  I didn't mean to do that, so I've re-added that info.

armchairlinguist:
No worries! I'm glad it was accidental actually. I'll re-review now

PR #1839, opened 2024-11-09
add entry into rule syntax

s-santillan:

### Please ensure

FYI: 
I said that it's **required** for rules that use cross-function, cross-file, which is true, but to be fair, the rule won't fail if the user doesn't provide it, which makes me wonder if **required** is too strong. Still, if they want the functionality, they must set the option, so I think it's warranted to say **required**.

Preview:
image


khorne3 (approved):

s-santillan:
@armchairlinguist Hello! Any comments before this is merged?

armchairlinguist (approved):

PR #1840, opened 2024-11-11
add info

khorne3:

### Please ensure

### DIFF
--- a/docs/semgrep-code/remove-duplicates.md
+++ b/docs/semgrep-code/remove-duplicates.md
@@ -5,8 +5,8 @@ title: Remove duplicate findings
 hide_title: true
 description: Learn how to remove duplicate findings and prevent them from displayed in Semgrep AppSec Platform.
 tags:
-  - Semgrep Code
-  - Semgrep AppSec Platform
+ - Semgrep Code
+ - Semgrep AppSec Platform
 ---
 
 # Remove duplicate findings
@@ -14,17 +14,25 @@ tags:
 Semgrep scans are performed on both mainline (trunk) and non-mainline branches. The scope of the scan can differ depending on if Semgrep is called on a mainline or non-mainline branch.
 
 <dl>
-    <dt>Full scan</dt>
-    <dd>Scans the repository in its entirety. It is recommended to perform full scans on mainline branches, such as <code>master</code> or <code>main</code>. This scan is performed on a scheduled basis.</dd>
-    <dt>Diff-aware scan</dt>
-    <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch (or diverged from the last commit that was fully scanned)</dd>
+ <dt>Full scan</dt>
+ <dd>Scans the repository in its entirety. It is recommended to perform full scans on mainline branches, such as <code>master</code> or <code>main</code>. This scan is performed on a scheduled basis.</dd>
+ <dt>Diff-aware scan</dt>
+ <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch (or diverged from the last commit that was fully scanned)</dd>
 </dl>
 
-## Remove duplicate findings using Semgrep AppSec Platform
+Regardless of scan type, Semgrep generates a finding whenever a Semgrep rule matches a piece of code.
 
-Regardless of the scope of a scan, Semgrep correlates findings across branches based on their unique fingerprint, automatically deduplicating findings and making it simpler to triage.
+For each finding identified by Semgrep, Semgrep tags the finding with a unique fingerprint that allows Semgrep to track the finding as it exists in a file over time. Semgrep generates this fingerprint by combining and hashing information retrieved during runtime, including file path, rule name, and pattern, and the metavariables substituted into the pattern.
 
-If a finding is fixed in one branch (such as `main`) but open in another (such as `production`), and the code fixes are present in both branches, initiate a scan through your CI job or SCM tool on the branch(es) with open findings to have Semgrep mark the findings as fixed.
+Additionally, Semgrep appends an index value that reflects the number of occurrences of a finding matching the same rule in a single file. The index helps Semgrep determine how two or more findings are related to each other. In other words, a finding with fingerprint `123_0` and a finding with fingerprint `123_1` indicates that the findings were generated from the same rule that matched a given piece of code in the same file.
+
+With this fingerprint, Semgrep can determine if a finding with the same fingerprint already exists. If not, it is considered a new finding.
+
+## Deduplicate findings by rescanning the project
+
+Semgrep's correlation of findings across branches based on their unique fingerprint allows for automatic deduplication of findings and makes it simpler for you to triage findings.
+
+If a finding is fixed in one branch (such as `main`) but open in another branch (such as `production`), and the code fixes are present in both branches, initiate scans through your CI job or SCM tool on the branches with open findings. Semgrep will reconcile the findings and mark them as fixed.
 
 ## Remove duplicate findings using Semgrep API
 


docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,34 @@ title: Remove duplicate findings
...
 
-If a finding is fixed in one branch (such as `main`) but open in another (such as `production`), and the code fixes are present in both branches, initiate a scan through your CI job or SCM tool on the branch(es) with open findings to have Semgrep mark the findings as fixed.
+Additionally, Semgrep appends an index value that reflects the number of occurrences of a finding matching the same rule in a single file. The index helps Semgrep determine how two or more findings are related to each other. In other words, a finding with fingerprint `123_0` and a finding with fingerprint `123_1` indicates that the findings were generated from the same rule that matched a given piece of code in the same file.
+
+With this fingerprint, Semgrep can determine if a finding with the same fingerprint already exists. If not, it is considered a new finding.
```
pangngan, line 29: Passing thought (feel to shoot this down), but one way users can visualize this is in two places: 
1. The branch status widget on the finding details page
CleanShot 2024-11-11 at 14 17 07@2x

2. Hovering over the branch associated with the finding on the findings list page.

CleanShot 2024-11-11 at 14 18 14@2x

I think it's probably fine to skip (2), but if you'd like something visual for users to associate with, an option would be to point to the branch status widget (1).

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,34 @@ title: Remove duplicate findings
...
+ <dt>Diff-aware scan</dt>
+ <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch (or diverged from the last commit that was fully scanned)</dd>
 </dl>
 
-## Remove duplicate findings using Semgrep AppSec Platform
+Regardless of scan type, Semgrep generates a finding whenever a Semgrep rule matches a piece of code.
 
-Regardless of the scope of a scan, Semgrep correlates findings across branches based on their unique fingerprint, automatically deduplicating findings and making it simpler to triage.
+For each finding identified by Semgrep, Semgrep tags the finding with a unique fingerprint that allows Semgrep to track the finding as it exists in a file over time. Semgrep generates this fingerprint by combining and hashing information retrieved during runtime, including file path, rule name, and pattern, and the metavariables substituted into the pattern.
```
pangngan, line 25: I think it's fine to include this information in our docs, my suggestion would be to just include some sort of callout to let users know that this calculation could change and be improved upon any time. 

We'd ideally keep this doc updated when we do, so hopefully that will help with the reliability of this information

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+ <dd>Scans the repository in its entirety. It is recommended to perform full scans on mainline branches, such as <code>master</code> or <code>main</code>. This scan is performed on a scheduled basis.</dd>
+ <dt>Diff-aware scan</dt>
+ <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch (or diverged from the last commit that was fully scanned)</dd>
 </dl>
 
-## Remove duplicate findings using Semgrep AppSec Platform
+## How Semgrep distinguishes between new and duplicate findings
 
-Regardless of the scope of a scan, Semgrep correlates findings across branches based on their unique fingerprint, automatically deduplicating findings and making it simpler to triage.
+Semgrep generates a finding whenever it scans a project, and one of its rules matches a piece of code in the project. Since Semgrep scans a project multiple times, it needs a way to track the same finding in a file over time. Semgrep does this using two types of fingerprints: `syntactic_id` and `match_based_id`.
```
s-santillan, line 25:
```suggestion
Semgrep generates a finding whenever it scans a project (repository), and one of its rules matches a piece of code in the project. Since Semgrep scans a project multiple times, it needs a way to track the same finding in a file over time. Semgrep does this using two types of fingerprints: `syntactic_id` and `match_based_id`.
```
Or, I think we refer to the scan target as a repository in the above section, so it can also be "repository or project" - you decide!

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+- The file path
+- The name of the rule that generated the finding
+- The code syntax, or the literal piece of code that matched the rule
+- The index, a value generated by determining the number of times the rule involved matched code in the file
+
+This information is combined and then hashed before storing due to privacy concerns.
+
+Using the `syntactic_id`, Semgrep can determine if a given finding in a file is 
+the same as a finding identified during a different scan, even if the code snippet that the rule matched had been moved to a different location in the file. This allows Semgrep to avoid generating a new finding and deduplicating its records accordingly, even across multiple branches associated with the project. It also means that Semgrep can cross-correlated findings, so a finding that has been triaged in one branch will be flagged as triaged if it's identified in another branch.
```
s-santillan, line 44:
```suggestion
the same as a finding identified during a different scan, even if the code snippet that the rule matched had been moved to a different location in the file. This allows Semgrep to avoid generating a new finding and deduplicating its records accordingly, even across multiple branches associated with the project. It also means that Semgrep can cross-correlate findings, so a finding that has been triaged in one branch will be flagged as triaged if it's identified in another branch.
```

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+- The name of the rule that generated the finding
+- The code syntax, or the literal piece of code that matched the rule
+- The index, a value generated by determining the number of times the rule involved matched code in the file
+
+This information is combined and then hashed before storing for privacy.
+
+Using the `syntactic_id`, Semgrep can determine if a given finding in a file is 
+the same as a finding identified during a different scan, even if the code snippet that the rule matched had been moved to a different location in the file. This allows Semgrep to avoid generating a new finding and deduplicating its records accordingly, even across multiple branches associated with the project. It also means that Semgrep can cross-correlate findings, so a finding that has been triaged in one branch will be flagged as triaged if it's identified in another branch.
+
+### `match_based_id`
```
armchairlinguist, line 46: I think we ought to highlight MBID first since it's considered the better one.

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+While the `syntactic_id` associated with a finding helps track findings within the same file, there are times when the way Semgrep generates the `syntactic_id` isn't sufficient for the identification and deduplication of findings. 
+
+Semgrep generates the `match_based_id` for a finding using the following information for a finding:
+
+- The file path
+- The name of the rule that generated the finding
+- The rule pattern with the metavariables' values substituted in
+
+This information is combined and then hashed. At this point, Semgrep appends the **index**, a value generated by determining the number of times the rule involved matched code in the file. Note that the index is appended to the hash, not combined with the other finding information before hashing. This is done to preserve information on how findings are related. For example, `finding0` with `match_based_id = 123_0` and `finding1` with `match_based_id = 123_0` indicate that both were generated from the same rule matching the same code in the same file.
```
armchairlinguist, line 56: This is confusing. There should rarely even be two distinct findings with the same match-based ID. If you want to explain the index, a better approach would be to change one of the indices to 1, and explain that these are the same rule matching two instances of similar code (remember, match-based ID doesn't use the literal code, it uses the rule pattern) in the same file.

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+Because Semgrep uses the rule pattern instead of the literal code syntax to generate `match_based_id`, code changes that do not impact the overall function don't hinder Semgrep's ability to recognize that the finding isn't a duplicate of an existing finding.
```
armchairlinguist, line 58: I don't know if this is very clear either. Maybe it would be better to say "code changes that don't impact the match".

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+![Semgrep AppSec Platform groups together findings on different branches](/img/matched-findings.png)
+_**Figure**. Semgrep AppSec Platform groups two findings on separate branches that are duplicates._
```
armchairlinguist, line 61: I wouldn't use the word "duplicates" here since it suggests this is somehow wrong and there shouldn't be two of them. It might be more clear to say "groups together the same finding on different branches" and reference how that helps us show that the same finding has been seen in multiple places.

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+The rule pattern identified and used in generating the `match_based_id` does not change:
+
+```python
+spcd.get("foo")
+...
+sink("foo")
+```
+
+This means that the `match_based_id` itself doesn't change, allowing Semgrep to identify that the two findings are the same, allowing deduplication. Furthermore, this process enables Semgrep to ignore lines that do not impact code function. For example, the addition of comments to a code file would create a new `syntactic_id`, but not a new `match_based_id`.
```
armchairlinguist, line 102: I don't know that comments would change the syntactic ID - do you have verification of that?

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+### How Semgrep AppSec Platform handles findings using `syntactic_id` and `match_based_id`
```
armchairlinguist, line 104: I'm not convinced we should be this detailed here. The actual interaction of syntactic ID and match-based ID in this context still confuses even me (for example, I've been told there are cases where we still use syntactic ID for certain things that I don't think fall into item 1)

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
+
+For each finding generated during a scan, Semgrep performs the following actions before displaying a finding in Semgrep AppSec Platform:
+
+1. If  `match_based_id = NONE`, set `match_based_id` equal to `syntactic_id`. This applies only to legacy scans run using Semgrep version `0.102.0` and earlier.
+1. Determine if the finding exists already using `match_based_id`.
+1. If the finding exists, update the existing finding's `syntactic_id`, status, and metadata to the new finding's `syntactic_id`, status, and metadata.
+1. If the finding does NOT exist, create a new finding.
+
+## Deduplicate findings by rescanning the project
```
armchairlinguist, line 113: This isn't really what's being described. This doesn't deduplicate findings (and they aren't duplicates to begin with, imo), it just ensures the status is up to date on all the branches.

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
 
 <dl>
-    <dt>Full scan</dt>
-    <dd>Scans the repository in its entirety. It is recommended to perform full scans on mainline branches, such as <code>master</code> or <code>main</code>. This scan is performed on a scheduled basis.</dd>
-    <dt>Diff-aware scan</dt>
-    <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch (or diverged from the last commit that was fully scanned)</dd>
+ <dt>Full scan</dt>
+ <dd>Scans the repository in its entirety. It is recommended to perform full scans on mainline branches, such as <code>master</code> or <code>main</code>. This scan is performed on a scheduled basis.</dd>
+ <dt>Diff-aware scan</dt>
+ <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch (or diverged from the last commit that was fully scanned)</dd>
```
armchairlinguist, line 20:
```suggestion
 <dd>Diff-aware scans are performed on non-mainline branches, such as in pull requests and merge requests. Diff-aware scans traverse the repository's files based on the commit where the branch diverged from the mainline branch.</dd>
```
I know this content didn't change but I'm pretty sure the last part isn't right. Our diff-aware scans are essentially two scans in one, and are entirely self-contained - they don't refer back to previous full scans, except with regard to which issues have been triaged.

docs/semgrep-code/remove-duplicates.md
```diff
@@ -5,26 +5,116 @@ title: Remove duplicate findings
...
-Regardless of the scope of a scan, Semgrep correlates findings across branches based on their unique fingerprint, automatically deduplicating findings and making it simpler to triage.
+Semgrep generates a finding whenever it scans a project, and one of its rules matches a piece of code in the project. Since Semgrep scans a project multiple times, it needs a way to track the same finding in a file over time. Semgrep does this using two types of fingerprints: `syntactic_id` and `match_based_id`.
 
-If a finding is fixed in one branch (such as `main`) but open in another (such as `production`), and the code fixes are present in both branches, initiate a scan through your CI job or SCM tool on the branch(es) with open findings to have Semgrep mark the findings as fixed.
+
+:::info
+The calculations used to determine whether findings are new are subject to change at any time as Semgrep improves its deduplication logic.
+:::
+
+### `syntactic_id`
```
pangngan, line 32: I would +1 to Alexis's comment to talk about `match_based_id` first. While we do still expose `syntactic_id` as a field in the response of the public API endpoint to retrieve a list of issues, and this is still used on the engine to do some sort of deduplication there (this is totally separate from the platform-side deduplication just to be clear), I think for the user, this information might not be as important/relevant as `match_based_id`. 

Even for us internally, it's mostly used to indicate if a finding closer to an _exact_ match to another finding, but we still use `match_based_id` as the source of truth for platform side deduplication.

docs/semgrep-code/remove-duplicates.md
pangngan: yeah, I don't have a real-life example handy, but one of the benefits we got from changing to `match_based_id` was there code comments no longer affected the calculation of the hash since we'll strip it because hashing the information.

docs/semgrep-code/remove-duplicates.md
pangngan: I think I would agree with Alexis's comment here as well (sorry I know I mentioned that most of the information in the doc was fine to expose). I think we can most likely just push `match_based_id` as the hash that we use to deduplicate while `syntactic_id` is another hash that we use more for internal debugging purposes because we don't store code (of course unless the user gives us code access to view their code).

docs/semgrep-code/remove-duplicates.md
khorne3: Sorry, the identical index was a typo!

docs/semgrep-code/remove-duplicates.md
```diff
@@ -3,29 +3,111 @@ slug: remove-duplicates
...
+
+The rule pattern identified and used in generating the `match_based_id` does not change:
+
+```python
+spcd.get("foo")
+...
+sink("foo")
+```
+
+This means that the `match_based_id` itself doesn't change, allowing Semgrep to identify that the two findings are the same, allowing deduplication. Furthermore, this process enables Semgrep to ignore lines that do not impact code function. For example, adding comments to a code file would create a new `syntactic_id`, but not a new `match_based_id`.
```
s-santillan, line 85: > This means that the `match_based_id` itself doesn't change, **allowing** Semgrep to identify that the two findings are the same, **allowing** deduplication.

I am usually pro repetition in technical writing because it is not creative writing but this one seemed off - maybe we can try:

This means that the `match_based_id` itself doesn't change, allowing Semgrep to identify that the two findings are the same, thus deduplicating them.

docs/semgrep-code/remove-duplicates.md
```diff
@@ -3,29 +3,111 @@ slug: remove-duplicates
...
+### `syntactic_id`
+
+Semgrep generates the `syntactic_id` for a finding using the following information:
+
+- The file path
+- The name of the rule that generated the finding
+- The code syntax, or the literal piece of code that matched the rule
+- The index, a value generated by determining the number of times the rule involved matched code in the file
+
+This information is combined and then hashed before being stored for privacy.
```
s-santillan, line 96:
```suggestion
This information is combined and then hashed for privacy before being stored.
```

"being stored for privacy" - the modifier "for privacy" was too far from the "hash", it could be interpreted two ways (hashed for privacy, or stored for privacy) at first read, so I am suggesting to put "for privacy" nearer to "hash".

s-santillan (approved):

docs/semgrep-code/remove-duplicates.md
```diff
@@ -3,29 +3,111 @@ slug: remove-duplicates
...
+
+Semgrep generates the `match_based_id`, the primary fingerprint used to track and deduplicate findings, for a finding using the following information:
+
+- The file path
+- The name of the rule that generated the finding
+- The rule pattern with the metavariables' values substituted in
+
+This information is combined and then hashed. At this point, Semgrep appends the **index**, a value generated by determining the number of times the rule involved matched code in the file. Note that the index is appended to the hash, not combined with the other finding information before hashing. This is done to preserve information on how findings are related. For example, `finding0` with `match_based_id = 123_0` and `finding1` with `match_based_id = 123_1` indicate that both were generated from the same rule matching the same code pattern in the same file.
+
+Because Semgrep uses the rule pattern instead of the literal code syntax to generate `match_based_id`, code changes that don't impact the code pattern matched also don't hinder Semgrep's ability to recognize that the finding isn't a duplicate of an existing finding.
```
armchairlinguist, line 41: This is definitely much closer, but our usage of "rule pattern" and "code pattern" here is a bit inconsistent. The content of the code could still be different here, but when abstracted to the level of the match with the substituted values then that's what's the same. But this ischallenging because "rule pattern with the metavariables' values [from the code] substituted in" (which is the actual thing that's the same) is cumbersome to say.
@pangngan do you have ideas about how best to express this?

docs/semgrep-code/remove-duplicates.md
armchairlinguist: I might even go for " allowing Semgrep to identify that the two findings are the same, and deduplicate them"

docs/semgrep-code/remove-duplicates.md
```diff
@@ -3,29 +3,111 @@ slug: remove-duplicates
...
+
+- The file path
+- The name of the rule that generated the finding
+- The code syntax, or the literal piece of code that matched the rule
+- The index, a value generated by determining the number of times the rule involved matched code in the file
+
+This information is combined and then hashed before being stored for privacy.
+
+Using the `syntactic_id`, Semgrep can determine if a given finding in a file is 
+the same as a finding identified during a different scan, even if the code snippet that the rule matched had been moved to a different location in the file. This allows Semgrep to avoid generating a new finding and deduplicating its records accordingly, even across multiple branches associated with the project. It also means that Semgrep can cross-correlate findings, so a finding that has been triaged in one branch will be flagged as triaged if it's identified in another branch.
```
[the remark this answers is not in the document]
armchairlinguist: This is actually still not quite right I think. We use the match-based ID for triage checks (I know because I spend a lot of time looking at the configs and results exchanged by the platform and the CLI). @pangngan what do you think about how much we should share about syntactic_id? I know there are cases where we use it in the CLI but I've never understood entirely why/how. Maybe it's enough to say that it's an older ID and is used as a fallback or for internal debugging?

docs/semgrep-code/remove-duplicates.md
```diff
@@ -3,29 +3,111 @@ slug: remove-duplicates
...
+
+If a finding is fixed in one branch (such as `main`) but open in another (such as `production`), and the code fixes are present in both branches, initiate scans through your CI job or SCM tool on the branches with open findings. Semgrep will reconcile the findings and mark them as fixed.
 
 ## Remove duplicate findings using Semgrep API
 
-Semgrep API does not automatically deduplicate findings. If you are using Semgrep API to receive or pull findings data, set the `dedup` flag to `true` to deduplicate findings across refs or branches. Refer to [List all findings](https://semgrep.dev/api/v1/docs/#tag/Finding/operation/semgrep_app.saas.handlers.issue.openapi_list_recent_issues) in the Semgrep API docs for more information.
+Semgrep API does not automatically deduplicate findings. If you use Semgrep API to receive or pull findings data, set the `dedup` flag to `true` to deduplicate findings across refs or branches. Refer to [List all findings](https://semgrep.dev/api/v1/docs/#tag/Finding/operation/semgrep_app.saas.handlers.issue.openapi_list_recent_issues) in the Semgrep API docs for more information.
```
armchairlinguist, line 113: In the context of the rest of this information, this feels kind of confusing. I wonder if we could instead say "does not automatically group findings with the same match-based ID across branches"? I know we call the parameter dedup though, so maybe it's okay even if it's slightly weird language.

docs/semgrep-code/remove-duplicates.md
khorne3: @armchairlinguist I don't know if this is any better, but

> Semgrep uses the rule pattern with the metavariables' values, which originate from the code itself, substituted in to generate `match_based_id`. Therefore, code changes that result in the same rule pattern when abstracted to the level of the rule pattern match, with the appropriate values substituted, in don't hinder Semgrep's ability to recognize that the finding isn't a duplicate of an existing finding.

docs/semgrep-code/remove-duplicates.md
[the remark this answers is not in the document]
khorne3: You're right -- I think this makes more sense for match ID so I'll change it out and move it. I think the note on why we have syntactic_id should be sufficient

docs/semgrep-code/remove-duplicates.md
```diff
@@ -3,29 +3,110 @@ slug: remove-duplicates
...
+
+:::info
+The `syntactic_id` is primarily used by Semgrep for internal debugging purposes, since no code is stored except in cases where you have provided code access permissions to Semgrep.
+:::
+
+## Update findings by rescanning the project
+
+Semgrep's correlation of findings across branches based on their unique fingerprint allows for automatic consolidation of findings and makes it simpler to triage findings.
+
+If a finding is fixed in one branch (such as `main`) but open in another (such as `production`), and the code fixes are present in both branches, initiate scans through your CI job or SCM tool on the branches with open findings. Semgrep will reconcile the findings and mark them as fixed.
```
pangngan, line 108: I wonder if it would be helpful here to point out that the finding remains open in production most likely because a follow-up scan hasn't been triggered. I think it's implied, but given that readers may not be as familiar with how the logic works, it might be worth calling that out a little more explicitly here?

pangngan (approved):
Thank you so much for all the changes! This is definitely a really complicated topic even for internal folks to grasp, but I think the doc has landed in a place that will really help external folks understand even a little bit more how this logic works behind the scenes! 🙇‍♀️

PR #1854, opened 2024-11-19
Clarify default behavior vs options for handling exit codes

armchairlinguist:
This paragraph was confusing because it suggested that the default result for an internal error would be 2 when it's actually 0. The content below is correct so refer to that instead.

### Please ensure

### DIFF
--- a/docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
+++ b/docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
@@ -26,7 +26,7 @@ When Semgrep identifies one or more blocking findings, it returns exit code `1`.
 
 The process to enforce a block on a PR or MR after Semgrep exits with error code `1` is dependent on your CI provider. Review your CI provider's documentation for further information.
 
-If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server and returns exit code `2`. Review your CI provider's documentation for further information on how to handle this error code, such as failing the pipeline job. 
+If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server. By default, it returns exit code 0 in CI. If you would prefer to catch internal errors, review the options below and the Semgrep exit codes reference to determine how you want to handle each exit code.
 
 :::info
 See [CLI reference](/cli-reference#exit-codes) for more information about Semgrep exit codes.
--- a/docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
+++ b/docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
@@ -26,11 +26,7 @@ When Semgrep identifies one or more blocking findings, it returns exit code `1`.
 
 The process to enforce a block on a PR or MR after Semgrep exits with error code `1` is dependent on your CI provider. Review your CI provider's documentation for further information.
 
-If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server. By default, it returns exit code 0 in CI. If you would prefer to catch internal errors, review the options below and the Semgrep exit codes reference to determine how you want to handle each exit code.
-
-:::info
-See [CLI reference](/cli-reference#exit-codes) for more information about Semgrep exit codes.
-:::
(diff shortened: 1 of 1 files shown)


s-santillan:
Huh so... it looks like `semgrep ci` and `semgrep` (or `semgrep scan`) give different exit codes for internal errors! Here is one in OSS: https://github.com/semgrep/semgrep/issues/6631 (I referenced this link when Katie touched this doc 2 weeks ago in [PR 1820](https://github.com/semgrep/semgrep-docs/pull/1820#discussion_r1827324535))

armchairlinguist:
@s-santillan that's correct - it's one of the behavior differences between them. As noted in the content below the para I changed, the default behavior for `semgrep ci` is to suppress errors other than for blocking findings. That's also a pretty old GH issue you're referencing, but I don't think that `scan` has changed behavior. IIUC the goal for CI is not to block the pipeline because of Semgrep failures, since that puts developers in a bad position of being blocked for things they can't control.

docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
```diff
@@ -26,7 +26,7 @@ When Semgrep identifies one or more blocking findings, it returns exit code `1`.
 
 The process to enforce a block on a PR or MR after Semgrep exits with error code `1` is dependent on your CI provider. Review your CI provider's documentation for further information.
 
-If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server and returns exit code `2`. Review your CI provider's documentation for further information on how to handle this error code, such as failing the pipeline job. 
+If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server. By default, it returns exit code 0 in CI. If you would prefer to catch internal errors, review the options below and the Semgrep exit codes reference to determine how you want to handle each exit code.
```
khorne3, line 29:
```suggestion
If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server. By default, it returns exit code 0 in CI. If you would prefer to catch internal errors, review the options below and the Semgrep exit codes reference to determine how you want to handle each exit code.
```

Maybe something like "If Semgrep encounters an internal error, it sends an anonymous crash report to a crash-reporting server and returns exit code `0`. If you want to catch internal errors, review the [CLI reference](/cli-reference#exit-codes) for more information about Semgrep's exit codes and the options explained in this article to determine how you want to handle each exit code." and then remove the admonition that immediately follows since you're linking to the CLI docs in the text itself

docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
```diff
@@ -26,7 +26,7 @@ When Semgrep identifies one or more blocking findings, it returns exit code `1`.
...
 
 :::info
```
khorne3, line 31:
```suggestion

```

docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
```diff
@@ -26,7 +26,7 @@ When Semgrep identifies one or more blocking findings, it returns exit code `1`.
...
 See [CLI reference](/cli-reference#exit-codes) for more information about Semgrep exit codes.
```
khorne3, line 32:
```suggestion

```

docs/semgrep-ci/configuring-blocking-and-errors-in-ci.md
armchairlinguist: That makes sense to me - are you open to making the changes? Or should I just commit all the suggestions to get that result? (Sorry, brain is a little fried this week.)

khorne3: No worries -- my brain is mushy this week, too. I'm sure there's a better way to suggest multi-line changes, but I can't figure it out today. I'll push a commit with my suggestions

khorne3 (approved):

armchairlinguist:
thanks @khorne3 ! that looks great

PR #1855, opened 2024-11-19
Clarify difference between code links

relsqui:
The git hash and line of code link in findings details go to different places. This adds a couple of bullet points about what those places are.

AFAIK the same thing is true for Secrets and SSC but I didn't see a parallel page for either of them where we could make this same change.

### Please ensure


armchairlinguist (approved):
I like this as-is, since there's some oddities on the SSC side. Thanks!

PR #1864, opened 2024-12-04
convert SSC admonition to stand-alone section

khorne3:
[Context](https://semgrepinc.slack.com/archives/C032JB2HRDK/p1733134144372029)

### Please ensure


s-santillan (approved):
FWIW I agree with leaving the body/content as-is, btw, as it gives us some leeway given that some rules of extremely high severity do get a reachability rule out within 24 hrs, in addition to parity rules within 24 hrs.

PR #1879, opened 2024-12-13
Add caveat about hosted runners for Azure

armchairlinguist:
We test our sample config on Azure hosted runners; this adds a note to let folks using self-hosted runners know they might need to adjust. It's planned to be followed by adding KBs with examples of how this could work for the default *nix self-hosted image.

### Please ensure

~- [ ] A subject matter expert (SME) reviews the content~ xs PR

### DIFF
--- a/docs/semgrep-ci/sample-ci-configs.md
+++ b/docs/semgrep-ci/sample-ci-configs.md
@@ -440,6 +440,8 @@ To add Semgrep into Azure Pipelines:
 
 ### Sample Azure Pipelines configuration snippet
 
+This configuration snippet is tested with hosted Azure runners. If you are using self-hosted runners, you may need to make adjustments to ensure that the necessary software is available.
+
 <Tabs
     defaultValue="azure-semgrep"
     values={[


docs/semgrep-ci/sample-ci-configs.md
```diff
@@ -440,6 +440,8 @@ To add Semgrep into Azure Pipelines:
 
 ### Sample Azure Pipelines configuration snippet
 
+This configuration snippet is tested with hosted Azure runners. If you are using self-hosted runners, you may need to make adjustments to ensure that the necessary software is available.
```
khorne3, line 443:
```suggestion
This configuration snippet is tested with hosted Azure runners. If you are using self-hosted runners, you may need to make adjustments to ensure that the necessary dependencies is available.
```

I don't have a strong preference, but maybe dependencies is more descriptive?

khorne3 (approved):

docs/semgrep-ci/sample-ci-configs.md
armchairlinguist: I don't know that it's really more helpful here - I'm going to go with merging and we can revisit later if needed.

PR #1881, opened 2024-12-17
remove references to lockfile-only

khorne3:

### Please ensure


docs/semgrep-supply-chain/glossary.md
```diff
@@ -30,9 +30,9 @@ A lockfile describes a dependency tree to ensure that deployments and organizati
 
 Semgrep Supply Chain uses lockfiles as part of its analysis to determine the exact version of a dependency that a codebase is using.
 
-## Lockfile-only rules
+## Rules without reachability analysis
```
s-santillan, line 33:
```suggestion
## No-reachability rules
```
Suggestion to at least make it a term here in the glossary.

The hyphen is added between "no" and "reachability" to mean that they are one word. Without the hyphen it becomes "no reachability rules", which is a different meaning entirely.

docs/semgrep-supply-chain/glossary.md
```diff
@@ -68,7 +68,7 @@ See [Overview of Semgrep Supply Chain](/semgrep-supply-chain/overview) to learn
 
 A type of Semgrep Supply Chain rule that performs reachability analysis. A reachability rule can determine if the vulnerable piece of code from a dependency is used in the codebase that imports it.
 
-Compare its opposite: [Lockfile-only rules](#lockfile-only-rules).
+Compare its opposite: [rules without reachability analysis](#rules-without-reachability-analysis)
```
s-santillan, line 71: If you accept the above comment about **No-reachability rules** then update this header as well.

docs/supported-languages.md
```diff
@@ -227,29 +227,29 @@ Additionally, Semgrep offers beta support for the scanning of Java projects **wi
    <td>Rust</td>
    <td>Cargo*</td>
    <td><code>cargo.lock</code></td>
-   <td style={{"text-align": "center"}}>Lockfile-only</td>
+   <td style={{"text-align": "center"}}>--</td>
```
s-santillan, line 230: The `--` makes it look like we don't have rules at all?

docs/semgrep-supply-chain/glossary.md
s-santillan: Other ideas:

- "Version-only" or "Version" rules (we literally compare if the library has a vulnerable version)
- "Dependency-only" or "Dependency"

docs/supported-languages.md
s-santillan: If we do have an alternative term, we can update this table to use that term.

docs/semgrep-supply-chain/glossary.md
s-santillan: Looks like the decision for it has been made, so I'll resolve this

docs/supported-languages.md
s-santillan: Resolving as "won't fix"

s-santillan (approved):

PR #1886, opened 2024-12-18
Small changes to export field docs for accuracy on 1.101.0

milanwilliams:
update the json/sarif export fields

### Please ensure


s-santillan:
@milanwilliams Do we have some fields that are locked to AppSec Platform within the sarif `rules` object? Just checking because maybe we should expand the `rules` object if so?

s-santillan (approved):

milanwilliams:
> @milanwilliams Do we have some fields that are locked to AppSec Platform within the sarif `rules` object? Just checking because maybe we should expand the `rules` object if so?

Right now, there's no fields that are locked under rules, so it's good as-is for now!

PR #1887, opened 2024-12-19
[KB]: Add self hosted config options for Azure Pipelines + Ubuntu

armchairlinguist:
We received a report that the sample CI config didn't work for self-hosted runners, and a caveat has been added to that config. This guide supplements that config by providing guidance for the default type of Linux self-hosted runner (Ubuntu).

### Please ensure


docs/semgrep-ci/sample-ci-configs.md
```diff
@@ -440,7 +440,7 @@ To add Semgrep into Azure Pipelines:
 
 ### Sample Azure Pipelines configuration snippet
 
-This configuration snippet is tested with hosted Azure runners. If you are using self-hosted runners, you may need to make adjustments to ensure that the necessary software is available.
+This configuration snippet is tested with hosted Azure runners. If you are using self-hosted runners, you may need to make adjustments to ensure that the necessary software is available. Consult [Semgrep with self-hosted Ubuntu runners in Azure Pipelines](/docs/kb/semgrep-ci/azure-self-hosted-ubuntu) for two recommended options.
```
s-santillan, line 443:
```suggestion
This configuration snippet is tested with **hosted** Azure runners. If you are using self-hosted runners, you may need to make adjustments to ensure that the necessary software is available. Consult [Semgrep with self-hosted Ubuntu runners in Azure Pipelines](/docs/kb/semgrep-ci/azure-self-hosted-ubuntu) for two recommended options.
```

extremely minor suggestion

docs/kb/semgrep-ci/azure-self-hosted-ubuntu.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Run a scan
+
+On adding or editing the `azure-pipelines.yml` file, the Semgrep scan job should run on the desired runner pool. Verify that the scan is successful.
+
+## Using pip with a virtual environment
+
+### Prepare your runner
+
+This approach uses built-in Azure DevOps tasks, including `UsePythonVersion` and `Bash`, and uses a virtual environment to install pip, another approach that prevents issues with system-managed Python vs user-installed Python.
```
s-santillan, line 80:
```suggestion
This approach uses built-in Azure DevOps tasks, including `UsePythonVersion` and `Bash`, and uses a virtual environment to install `pip`, another approach that prevents issues with system-managed Python vs user-installed Python.
```

s-santillan (approved):
Thank you as always for this addition to the docs! I really should make the generated index pages look nice so that we can advertise the KB better especially now that we have our own card designs...

docs/kb/semgrep-ci/azure-self-hosted-ubuntu.md
```diff
@@ -0,0 +1,130 @@
...
+    if [ $(Build.SourceBranchName) = "master" ]; then
+        echo "Semgrep full scan"
+        semgrep ci
+    elif [ $(System.PullRequest.PullRequestId) -ge 0 ]; then
+        echo "Semgrep diff scan"
+        git fetch origin master:origin/master
+        export SEMGREP_PR_ID=$(System.PullRequest.PullRequestId)
+        export SEMGREP_BASELINE_REF='origin/master'
+        semgrep ci
+    fi
```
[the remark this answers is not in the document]
armchairlinguist: So the way this is set up right now uses the `Semgrep_Variables` approach that we use in the sample config (and refers back there). If we want to generally change our approach to use the app token as a solo variable without a group (like we've been discussing internally), then we should do that here too, but otherwise I'd prefer to follow our sample config as much as possible.

docs/kb/semgrep-ci/azure-self-hosted-ubuntu.md
```diff
@@ -0,0 +1,131 @@
...
+  env:
+    SEMGREP_APP_TOKEN: $(SEMGREP_APP_TOKEN)
+```
+
```
armchairlinguist, line 66: This is basically where the problem is (and the equivalent place for the second config) - in order for these configs to work, the user would need to also follow the instructions on line 29+ of `src/components/code_snippets/_azure-semgrep-app-sast.mdx`, but it's difficult to refer to on its own.

s-santillan: Maybe we should also snippetify those instructions so that the user doesn't have to leave the KB article?

armchairlinguist: @s-santillan I gave this a shot, if you don't mind taking another look! Sorry for the delay, and I hope I put the snippet in the right place.

s-santillan: I checked the [preview](https://deploy-preview-1887--semgrep-docs-prod.netlify.app/docs/kb/semgrep-ci/azure-self-hosted-ubuntu) and it looks good! But in the sample CI configs looks like there's a [duplicate](https://deploy-preview-1887--semgrep-docs-prod.netlify.app/docs/semgrep-ci/sample-ci-configs#setting-environment-variables-in-azure-pipelines)?

armchairlinguist: Oh, thank you. I probably just forgot to remove the old one when I added the snippet, I'll get that fixed up.

armchairlinguist: Should be fixed now!

s-santillan (approved):
I think I forgot to stamp this! Sorry about that.

armchairlinguist:
@s-santillan no worries, I feel bad I had to ask for review so many times, it turned out more complicated than I thought.

PR #1889, opened 2025-01-03
[KB] Customize Semgrep in pre-commit

armchairlinguist:
I've fielded a few questions about ways to modify Semgrep's behavior in pre-commit, so this guide offers both a couple of specific examples and some more general notes about how the existing hooks work that I hope will help folks self-serve more generally.

I added a link to the related doc, and Vale was bugging me, so I also cleaned it up a bit.

### Please ensure

### DIFF
--- a/docs/kb/integrations/customize-semgrep-precommit.md
+++ b/docs/kb/integrations/customize-semgrep-precommit.md
@@ -45,7 +45,7 @@ Semgrep Secrets is an ideal product to run before commit, since it can help prev
       args: ["ci", "--dry-run", "--baseline-commit", "HEAD", "--secrets"]
 ```
 
-## Scan with Pro Engine
+## Scan with Pro rules and cross-function analysis
 
 The `semgrep-ci` hook requires the user to be logged in locally and runs with the engine configured in the organization, but the standard `semgrep` hook can also take advantage of local login to run with Pro rules and cross-function analysis.
 
--- a/docs/kb/integrations/customize-semgrep-precommit.md
+++ b/docs/kb/integrations/customize-semgrep-precommit.md
@@ -47,7 +47,7 @@ Semgrep Secrets is an ideal product to run before commit, since it can help prev
 
 ## Scan with Pro Engine
 
-The `semgrep-ci` hook requires the user to be logged in locally and runs with the engine configured in the organization, but the standard `semgrep` hook can also take advantage of local login to run with the Pro Engine.
+The `semgrep-ci` hook requires the user to be logged in locally and runs with the engine configured in the organization, but the standard `semgrep` hook can also take advantage of local login to run with Pro rules and cross-function analysis.
 
 ```yaml
 - id: semgrep
--- a/docs/kb/integrations/customize-semgrep-precommit.md
+++ b/docs/kb/integrations/customize-semgrep-precommit.md
@@ -1,0 +1,63 @@
+---
+description: Understand how to customize Semgrep's behavior when using it with pre-commit.
+tags:
+  - Extensions
+---
+
+# Customize Semgrep in pre-commit
+
+Semgrep offers two standard pre-commit hooks [outlined here](/docs/extensions/overview/#pre-commit).
+
+* The `semgrep` hook is optimized to run with a provided set of rules, either local rules or rules from the Semgrep Registry.
+* The `semgrep-ci` hook is optimized to run with the rules from your organization in the Semgrep AppSec Platform, without sending results to the platform, since results from pre-commit are temporary and not linked to the final code in the repository.
+
+You can also create your own Semgrep hooks for pre-commit if you have particular needs or preferences, such as if you want users to see the output of the Semgrep scan even if it passed, or if you want to limit the scan to a particular Semgrep product.
+
+## Show the scan output
+
+By default, pre-commit does not show the scan output if the hook passes. If you're using the `semgrep-ci` hook, but want your developers to see the output even if non-blocking findings are found, you can set the `verbose` option to print the output. To prevent the output from being too chatty, redirect `stderr` to `/dev/null` so that only the findings are printed, or use `--quiet`.
+
+```yaml
+repos:
+- repo: https://github.com/semgrep/pre-commit
+  rev: 'v1.101.0'
+  hooks:
+    - id: semgrep-verbose
+      entry: semgrep
+      args: ["ci", "--dry-run", "--baseline-commit", "HEAD" "2>/dev/null"]
+      verbose: true
+      pass_filenames: false
+```
+
+## Limit the scan to a particular product
+
+Semgrep Secrets is an ideal product to run before commit, since it can help prevent secrets from ever making it into the Git history, even locally. To run only Secrets in pre-commit, add the product flag to the `args`:
+
+```yaml
+- repo: https://github.com/semgrep/pre-commit
+  rev: 'v1.101.0'
+  hooks:
+    - id:  semgrep-secrets
+      pass_filenames: false
+      args: ["ci", "--dry-run", "--baseline-commit", "HEAD", "--secrets"]
+```
+
+## Customization tips
+
+### Useful arguments
+
+The provided hooks include useful arguments for Semgrep that you may want to include in your hooks as well.
+
+For example, for the `semgrep` hook that runs with a provided `--config`, the arguments also include `--disable-version-check` (skips checking whether there's a new version of Semgrep, speeding things up and avoiding irrelevant output), `--quiet` (only print findings, not other messages) and `--skip-unknown-extensions` (if files are modified that aren't in a recognized language, just skip them).
+
+The `semgrep-ci` hook runs with the pre-commit option `pass_filenames: false` because `semgrep ci` doesn't expect a list of filenames (which is what pre-commit normally provides); it just expects to scan the folder. Adding `baseline-commit HEAD` performs a diff scan; this diff scan only scans the changes being added in the current commit (as intended for a pre-commit hook). It uses `--dry-run` so that findings that have no existence except in the local filesystem aren't added to the platform, which would otherwise result in clutter.
+
+### Entry point
+
+Both of the default hooks use `semgrep` as the `entry`, which means the command executed is just `semgrep` with the `args` provided. However, you can write your own `entry` point:
+
+```bash
+ entry: sh -c 'env SEMGREP_REPO_NAME=pre-commit-diff SEMGREP_BRANCH=$(git branch --show-current) semgrep ci --code --no-suppress-errors --baseline-commit HEAD'
+```
+
+This lets you set environment variables (like `SEMGREP_REPO_NAME`) that can’t be set as command-line arguments.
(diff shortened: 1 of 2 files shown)


docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -0,0 +1,71 @@
...
+Semgrep offers two standard pre-commit hooks [outlined here](/docs/extensions/overview/#pre-commit).
+
+* The `semgrep` hook is optimized to run with a provided set of rules, either local rules or rules from the Semgrep Registry.
+* The `semgrep-ci` hook is optimized to run with the rules from your organization in the Semgrep AppSec Platform, without sending results to the platform, since results from pre-commit are temporary and not linked to the final code in the repository.
+
+You can also create your own Semgrep hooks for pre-commit if you have particular needs or preferences, such as if you want users to see the output of the Semgrep scan even if it passed, or if you want to limit the scan to a particular Semgrep product.
+
+## Show the scan output
+
+By default, pre-commit does not show the scan output if the hook passes. If you're using the `semgrep-ci` hook, but want your developers to see the output even if non-blocking findings are found, you can set the `verbose` option to print the output. To prevent the output from being too chatty, redirect `stderr` to `/dev/null` so that only the findings are printed, or use `--quiet`.
```
s-santillan, line 18:
```suggestion
By default, pre-commit does not show the scan output if the hook passes. If you're using the `semgrep-ci` hook, but want your developers to see the output even if non-blocking findings are found, you can set the `verbose` option to print the output. To prevent the output from being too chatty, redirect `stderr` to `/dev/null` so that only the findings are printed, or use `--quiet`.
```
Extremely minor nit, Instead of "too chatty" do we want to mention that verbose prints out errors too, then provide the redirection?
Something like "The `verbose` option logs error messages as well. You can reduce the logs by redirecting `stderr`..."

s-santillan (approved):
Your first (of many) PR(s) for the year!

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: That rephrase wouldn't be quite right here, since we print out all scan progress (diagnostic) information on stderr, not just errors. Findings are the only thing that goes to stdout. But I can choose a more neutral phrase for sure :)

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -0,0 +1,71 @@
...
+tags:
+  - Extensions
+---
+
+# Customize Semgrep in pre-commit
+
+Semgrep offers two standard pre-commit hooks [outlined here](/docs/extensions/overview/#pre-commit).
+
+* The `semgrep` hook is optimized to run with a provided set of rules, either local rules or rules from the Semgrep Registry.
+* The `semgrep-ci` hook is optimized to run with the rules from your organization in the Semgrep AppSec Platform, without sending results to the platform, since results from pre-commit are temporary and not linked to the final code in the repository.
```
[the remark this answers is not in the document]
armchairlinguist: Credit to pad on that I think!

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: OK, updated :) Good call.

ajbt200128:
@armchairlinguist didn't realize we had

> To implement a custom extension, wrap `semgrep lsp` to start the Semgrep Language Server, which communicates over `stdio`. Alternatively, this protocol is a great way to integrate Semgrep into a project, as it can perform incremental scans, and caches various computations to increase performance. Please let us know in our [Community Slack](https://go.semgrep.dev/slack) if you use this approach; we'd be happy to help in any way.

in our docs. We are no longer licensing semgrep in a way that we encourage this behavior (and eventually the LSP will be moved to pro to enforce this). Do you mind making another PR to remove this bit?

docs/kb/integrations/customize-semgrep-precommit.md
ajbt200128, line 12: wonder if it's worth calling out that this is near identical to what happens when semgrep runs in CI on a PR specifically?

ajbt200128: Also, maybe we want to mention something about this requiring login?

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -0,0 +1,71 @@
...
+
+You can also create your own Semgrep hooks for pre-commit if you have particular needs or preferences, such as if you want users to see the output of the Semgrep scan even if it passed, or if you want to limit the scan to a particular Semgrep product.
+
+## Show the scan output
+
+By default, pre-commit does not show the scan output if the hook passes. If you're using the `semgrep-ci` hook, but want your developers to see the output even if non-blocking findings are found, you can set the `verbose` option to print the output. To limit output to findings only (no scan information or diagnostics), redirect `stderr` to `/dev/null`, or use `--quiet`.
```
ajbt200128, line 18: what if I'm just using the semgrep hook, not semgrep-ci? Do I see the output then?

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -0,0 +1,71 @@
...
+      args: ["ci", "--dry-run", "--baseline-commit", "HEAD", "--secrets"]
+```
+
+## Customization tips
+
+### Useful arguments
+
+The provided hooks include useful arguments for Semgrep that you may want to include in your hooks as well.
+
+For example, for the `semgrep` hook that runs with a provided `--config`, the arguments also include:
```
ajbt200128, line 51: we probably want --pro as a useful argument too

ajbt200128 (approved):
cc @kopecs since you've also touched this stuff and might have other info to add

kopecs:
I'll take a look, but none of my recent changes have actually changed the functionality of the pre-commit hooks.

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: This is just written as an example, but I'll make that clearer in the language.

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: This text is meant to cover "hey some useful arguments are used already, I'm going to explain them", not to introduce other arguments. But I can add an example or a note about `--pro` separately.

armchairlinguist:
> We are no longer licensing semgrep in a way that we encourage this behavior (and eventually the LSP will be moved to pro to enforce this). Do you mind making another PR to remove this bit?

I'll just modify this PR to remove it, since I'm editing that section.

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: Plus on the login, minus on the "what happens when Semgrep runs on a PR" - it's actually significantly different from that since PR scans compare head and base, and pre-commit scans compare per-commit. We've had folks think that PR scans do per-commit analysis before and be confused that they don't, so I don't want to encourage the confusion.

armchairlinguist:
@s-santillan I made some changes where I'm not 100% on whether I used the right style, mind giving this another look?

docs/kb/integrations/customize-semgrep-precommit.md
ajbt200128: gotcha!

s-santillan:
@armchairlinguist will do first thing tomorrow! Thanks for the heads up.

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -0,0 +1,87 @@
...
+```yaml
+- repo: https://github.com/semgrep/pre-commit
+  rev: 'v1.101.0'
+  hooks:
+    - id:  semgrep-secrets
+      pass_filenames: false
+      args: ["ci", "--dry-run", "--baseline-commit", "HEAD", "--secrets"]
+```
+
+## Scan with Pro Engine
```
s-santillan, line 48:
```suggestion
## Scan with interfile analysis
```
From a marketing/product standpoint, we no longer use the term "Pro Engine" externally. (The only "Pro" product we have is "Pro rules".) We try to call it "interfile analysis" where possible.

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -0,0 +1,87 @@
...
+
+The `semgrep-ci` hook requires the user to be logged in locally and runs with the engine configured in the organization, but the standard `semgrep` hook can also take advantage of local login to run with the Pro Engine.
```
s-santillan, line 50:
```suggestion
The `semgrep-ci` hook requires the user to be logged in locally and runs with the engine configured in the organization, but the standard `semgrep` hook can also take advantage of local login to run with interfile analysis.
```

s-santillan (approved):
Reads nicely to me! Btw, parentheses are okay (it's just a "warning" level I think in Vale), for the TW team, we think google is too strict about parentheses. Sometimes it just makes sense to use them.

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: Unfortunately that wouldn't be right here, since diff-aware scans (scans with a baseline ref) don't run interfile analysis. Maybe just "Scan with Pro" then?

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: also wrong for the same reason, sadly. I don't have a ready suggestion here.

docs/kb/integrations/customize-semgrep-precommit.md
s-santillan: Hmm... When you say scan with Pro, this is the binary, and does NOT affect the Pro rules, right? Do we do interprocedural for diff scans? (my last knowledge was "this is in beta" but I could be wrong). We can say "cross-function analysis" if it's already in GA.

armchairlinguist: yeah, diff scan is intrafile - the beta was for interfile, and we killed it :rip:

armchairlinguist: I think this would apply to rules - let me check, though. The reason I think it will is that any token should unlock feature available to users of the platform, which includes pro rules.

armchairlinguist: OK yeah - this also gave me a better idea of what to write here, based on what the CLI prints - if you're logged in, it says "Semgrep Code", and that might be a good choice for this purpose. Thoughts @s-santillan ?

s-santillan: Yes, we can use "Scan with Semgrep Code" or "Scan with Pro rules and cross-function analysis". If we go with "Semgrep Code" we can just mention what that entails (the Pro rules and cross-function analysis) in the `<p>`.

docs/kb/integrations/customize-semgrep-precommit.md
s-santillan:
```suggestion
The `semgrep-ci` hook requires the user to be logged in locally and runs with the engine configured in the organization, but the standard `semgrep` hook can also take advantage of local login to run with Pro rules and cross-function analysis.
```
Does this work?

armchairlinguist: yeah, I think that's fine and probably clearer than Semgrep Code! Thanks!

PR #1893, opened 2025-01-08
create a new policies page

s-santillan:

### Please ensure

### DIFF
--- a/docs/semgrep-supply-chain/policies.md
+++ b/docs/semgrep-supply-chain/policies.md
@@ -57,14 +57,14 @@ After creating a policy, it is automatically **enabled** and will be applied to 
 1. Make your changes.
 1. Click **Save**.
 
-### Disable a policy
+### Disable or enable a policy
 
-From the Supply Chain policies tab, turn off the toggle for the policy you want to edit.
+From the Supply Chain policies tab, click the toggle for the policy you want to edit.
 
-You can also disable a policy from the policy's page:
+You can also disable or enable a policy from the policy's page:
 
 1. From the Supply Chain policies tab, click the **three-dot (...) button > Edit policy**.
-1. Turn off the **Enable policy** toggle.
+1. Turn off or on the **Enable policy** toggle.
 1. Click **Save**.
 
 ### Delete a policy
--- a/docs/semgrep-supply-chain/policies.md
+++ b/docs/semgrep-supply-chain/policies.md
@@ -45,8 +45,11 @@ Only **admins** can view, create, edit, or delete policies.
   1. For **Project** or **Project tag** values, a second drop-down box appears. Choose the **projects** or **project tags** to finish defining the scope. You can select more than one.
 1. Define the conditions of the policy by selecting either **Reachable** or **Critical or high severity, reachable, upgrades available**. The latter value is a broader set of conditions and typically results in more findings shown to developers.
 1. Define the actions of the policy. You can choose to **Leave a comment** or **Block and leave a comment**.
-1. Ensure that the **Enable policy** toggle is enabled.
-1. Click **Save**. Your policy is applied in subsequent scans.
+1. Click **Save**. This brings you back to the Supply Chain policies tab.
+
+## Enable a policy
+
+Creating a policy does **not** enable it by default. To enable it, click the **toggle** on the policy's entry in the Supply Chain policies tab.
 
 ## Other operations
 
--- a/docs/semgrep-supply-chain/policies.md
+++ b/docs/semgrep-supply-chain/policies.md
@@ -48,7 +48,7 @@ Only **admins** can view, create, edit, or delete policies.
 
 ### Edit a policy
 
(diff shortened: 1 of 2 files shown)


docs/semgrep-supply-chain/policies.md
```diff
@@ -0,0 +1,74 @@
...
+---
+
+# Manage policies (beta)
+
+Use Supply Chain policies to define the conditions in which developers are **notified** of a finding through a comment, or potentially **blocked** from merging their pull or merge request (PR or MR). This feature helps you manage noise and ensures that developers are only notified or potentially blocked based on the conditions you set.
+
+This feature enables you to configure the following:
+
+- **Scope**: These are the projects (repositories) that are affected by the policy.
+- **Conditions**: The conditions under which **actions** are performed. These conditions are typically related to the value of a finding's severity or reachability.
```
pangngan, line 18: I liked that you included examples of what could be an example of a condition, though I wonder if we could frame it as something like examples of finding attributes that can be defined as a condition. So something like the following maybe?
```suggestion
- **Conditions**: The conditions under which **actions** are performed. These conditions are typically attributes of a finding such as severity or reachability. 
```

docs/semgrep-supply-chain/policies.md
```diff
@@ -0,0 +1,74 @@
...
+1. From the navigation menu, click **Rules** to expand the drop-down box, then click **Policies**.
+1. Click **Supply Chain**. This takes you to the Supply Chain policies tab.
+
+## Create a policy
+
+1. From the Supply Chain policies tab, Click **Create policy**.
+1. Provide a **Policy name**.
+1. Define the scope of the policy:
+  1. Click the drop-down box to select between **All Projects**, **Project**, or **Project tag**. 
+  1. For **Project** or **Project tag** values, a second drop-down box appears. Choose the **projects** or **project tags** to finish defining the scope. You can select more than one.
```
pangngan, line 45: I'm checking out the live preview and it seemed like these may have not been indented properly (though I'm going off of the live preview you linked in your slack message so it might be old). Currently seeing this in the live preview:
image

pangngan, line 45: super nit: "You can select more than one" reads to me that you can select more than one type of value at once, which isn't quite true. I know that you're referring to how we can select more than one project / one project tag at a time, but I wonder if there's an alternate way of phrasing that?

docs/semgrep-supply-chain/policies.md
```diff
@@ -0,0 +1,74 @@
...
+1. Define the conditions of the policy by selecting either **Reachable** or **Critical or high severity, reachable, upgrades available**. The latter value is a broader set of conditions and typically results in more findings shown to developers.
```
pangngan, line 46:
```suggestion
1. Define the conditions of the policy by selecting either **Reachable** or **Critical or high severity, reachable, upgrades available**. The former value is a broader set of conditions and typically results in more findings shown to developers.
```

I think the latter actually defines a more narrow scope of conditions since the finding must be reachable in addition to the rest of the defined conditions.

docs/semgrep-supply-chain/policies.md
s-santillan: Yup, looks like it requires a 4-space indent. Nice catch, thank you.

docs/semgrep-supply-chain/policies.md
```diff
@@ -0,0 +1,74 @@
...
+1. Define the actions of the policy. You can choose to **Leave a comment** or **Block and leave a comment**.
+1. Click **Save**. This brings you back to the Supply Chain policies tab.
+
+## Enable a policy
+
+Creating a policy does **not** enable it by default. To enable it, click the **toggle** on the policy's entry in the Supply Chain policies tab.
```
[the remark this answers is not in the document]
s-santillan: :ACK: thanks to Kevin and Lisa for considering my suggestion and you for keeping me updated. :)

docs/semgrep-supply-chain/policies.md
s-santillan: On second look, "select more than one" is redundant since "projects" and "project tags" are already plural. I have just removed it. The previous step already states that the user must select only one (all projects OR project tags OR project) so that should help as well.

docs/semgrep-supply-chain/policies.md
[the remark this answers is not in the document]
s-santillan: Resolved in latest commit.

s-santillan:
@pangngan Thanks for the review! Addressed your notes. The preview link, btw, is always updated with the latest commits, so you can refresh the page in a few minutes (the `netlify/semgrep-docs-prod/deploy-preview` job just has to finish)

pangngan (approved):
lgtm! thank you for all the edits!!

s-santillan:
Going to merge it for now and we can make subsequent updates to it as the feature develops.

PR #1896, opened 2025-01-08
[KB] Add guide for handling API 404s

armchairlinguist:
I don't expect a ton of folks to read this proactively, but it'll make it easier for us to answer questions, and it'd be great if even a few people used it to self-serve. :)

### Please ensure

### DIFF
--- a/docs/kb/semgrep-appsec-platform/api-404-token-scope.md
+++ b/docs/kb/semgrep-appsec-platform/api-404-token-scope.md
@@ -1,0 +1,23 @@
+---
+tags:
+  - API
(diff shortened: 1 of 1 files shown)


armchairlinguist:
@s-santillan just checking, based on what I see for other pages, none of the fontawesome icons show up on the Netlify preview, so this one being missing shouldn't be a problem either, right? I double-checked which icon it is.

bandarisantosh (approved):
Looks good to me, ty!

s-santillan:
Yeah, on the netlify preview, they don't show up, but they do in live @armchairlinguist 😅

s-santillan (approved):

PR #1921, opened 2025-01-23
add electron as a source (xs sized PR)

s-santillan:

### Please ensure

Tiny PR. I checked our **Advisories** tab and we don't show the source of the CVE (yet) so I just listed electron release notes as a source.


khorne3 (approved):
Do you think we should update the header on the Supported Languages table where it says, "Period of reachability rule coverage for CVEs/GHSAs" too?

s-santillan:
@khorne3 I thought about it - my first pass was "Period of reachability rule coverage for advisory sources" which sounded pretty weird. (We're going to have a third rule source within 1-2 months so I had to generalize to "security advisory sources" or some such.) Also there was no mention of "time period coverage" during the product-updates announcement so I'm not sure we go back to 201X for Electron specifically.

khorne3:
@s-santillan That makes sense. If/when we need to make these updates, maybe we just shorten the header to "Period of reachability rule coverage" and make sure the text is descriptive enough to be clear

PR #1926, opened 2025-01-24
assorted SSO/SAML updates

khorne3:
- added tags to frontmatter + minor edits re: capitalization and spacing
- add admonition re: org owner admin/member accounts to primary SSO page
- moved info re: Microsoft Entra ID to the KB
- added/consolidated links to Google Workspace and Microsoft Entra ID KB docs on the main SSO page

### Please ensure


docs/deployment/sso.md
```diff
@@ -66,85 +63,12 @@ SAML2.0 is configured through **Semgrep AppSec Platform**. To set up SSO:
...
-
-The following references are used:
-
--->
-
-:::info Prerequisites
-* An existing Microsoft Entra ID account.
-* Sufficient permissions within Microsoft Entra ID to create enterprise apps. See [Microsoft Entra ID roles](https://learn.microsoft.com/en-us/azure/active-directory/roles/permissions-reference).
+:::note Admin and org owner accounts
+If you're an admin setting up SSO, Semgrep creates an SSO account for you that is a **Member** by default. To elevate the permissions granted to this account, log in to Semgrep with your admin account using the original login method, then change the role of your newly created SSO account to **Admin**.
```
khorne3, line 69: The Linear card suggested a KB article, but I thought it would be more useful here; let me know if you think it should be moved

s-santillan: Looks good to me!

docs/deployment/sso.md
```diff
@@ -157,3 +81,8 @@ If you have SSO enabled, you can turn off login using GitHub or GitLab credentia
 :::warning
 Ensure that you have at least one user who can log in through SSO before disabling sign in with GitHub or GitLab.
 :::
+
+## See also
```
s-santillan, line 85: Thank you for moving these to their specific KB articles.

s-santillan (approved):

docs/deployment/sso.md
armchairlinguist: This isn't actually quite true. It's dependent on a different setting in the deployment, whose default value for more recent orgs is Member, so in that sense it's what most people see. But the result is controlled by the setting at https://semgrep.dev/orgs/armchairlinguist/settings/access/defaults

docs/kb/semgrep-appsec-platform/saml-google-workspace.md
```diff
@@ -5,7 +5,7 @@ tags:
   - SSO
 ---
 
-# SAML Single Sign-on with Google Workspace
+# SAML single sign-on with Google Workspace
```
armchairlinguist, line 8: Actually, after I looked at all the others, I realized they all say "SAML SSO". 

Possible options:
* Change this to `SAML SSO` too
* Change all the others to `SAML single sign-on`
* Distinguish between setup guides, which use the full name, and error troubleshooting, which has the abbreviation

docs/deployment/sso.md
```diff
@@ -21,10 +22,6 @@ This article walks you through single-sign on (SSO) configuration. Semgrep suppo
 
 ## OpenID Connect / OAuth 2.0
 
-:::warning
-Semgrep AppSec Platform does not support using OpenID with Microsoft Entra ID. Follow the instructions for [setting up SAML SSO with Microsoft Entra ID](#set-up-saml-sso-with-microsoft-entra-id) instead.
-:::
```
armchairlinguist, line 26: I think we should retain this callout, and just link to the KB. We used to have a lot of people trying to do this, and the callout seems to have helped.

docs/kb/semgrep-appsec-platform/saml-google-workspace.md
khorne3: Normally, I would change everything to SAML single-sign on, but I think SSO is common enough that it can stand on its own, so I opted for SAML SSO. It also lines up with the error messages, which I think is helpful

docs/deployment/sso.md
khorne3: Sounds good -- will bring this back

docs/deployment/sso.md
```diff
@@ -66,85 +67,14 @@ SAML2.0 is configured through **Semgrep AppSec Platform**. To set up SSO:
 5. Select the box next to **This SSO supports non-password authentication mechanisms (e.g. MFA, X509, PasswordLessPhoneSignin)** if applicable.
 6. Click **Save** to proceed.
 
-If you are using Google Workspace SAML, see [SAML Single Sign-on with Google Workspace](/docs/kb/semgrep-appsec-platform/saml-google-workspace) for more specific guidance.
```
armchairlinguist, line 69: I'd like it if we could retain a more obvious link for this page too. I thought this placement was most helpful, but above the general instructions would also work. Because there are other sections between the general SAML instructions and the references where the final link is, I think this gets a bit lost otherwise.

khorne3: I made several changes: I put this note back, but I put it at the top the way we did for Entra, both for consistency's sake and so people don't go through the steps, then realize they shouldn't. I also changed them to notes boxes so they aren't so in-your-face. Additionally, we added the admonition regarding admin accounts at the end here, and I thought two notes would be a bit much. WDYT?

armchairlinguist: Yeah, this looks good!

armchairlinguist (approved):

PR #1930, opened 2025-01-27
Fix incorrect references to manifest files

armchairlinguist:
This KB is about generating lockfiles, defined as files that list specific pinned dependency versions. These are the only files we currently support for Python with SSC.

### Please ensure


khorne3 (approved):

sebasrevuelta (approved):
Thanks for the update.

PR #1936, opened 2025-01-29
update broker docs to remove publicKey, allowedIps, heartbeat

khorne3:
[Preview](https://deploy-preview-1936--semgrep-docs-prod.netlify.app/docs/semgrep-ci/network-broker#create-the-config-file)

### Please ensure

### DIFF
--- a/docs/semgrep-ci/network-broker.md
+++ b/docs/semgrep-ci/network-broker.md
@@ -131,7 +131,7 @@ The broker requires a WireGuard keypair to establish a secure connection. To gen
 
 ### Update the config with your SCM information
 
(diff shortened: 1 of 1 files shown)


tpetr (approved):
Looks good but we might want to add a caveat that customers running a version of the broker before v0.25.0 will still need to specify these fields (and ideally, upgrade to the latest version so that they can simplify their config file). I'll abstain from posting the list since this is a public repo, but it looks like ~2/3 of our brokers are on v0.24.0 or older.

khorne3:
@tpetr I added back the old instructions as as a second tab, along with the caveat that people should upgrade

armchairlinguist:
The changes look good to me, good spot that we have to keep both due to versioning. While we're in here, the new line 134 should include ADO, since we have it as an option in the next tabset.

PR #1950, opened 2025-02-03
fix: clarify we support all versions of npm, yarn and pnpm for SSC

mmcqd:
Yarn is the only package manager we currently mention multiple versions for, and we only say 1-3, even though we also support yarn v4. This change makes it clear we support all versions of npm, yarn and pnpm. I think it would also be reasonable to just drop mention of versions from all three, to keep consistent with the rest of the able.

### Please ensure

### DIFF
--- a/docs/supported-languages.md
+++ b/docs/supported-languages.md
@@ -115,20 +115,20 @@ Additionally, Semgrep offers beta support for the scanning of Java projects **wi
   </tr>
   <tr>
    <td rowspan="3">JavaScript or TypeScript</td>
-   <td>npm (all versions)</td>
+   <td>npm</td>
    <td><code>package-lock.json</code></td>
    <td style={{"text-align": "center"}}>GA</td>
    <td>✅</td>
    <td rowspan="3" style={{"text-align": "center"}}>GA</td>
   </tr>
   <tr>
-   <td>Yarn (all versions)</td>
+   <td>Yarn</td>
    <td><code>yarn.lock</code></td>
    <td style={{"text-align": "center"}}>GA</td>
    <td>✅</td>
   </tr>
   <tr>
-   <td>pnpm (all versions)</td>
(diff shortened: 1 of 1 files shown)


s-santillan:
Would suggest the latter suggestion from @mmcqd:

>  I think it would also be reasonable to just drop mention of versions from all three, to keep consistent with the rest of the able.

I am OOO so this is ultimately @khorne3 's decision :)

docs/supported-languages.md
```diff
@@ -115,20 +115,20 @@ Additionally, Semgrep offers beta support for the scanning of Java projects **wi
...
-   <td>npm (Node.js)</td>
+   <td>npm (all versions)</td>
    <td><code>package-lock.json</code></td>
    <td style={{"text-align": "center"}}>GA</td>
    <td>✅</td>
    <td rowspan="3" style={{"text-align": "center"}}>GA</td>
   </tr>
   <tr>
-   <td>Yarn, Yarn 2, Yarn 3</td>
+   <td>Yarn (all versions)</td>
```
khorne3, line 125:
```suggestion
   <td>Yarn</td>
```

docs/supported-languages.md
```diff
@@ -115,20 +115,20 @@ Additionally, Semgrep offers beta support for the scanning of Java projects **wi
...
    <td><code>yarn.lock</code></td>
    <td style={{"text-align": "center"}}>GA</td>
    <td>✅</td>
   </tr>
   <tr>
-   <td>pnpm</td>
+   <td>pnpm (all versions)</td>
```
khorne3, line 131:
```suggestion
   <td>pnpm</td>
```

khorne3 (approved):

khorne3:
GitHub behavior is weird this morning, so I'll fix this PR (without version info) and merge a bit later today.

PR #1960, opened 2025-02-11
Clarify how rule timeout works

armchairlinguist:
It wasn't super clear from the wording that the rule timeout is per file and that it doesn't set a timeout on the whole job at any point, so I tried to clarify that. Open to further adjustments!

### Please ensure

- [x] A subject matter expert (SME) reviews the content (considering myself an SME here, also it's a small PR :) )

### DIFF
--- a/docs/deployment/customize-ci-jobs.md
+++ b/docs/deployment/customize-ci-jobs.md
@@ -55,12 +55,12 @@ Some Semgrep CI jobs require manual configuration of [diff-aware scans](/semgrep
 
 ## Set a custom timeout
 
-By default, Semgrep spends **5 seconds** to run **per rule**. To **set a custom timeout** for the Semgrep job, set the `SEMGREP_TIMEOUT` environment variable in seconds. Decreasing this value speeds up your scans, but with the possibility of skipping some rules. Alternatively, increasing this value ensures that your most complex rules finish running. For example:
+By default, Semgrep spends **5 seconds** to run **per rule** on each relevant file. To **set a different timeout**, set the `SEMGREP_TIMEOUT` environment variable (the value is in seconds). Decreasing this value speeds up your scans, but with the possibility of skipping some rules. Alternatively, increasing this value ensures that your most complex rules finish running. For example:
 
 ```sh
 SEMGREP_TIMEOUT="3" # Sets the per-rule timeout to 3 seconds.
 ```
 
 :::caution
-Setting this variable to **0** removes the time limit, meaning that rules can take any amount of time to run.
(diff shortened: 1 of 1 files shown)


docs/deployment/customize-ci-jobs.md
```diff
@@ -55,12 +55,12 @@ Some Semgrep CI jobs require manual configuration of [diff-aware scans](/semgrep
 
 ## Set a custom timeout
 
-By default, Semgrep spends **5 seconds** to run **per rule**. To **set a custom timeout** for the Semgrep job, set the `SEMGREP_TIMEOUT` environment variable in seconds. Decreasing this value speeds up your scans, but with the possibility of skipping some rules. Alternatively, increasing this value ensures that your most complex rules finish running. For example:
+By default, Semgrep spends **5 seconds** to run **per rule** on each relevant file. To **set a different timeout**, set the `SEMGREP_TIMEOUT` environment variable (the value is in seconds). Decreasing this value speeds up your scans, but with the possibility of skipping some rules. Alternatively, increasing this value ensures that your most complex rules finish running. For example:
```
khorne3, line 58:
```suggestion
By default, Semgrep spends **5 seconds** to run **per rule** per file scanned. To **set a different timeout**, set the `SEMGREP_TIMEOUT` environment variable (the value is in seconds). Decreasing this value speeds up your scans, but with the possibility of skipping some rules. Alternatively, increasing this value ensures that your most complex rules finish running. For example:
```

khorne3 (approved):

docs/deployment/customize-ci-jobs.md
armchairlinguist: @khorne3 I'm definitely open to revising 'relevant' to something more clear, but this is not clear, unfortunately. It doesn't clarify that every rule does not scan every file but only files which are within the language targets. I'm pretty sure we do talk about this elsewhere but it's not easy to search for so I don't know exactly where. If there's a good / approved way to talk about it I can use that here too.

khorne3: Got it -- maybe "relevant file scanned"? It seems like both pieces are crucial to understanding

armchairlinguist: Yeah, I think I actually didn't understand the purpose of your change either, sorry. If relevant is good enough as a word, we can just do both, but let me see if I can come up with something both short and clear.

armchairlinguist: I updated, mind taking another look?

khorne3:
@armchairlinguist looks good -- it's not the simplest sentence, but I think we need all of it to convey what we're trying to say

khorne3 (approved):

PR #1969, opened 2025-02-12
Various clarifications and updates for the Jira integration

armchairlinguist:
A few updates and clarifications on operation for the Jira integration.

I'd also ideally like to clarify the wording on "multiple Jira projects" (it's a one-at-a-time but multiple options thing), but it's not a very simple change and these other changes are pretty easy to make. Totally open to adding that though if we come up with something.

### Please ensure


khorne3 (approved):
LGTM!

Re: multiple tickets -- maybe something like, "You can create tickets for a finding in one or more Jira projects. To do so, manually select the projects when creating the initial ticket in Semgrep. Semgrep creates a ticket for the finding in each of the selected projects." I don't know how to shorten it without (re)introducing ambiguity.

PR #1970, opened 2025-02-12
Update connect scm instructions and improve readability

s-santillan:

### Please ensure

This is an expansion of the original request.

- The doc seems long because we have instructions per-provider. We can use tabs for this.
- The UI changed so the steps changed.
- Clarified a bit about choosing accounts to install SCMs on because some folks may wonder "where did the other accounts come from?"
- Added screenshots at my peril.

### DIFF
--- a/docs/deployment/connect-scm.md
+++ b/docs/deployment/connect-scm.md
@@ -9,6 +9,9 @@ tags:
   - Semgrep AppSec Platform
 ---
 
+import Tabs from '@theme/Tabs';
+import TabItem from '@theme/TabItem';
+
 # Connect a source code manager
 
 :::note Your deployment journey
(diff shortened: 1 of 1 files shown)


docs/deployment/connect-scm.md
```diff
@@ -33,38 +36,54 @@ The process to connect a source code manager depends on whether your SCM tool is
...
 ### Azure DevOps Cloud
 
-1. Sign in to Semgrep AppSec Platform.
-1. On the sidebar, click the account name to open the drop-down menu.
-1. Using the drop-down menu, select the account you want to make a connection for.
-1. Go to **Settings** > **Source Code Managers**, and click **Add Azure DevOps**.
-![Source code manager tab](/img/source-code-manager.png#md-width)
+1. Sign in to [<i class="fas fa-external-link fa-xs"></i> Semgrep AppSec Platform](https://semgrep.dev/login).
+1. Optional: If you have created more than one Semgrep account, select the account you want to make a connection for by clicking on the **Navigation bar > Your account name > The account you want to connect**.
+   <img src="/docs/img/more-accounts-dropdown.png" height="350px" />
```
khorne3, line 55: Not a big deal, but I think the screenshot should be on its own line (this is present for all tabs)

docs/deployment/connect-scm.md
```diff
@@ -33,38 +36,54 @@ The process to connect a source code manager depends on whether your SCM tool is
...
+    {label: 'Azure DevOps Cloud', value: 'azure-devops-cloud'},
+    {label: 'Bitbucket Cloud', value: 'bitbucket-cloud'},
+    {label: 'GitHub Cloud', value: 'github-cloud'},
+    {label: 'GitLab Cloud', value: 'gitlab-cloud'},
+    ]}
+>
+
+<TabItem value='azure-devops-cloud'>
+
 ### Azure DevOps Cloud
```
khorne3, line 51: Since we have tabs now, I think these headers can be removed (this goes for all the H3s that are SCM names), unless you're keeping 'em to generate the right-hand nav bar?

khorne3 (approved):

docs/deployment/connect-scm.md
s-santillan: Oh dear. It's probably formatted slightly differently for our desktops, I'll add a br.

docs/deployment/connect-scm.md
s-santillan: Yeah, I'll remove them. The h3 headers are nice but they seem "unclickable" when they're intabs (except for the tab currently displayed). Thanks for the catch.

s-santillan: @khorne3 WDYT - here's a screenshot of both w/o headers and with:

image

s-santillan: I am very slightly in favor of removing the headers because the tabs are clear enough to me, but posting as a sanity check.

khorne3: I'm in favor of removing the headers; it looks cleaner and more deliberate to me

PR #1974, opened 2025-02-13
update lockfiles

s-santillan:

### Please ensure

- [ ] It builds, all components green

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,9 +16,9 @@
   },
   "dependencies": {
     "@algolia/client-search": "^4.23.3",
-    "@docusaurus/core": "^3.5.2",
-    "@docusaurus/plugin-client-redirects": "^3.5.2",
-    "@docusaurus/preset-classic": "^3.5.2",
+    "@docusaurus/core": "^3.7.0",
+    "@docusaurus/plugin-client-redirects": "^3.7.0",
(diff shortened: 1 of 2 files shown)


khorne3 (approved):
I don't know what the neutral tests mean, but it builds and looks fine to me

s-santillan:
Good call out on specifying what the warnings are. I'll list their broad categories in the order in which they appear:

1. Previously, the build log complained about the .svg files (which are rendered correctly). The warning still appears in 3.7.0. 
2. This one is new: 
    ```
    Resolving 'terser-webpack-plugin/dist/minify' in /opt/build/repo/node_modules/terser-webpack-plugin/dist for build dependencies doesn't lead to expected result '/opt/build/repo/node_modules/terser-webpack-plugin/dist/minify.js', but to '/opt/build/repo/node_modules/terser-webpack-plugin/node_modules/terser-webpack-plugin/dist/minify.js' instead. Resolving dependencies are ignored for this path.
    ```
    It also doesn't seem to affect the actual serving of pages. But it is weird that it suddenly popped out in Docusaurus 3.7.0.
3. We have a new slew of broken links related to my changes in Supported languages + a few others that have come up over time - I will fix these in a separate PR.

PR #1980, opened 2025-02-20
Update contributing instructions for OCaml 5

nmote:

In https://github.com/semgrep/semgrep-proprietary/pull/2932, I am updating nearly everything to OCaml 5. For now, we can still build on OCaml 4, but on macOS we need to use OCaml 5 because there are some specific things that we had to change regarding static linking. At this point, the path of least resistance is for contributors to use OCaml 5.

I also slightly simplified the instructions here, and modified them to create an opam switch named `semgrep` (it's a good idea to keep a separate switch for each project).

### Please ensure


s-santillan (approved):

s-santillan:
@nmote Thank you! Merging.

PR #1989, opened 2025-02-25
Sara/tec 294 docs add GitHub app installation and connection workflow

s-santillan:

### Please ensure

- Fixes TEC-294. Additionally adds a callout that says SSO must have been set up first (because "usually" SSO is set up after) before performing the connection.
- Updates "integration" with "connection" because "connect" is used in the Semgrep AppSec Platform UI.
- **I am not sure about using double nested tabs**, feedback welcome.

### Static preview

Snag_106b5b9

### DIFF
--- a/docs/deployment/connect-scm.md
+++ b/docs/deployment/connect-scm.md
@@ -106,7 +106,7 @@ These instructions are for orgs that set up a non-GitHub SSO **before** performi
 
 
 1. Navigate to the following link: [<i class="fas fa-external-link fa-xs"></i> Semgrep GitHub app](https://github.com/marketplace/semgrep-dev) and install the Semgrep GitHub app onto the GitHub org you want to connect to.
-1. Sign in to [<i class="fas fa-external-link fa-xs"></i> Semgrep AppSec Platform](https://semgrep.dev/login) through your SSO.
+1. Sign in to [<i class="fas fa-external-link fa-xs"></i> Semgrep AppSec Platform](https://semgrep.dev/login) using SSO.
 1. Optional: If you have created more than one Semgrep account, select the account you want to make a connection for by clicking on the **Navigation bar > Your account name > The account you want to connect**.<br />
    <img src="/docs/img/more-accounts-dropdown.png" height="350px" />
 1. From the **Navigation bar**, click **<i class="fa-solid fa-gear"></i> Settings > Source code managers**.
--- a/docs/deployment/connect-scm.md
+++ b/docs/deployment/connect-scm.md
@@ -69,7 +69,7 @@ If you opted to scan a GitHub or GitLab repository when you initially signed in,
    - The **Name of your Bitbucket Workspace**
    - Your **Access token**. Semgrep requires a [workspace-level access token](https://support.atlassian.com/bitbucket-cloud/docs/create-a-workspace-access-token/).
 1. Click **Connect** to save and proceed.
-1. The Bitbucket project is now listed under **Source Code managers**. Click **Test** to verify that the new connection is installed correctly.
+1. The Bitbucket project is now listed under **Source code managers**. Click **Test** to verify that the new connection is installed correctly.
 
 </TabItem>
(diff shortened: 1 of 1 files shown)


khorne3 (approved):

jkinsfather:
Overall, an accurate and helpful addition.

One of the things I worry about is confusion due to the limitations of tabs and hyperlinks. 

Often, users do not navigate to this page organically. They are sent links by our support and sales teams. Since we can't send someone a hyperlink of a tab, the process for the non-GitHub SSO user would be the following:
1. They are sent and open a hyperlink to the "Connect to cloud-hosted orgs" header. 
2. They click on the GitHub cloud tab. 
3. They are automatically shown the "Default GitHub Cloud" tab. Now they need to either:
    3a. Read and correctly follow written instructions given by our team to switch to the `...with non-GitHub SSO` tab. 
    3b. Notice and infer that ' ... with non-GitHub SSO' refers to the method by which they logged in to Semgrep and NOT the log-in method to say, their GitHub Cloud Org. 
    
To avoid confusion, would it make sense to do one or both of the following?:
1. Add some kind of anchor under the tabs/sub-tabs so we can link directly to the tabs without needing to deliver additional instructions to the customer for navigation? 
2. Add an info box under the `Default GitHub Cloud` sub-tab explaining that `This is for Semgrep orgs set up with GitHub SSO for authentication. If your org is set up with non-GitHub SSO, follow the instructions in the Non-GitHub SSO tab.`

s-santillan:
To avoid confusion, would it make sense to do one or both of the following?:

> Add some kind of anchor under the tabs/sub-tabs so we can link directly to the tabs without needing to deliver additional instructions to the customer for navigation?

Let's say the user needs the instructions for Bitbucket and we add a header `### Bitbucket Cloud` nested in the tab - there's unfortunately no capability for the tab to "select itself" or to "show itself" - you'd still have to click on the tab, then the TOC link on the right hand side will work. This still isn't ideal as it requires the user to press the tab. 

I see your point about the need to link directly, or with the least amount of steps, especially considering how crucial deployment is. In this case it may be best **to break out each SCM into its own doc**. (cc @khorne3 for feedback)

For now, I can:

1. Remove the subnesting (as it will require the user to Click **Connect to cloud-hosted orgs > Click GitHub Cloud > Click GitHub with non-GitHub SSO** which is a frankly painful sentence to type out.
2. Follow your second recommendation: _Add an info box under the Default GitHub Cloud sub-tab explaining that This is for Semgrep orgs set up with GitHub SSO for authentication. If your org is set up with non-GitHub SSO, follow the instructions in the Non-GitHub SSO tab._

s-santillan:
@jkinsfather @khorne3 I updated the PR, addressing the comments above: https://deploy-preview-1989--semgrep-docs-prod.netlify.app/docs/deployment/connect-scm#github-cloud-with-non-github-sso

khorne3 (approved):
@s-santillan I agree with the need for individual SCM pages, both here and elsewhere. Perhaps our next big non-feature related project?

docs/deployment/connect-scm.md
```diff
@@ -67,34 +67,46 @@ If you opted to scan a GitHub or GitLab repository when you initially signed in,
...
 
-</TabItem>
+### GitHub Cloud with non-GitHub SSO
 
-<!-- removed temporarily because we're using the "old flow"
-:::tip
-- Getting Assistant recommendations grants Semgrep **code access**.
-- **Leave PR comments** refers to Semgrep's capability to post findings to developers in PRs.
-:::
+These instructions are for orgs that set up a non-GitHub SSO **before** performing any of the steps in this document.
```
jkinsfather, line 95: Use of orgs is ambiguous. Perhaps re-phrase to mirror the GitHub SSO instructions? "These steps are for users that sign in to Semgrep through a non-GitHub SSO provider"

s-santillan: done!

jkinsfather (approved):

PR #1996, opened 2025-02-27
update docs to mention/display critical severity status

khorne3:
### Please ensure


docs/kb/rules/understand-severities.md
```diff
@@ -9,7 +9,7 @@ tags:
 
 ## Semgrep Code and Secrets
 
-Semgrep Code and Secrets rules have one of three severity levels: `ERROR` (High), `WARNING` (Medium), or `INFO` (Low). The severity indicates how critical the issues that a rule potentially detects are.
+Semgrep Code and Secrets rules have one of four severity levels: Critical, High, Medium, and Low. The severity indicates how critical the issues that a rule potentially detects are.
```
s-santillan, line 12: (Sorry for getting to this so late!) - curious, do we not display `ERROR` and `WARNING` in the SARIF or JSON or API?

s-santillan (approved):

docs/kb/rules/understand-severities.md
khorne3: Not the API, but in the JSON and SARIF outputs, yes. But it looks like those are kept as-is to be backward compatible. Do you think I should update and mention this?

s-santillan: Maybe I should mention it in the JSON/SARIF doc instead. Let's ship this as is, it looks cleaner. Feel free to resolve the comment!

PR #2008, opened 2025-03-07
Add permissions for assistant features

s-santillan:

### Please ensure

- [x] Katie knows best


s-santillan:
@khorne3 I value your opinion on whether or not we should separate the table, as I have done here with the PR, or put the features under the "Operations permitted per role" table. They are to my knowledge accurate based on the docs you have written.

khorne3:
@s-santillan I think this is good for now. At a later date, we might grant broader access to these features and at that point, we might want to consolidate the tables. But, we're not there yet.

khorne3 (approved):

PR #2024, opened 2025-03-14
Replace 'scanning root' with 'scan root' because it sounds better

mjambon:
This a global search-and-replace for "scanning root" that becomes "scan root". I'm the one who originally started using the term "scanning root" but I realized "scan root" makes more sense after noticing a user using the term.

### Please ensure

### DIFF
--- a/src/components/reference/_cli-help-scan-output.md
+++ b/src/components/reference/_cli-help-scan-output.md
@@ -336,10 +336,10 @@ OPTIONS
            option forces the project root to be a specific folder and assumes
            a local project without version control (novcs). This option is
            useful to ensure the '.semgrepignore' file that may exist at the
-           project root is consulted when the scanning root is not the
+           project root is consulted when the scan root is not the
            current folder '.'. A valid project root must be a folder (path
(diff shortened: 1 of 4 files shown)


src/components/reference/_cli-help-scan-output.md
```diff
@@ -336,10 +336,10 @@ OPTIONS
            option forces the project root to be a specific folder and assumes
            a local project without version control (novcs). This option is
            useful to ensure the '.semgrepignore' file that may exist at the
-           project root is consulted when the scanning root is not the
+           project root is consulted when the scan root is not the
```
s-santillan, line 339: @mjambon FYI this is autogenerated from the :latest docker image's CLI --help output, so this will be overwritten unless the help text is also updated. Just an FYI.

s-santillan (approved):
I agree with this change!

src/components/reference/_cli-help-scan-output.md
mjambon: Good to know. I'll get to it when week or so.

PR #2025, opened 2025-03-14
Clarify "Excluding a rule in paths"

mjambon:
A user was confused by the original phrasing.

Also, I changed the `**/*.jinja2` example into one that makes more sense (since `**/*.jinja2` is equivalent to `*.jinja2`).

### Please ensure


nmote (approved):
Makes sense to me. I guess I can be a SME here?

s-santillan:
SGTM!

PR #2038, opened 2025-03-19
clarify why setting up GitLab SMS requires a higher role 

khorne3:
The token needs a higher role to set up webhooks, but this can be downgraded

### Please ensure

### DIFF
--- a/docs/deployment/managed-scanning/gitlab.md
+++ b/docs/deployment/managed-scanning/gitlab.md
@@ -10,6 +10,7 @@ tags:
   - Semgrep AppSec Platform
 ---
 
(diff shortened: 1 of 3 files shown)


s-santillan:
Do we want to link to this doc from the **SMS Quickstart** and maybe the **Add a Source Code Manager** docs?

khorne3:
@s-santillan I turned this info into a component so it can be used on the GL SMS page and quickstart; what do you think? I didn't link to Connect an SCM, since I have a backlog task to rethink how that page is laid out now that we have SMS, but I think the component would be helpful for that reorg, too

s-santillan (approved):

PR #2039, opened 2025-03-19
add link to doc on scanning without lockfiles from doc on building Maven trees

khorne3:
This PR adds a link from the Maven page to the page on scanning without lockfiles. Once the latter comes out of beta, we can revisit this page and update it to indicate which is the preferred method.

### Please ensure


s-santillan (approved):
LGTM!

PR #2056, opened 2025-03-26
docs: initial `metavariable-name.module{,s}`

kopecs:

### Please ensure

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -763,7 +763,7 @@ This removes quotes (`'`, `"`, and `` ` ``) from both ends of the metavariable c
 ### `metavariable-name`
 
 :::tip
-- `metavariable-name` requires a Semgrep account.
+- `metavariable-name` requires a Semgrep account and the use of Semgrep's proprietary engine. This means that it does **not** work with the `--oss-only` flag.
 - While optional, you can improve the performance of `metavariable-name` by enabling **[cross-file analysis](/docs/getting-started/cli#enable-cross-file-analysis)**. 
 :::
 
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -772,7 +772,7 @@ The `metavariable-name` operator requires the Pro engine, since it requires addi
 rules:
   - id: insecure-method
     patterns:
-      - pattern: $MODULE.method(...)
+      - pattern: $MODULE.insecure(...)
       - metavariable-name:
           metavariable: $MODULE
           module: "@foo-bar"
@@ -785,16 +785,27 @@ rules:
 The pattern immediately above matches the following:
 
 ```javascript
+// ECMAScript modules
 import * as lib from '@foo-bar';
 import * as lib2 from 'myotherlib';
 
+// CommonJS modules
+const { insecure } = require('@foo-bar');
+const lib3 = require('myotherlib');
+
 // ruleid: insecure-method
 // highlight-next-line
-lib.method("test")
+lib.insecure("test");
+// ruleid: insecure-method
+// highlight-next-line
+insecure("test");
+
 // ok: insecure-method
-lib.method2("test")
+lib.secure("test");
 // ok: insecure-method
-lib2.method("test")
+lib2.insecure("test");
+// ok: insecure-method
+lib3.insecure("test");
 ```
 
 In the event that a match should occur if the metavariable matches one of a variety of matches, there is also a shorthand `modules` key, which takes a list of module names.
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -762,8 +762,9 @@ This removes quotes (`'`, `"`, and `` ` ``) from both ends of the metavariable c
 
 ### `metavariable-name`
 
-:::info
-`metavariable-name` requires a Semgrep account and **[cross-file analysis](/docs/getting-started/cli#enable-cross-file-analysis)**. 
+:::tip
+- `metavariable-name` requires a Semgrep account.
+- While optional, you can improve the performance of `metavariable-name` by enabling **[cross-file analysis](/docs/getting-started/cli#enable-cross-file-analysis)**. 
 :::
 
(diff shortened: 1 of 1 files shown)


docs/writing-rules/rule-syntax.md
```diff
@@ -759,6 +760,63 @@ int("2147483646")
...
+    message: Uses insecure method from @foo-bar.
+    languages:
+      - javascript
+    severity: ERROR
+```
+
+The pattern immediately above matches the following:
+
+```javascript
+import * as lib from '@foo-bar';
```
inkz, line 788: I would show both import and require in the example, just to show that it works for both
```javascript
// ECMAScript modules
import * as lib from '@foo-bar'
import * as lib2 from 'myotherlib'

// CommonJS modules
const { method } = require('@foo-bar')
const lib3 = require('myotherlib')

// ruleid: insecure-method
lib.method("test")

// ruleid: insecure-method
method("test")

// ok: insecure-method
lib.method2("test")

// ok: insecure-method
lib2.method("test")

// ok: insecure-method
lib3.method("test")
```

inkz (approved):
Added a comment, but in general lgtm :+1:

docs/writing-rules/rule-syntax.md
```diff
@@ -759,6 +760,74 @@ int("2147483646")
 
 This removes quotes (`'`, `"`, and `` ` ``) from both ends of the metavariable content. As a result, Semgrep detects `"2147483648"`, but it does **not** detect `"2147483646"`. This is useful when you expect strings to contain integer or float data.
 
+### `metavariable-name`
+
+The `metavariable-name` operator adds a constraint to the types of identifiers a metavariable is able to match. Currently the only constraint supported is on module or namespace an identifier originates from. This is useful for filtering results in languages which don't have a native syntax for fully qualified names, or languages where module names may contain characters which are not legal in identifiers (e.g., JavaScript, TypeScript).
```
s-santillan, line 765:
```suggestion
The `metavariable-name` operator adds a constraint to the types of identifiers a metavariable is able to match. Currently the only constraint supported is on the **module** or **namespace** an identifier originates from. This is useful for filtering results in languages which don't have a native syntax for fully qualified names, or languages where module names may contain characters which are not legal in identifiers, such as JavaScript or TypeScript). 
```

docs/writing-rules/rule-syntax.md
```diff
@@ -759,6 +760,74 @@ int("2147483646")
 
 This removes quotes (`'`, `"`, and `` ` ``) from both ends of the metavariable content. As a result, Semgrep detects `"2147483648"`, but it does **not** detect `"2147483646"`. This is useful when you expect strings to contain integer or float data.
 
+### `metavariable-name`
+
+The `metavariable-name` operator adds a constraint to the types of identifiers a metavariable is able to match. Currently the only constraint supported is on module or namespace an identifier originates from. This is useful for filtering results in languages which don't have a native syntax for fully qualified names, or languages where module names may contain characters which are not legal in identifiers (e.g., JavaScript, TypeScript). 
+
+:::info
+The `metavariable-name` operator requires the Pro engine, since it requires additional name resolution information. Moreover, name resolution information is higher fidelity with interfile enabled, and thus `metavariable-name` will be more accurate with interfile naming.
```
s-santillan, line 768:
```suggestion
The `metavariable-name` operator requires cross-file analysis, which you can enable following the steps in [Enable cross-file analysis](/docs/getting-started/cli#enable-cross-file-analysis). This is because it requires additional name resolution information. Moreover, name resolution information is more precise with cross-file enabled, and thus `metavariable-name` will be more accurate with cross-file naming.
```

s-santillan (approved):
Hi Cooper! Thanks for this addition. It reads well! I have just added some notes to keep the tone consistent with the other body of materials that we produce across the docs, marketing website, and collateral. FYI We use the [Google Style Guide](https://developers.google.com/style) in general. There's no need for you to read it all to write the docs, (that is what we editors are for) but I am mentioning it here so you know where some of our notes come from.

docs/writing-rules/rule-syntax.md
kopecs: It does require Pro but not interfile (but works better with interfile). Do you have a preferred wording for the first sentence, given that?

s-santillan: @kopecs in the **Semgrep AppSec Platform > Settings** tab, there is no setting in which we enable "only" Pro - the UI toggle enables both. Also, the toggle explicitly calls it **Cross-file** 😅 so I can't change the toggle name. 

Maybe we can say:

> The `metavariable-name` operator requires _advanced_ analysis, which you can enable following the steps in [Enable cross-file analysis](/docs/getting-started/cli#enable-cross-file-analysis) ...

kopecs: OK. Let me confirm what the default here is if that toggle is off. Not sure if we still have Pro intrafile or if it is OSS (or it depends on some other thing). We should disclaim it here either way, since CE users won't have access to it, but let me get back to you.

s-santillan: Anything "Pro" branded is by default not OSS, so yeah, CE users will not have access to it. Are you ok with me pushing a commit to add a callout?

kopecs: Feel free to push whatever!

docs/writing-rules/rule-syntax.md
```diff
@@ -759,6 +760,78 @@ int("2147483646")
 
 This removes quotes (`'`, `"`, and `` ` ``) from both ends of the metavariable content. As a result, Semgrep detects `"2147483648"`, but it does **not** detect `"2147483646"`. This is useful when you expect strings to contain integer or float data.
 
+### `metavariable-name`
+
+:::info
+`metavariable-name` requires a Semgrep account and **[cross-file analysis](/docs/getting-started/cli#enable-cross-file-analysis)**.
```
kopecs, line 766: Seems redundant with the updated version at lines 771-3? Cross-file isn't required though, so if we want to keep this it should either just say Semgrep account or Pro engine.

kopecs: You do not _need_ cross-file. But, the results are better with cross-file enabled.

The pro engine is a requirement. This is not quite the same as being logged-in, since you can pass, e.g., `--oss-only`.

docs/writing-rules/rule-syntax.md
s-santillan: @kopecs This is the only remaining comment to resolve - are you ok with committing the suggestion?

kopecs: Bold/italics on module or namespace seem a bit gratuitous to me, but adding "the" and rewording to "such as" instead of "e.g." sounds good.

PR #2076, opened 2025-04-04
Don't overwrite the source files by their expansion when running the webserver

mjambon:
Each time I run the local web server to preview the documentation, I end up with 3 source files that are modified in place but shouldn't be committed. This PR fixes this problem by using different files for the source before expansion (`.md.template`) and for the output after expansion (`.md`).

(It is unfortunate that the ".md" format variant we use here doesn't support ordinary HTML/MarkDown comments at the top of the file. This prevents us from placing a "DO NOT EDIT" line at the beginning of the output files)

Test plan: launch the documentation server locally with `make` then check the appearance of http://localhost:3000/docs/cli-reference, http://localhost:3000/docs/ignoring-files-folders-code, and http://localhost:3000/docs/extensions/overview. Then check that `git status` doesn't report changes.

### Please ensure

- ~~[ ] A subject matter expert (SME) reviews the content~~ n/a
- ~~[ ] A technical writer reviews the content or PR~~ n/a


s-santillan (approved):
Verified on my end. Thank you Martin.

PR #2102, opened 2025-04-15
add permissions callout for ado sms

vivekkhimani:
Some users are hesitant about providing full access to the ADO access tokens and also the service accounts that generate these tokens. Add a callout to explain that the permissions can later be degraded.

### Please ensure


khorne3 (approved):
Thank you!

PR #2113, opened 2025-04-21
Jenkins diff scans kb syntax fix

stuartcmehrens:
Removing whitespace before and after equals sign so snippet can be copy pasted and behave as intended.

### Please ensure


armchairlinguist:
@stuartcmehrens When I was working on this, I referenced https://www.jenkins.io/doc/book/pipeline/multibranch/#additional-environment-variables which classes this as an env var: `additional information about the branch being built through the env global variable`. Do you have new info that contradicts that?

stuartcmehrens:
@armchairlinguist my understanding is that the env prefix may or may not be used. see screenshot below from the jenkins instance that i maintain.
image

in any event, if we dont want to make that change, i can revert that bit back as this PR is really more about removing the whitespace before and after the equals sign.

armchairlinguist:
Oh I missed that change entirely - a mention in the PR description would definitely help :) 
If you don't mind, let's leave the env part alone, but happy to approve the spacing change, not sure why I did that 😅

armchairlinguist (approved):

PR #2137, opened 2025-05-08
Add overview paragraph in notifications page...

s-santillan:
... and supplementary information on notifications from Semgrep Code and Semgrep Supply Chain (i.e. backlinks)

# Preview

### Please ensure

### DIFF
--- a/docs/semgrep-appsec-platform/notifications.md
+++ b/docs/semgrep-appsec-platform/notifications.md
@@ -4,7 +4,7 @@ append_help_link: true
 title: Alerts and notifications
 hide_title: true
 toc_max_heading_level: 4
-description: "Learn how to receive Slack or email alerts about findings and failures and how to integrate using webhooks."
+description: "Learn how to receive Slack, webhook, or email notifications about findings."
 tags:
     - Deployment
     - Semgrep AppSec Platform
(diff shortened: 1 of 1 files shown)


s-santillan:
Note: If the notifications page still looks light, it may be because the Rule modes component is in a `summary` (collapsible) box.

docs/semgrep-appsec-platform/notifications.md
```diff
@@ -4,25 +4,71 @@ append_help_link: true
...
+</thead>
+<tbody>
+<tr>
+<td>Slack</td>
+<td rowspan="3">Customize through rule modes in [Policies page](/semgrep-code/policies)</td>
+<td>Limited customizability; configured by default to send notifications on <strong>reachable</strong> findings</td>
+</tr>
+<tr>
+<td>Email</td>
+<td>--</td>
```
khorne3, line 45: Do you think "not available" might be clearer here instead of the double dash?

khorne3 (approved):

docs/semgrep-appsec-platform/notifications.md
s-santillan: That is the TW answer and I will make this change, but in other parts of the docs I acceded to Sales to use `--` because it felt "softer" (in case you're wondering why some parts have `--` instead of "No" or "Not available".

docs/semgrep-appsec-platform/notifications.md
```diff
@@ -4,25 +4,71 @@ append_help_link: true
...
     - Deployment
     - Semgrep AppSec Platform
 ---
 
+import RuleModes from "/src/components/reference/_rule-modes.md"
+
 # Alerts and notifications
 
-Users can receive notifications from Semgrep AppSec Platform in the following channels:
+You can receive notifications for Semgrep Code and Supply Chain findings in the following channels:
```
armchairlinguist, line 17: Alas, this isn't actually true. Only Slack works for SSC, the others only do code.

s-santillan: Good callout. Updated the opening, removed the `ul` because the table served it better. It may have made the steps redundant, though  I also think keeping the general steps makes sure that people don't miss them (if they skim the table). WDYT?

docs/semgrep-appsec-platform/notifications.md
```diff
@@ -4,25 +4,66 @@ append_help_link: true
...
+<th>Channel</th>
+<th>Semgrep Code</th>
+<th>Semgrep Supply Chain</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td>[Slack](/semgrep-appsec-platform/slack-notifications)</td>
+<td rowspan="3">Integrate with Semgrep through **Settings > Integrations**. Customize through rule modes in [Policies page](/semgrep-code/policies).</td>
+<td>Integrate with Semgrep through **Settings > Integrations**. Limited customizability; configured by default to send notifications on <strong>reachable</strong> findings</td>
```
armchairlinguist, line 31: Is this currently accurate? Might be worth double-checking - I know policies only cover block/comment, but I recall there being other conditions on notifications before.

docs/semgrep-appsec-platform/notifications.md
armchairlinguist: seems good, I left another comment about verifying specific behavior though 😅

docs/semgrep-appsec-platform/notifications.md
s-santillan: For SSC, it is reachable findings (including **Always reachable**, i.e. maldeps). I tested this recently with Derian. There is no visible "you're getting reachable findings" anywhere AFAIK, not in Policies > Supply Chain (see screenshot); the only place where we indicate that you can get Supply Chain findings is in the notification in the workspace Slack you've installed Semgrep (see screenshot) under **Select target scan type**

Snag_14614af

Snag_1462f0d

s-santillan: cc @armchairlinguist ?

armchairlinguist: Oh right, sorry - I did the thing again where I was thinking it was only critical and high, but it's not - it's always been just reachable, only most of our reachables for a long time were critical/high.

s-santillan: All good. I'll go ahead with the merge then! 👍

PR #2140, opened 2025-05-12
Update to ignoring documentation, troubleshooting, etc.

s-santillan:

### Please ensure

## About this PR

The information and discussion about how Semgrep handles ignorelists is spread across our comms channels and some very old tickets. I decided to place the information here in one place so that we have a unified location and place where we can refine how the various ignore features best add value to the user experience.

The central discussion involves Platform ignores vs `.semgrepignore`. The `nosem` annotation is largely straightfoward. 

## Summary table

> Don't forget! SAP local project ignorelist EXISTS as does implicit `.semgrepignore` BY DEFAULT when a project is scanned by `semgrep ci`.

Goal | User-defined .semgrepignore | Platform ignores | End result | Why? |
|--|--|--|--|--|
| Ignore `somefile` | Added to user-defined semgrepignore | No changes | Success | Defined in semgrepignore
| Ignore `/tests/` | No existing user-defined semgrepignore | No changes | Success | Part of default (implicit) semgrepignore + also defined in SAP project ignores
| Ignore `/tests/` | Added to user-defined semgrepignore | No changes | Success | Part of user-defined semgrepignore + also defined in SAP project ignores
| Ignore `/tests/` | With existing user-defined semgrepignore that does NOT contain /tests | No changes (`/tests` is part of default SAP list) | Success | Ignore is successful because of SAP project ignores.
| Include `/tests/` | No existing user-defined semgrepignore | REMOVED from SAP project ignores | FAIL - skipped - ignored | Unexpected behavior - if something was removed from the SAP project ignore, it should be ignored, but the lack of a semgrepignore is overriding it (semgrepignore logic takes precedence) |
| Include `/tests/` | With user-defined semgrepignore without `/tests` on ignore list) | No changes | FAIL - skipped - ignored | On an unchanged SAP local project ignorelist, `/tests` is present. |
| Include `/tests/` | With user-defined semgrepignore without `/tests` on ignore list) | REMOVED `/tests` | Success  |Both SAP project ignore list and user-defined semgrepignore without tests is needed. |

### Notes

- I believe that the SAP definitions are added as part of `--exclude` patterns based on what I'm seeing in the log.
- SAP project definitions exist by default as do implicit (no semgrepignore file) definitions. Meaning to say that the "instruction" to ignore `/tests/` is in 2 places.
  - To ignore something, it only needs to be ignored in 1 place at the minimum, but will be skipped.
  - To include something, it needs to be REMOVED from an explicit semgrepignore list AND from SAP > project/global ignores

### DIFF
--- a/docs/ignoring-files-folders-code.md.template
+++ b/docs/ignoring-files-folders-code.md.template
@@ -247,7 +247,8 @@ Because Semgrep ignore logic is configured at the file, repository, and platform
 
 - It is recommended to only create an custom, user-defined `.semgrepignore` file if you are **overriding** the Semgrep defaults. This means defining all other items to ignore through the global or project path ignores.
   - This method works well if your organization primarily scans using the `semgrep ci` command.
-- Be aware that creating a user-defined `.semgrepignore` file enables developers to edit it. Don't exclude the `.semgrepignore` file from Git tracking to keep a log of changes to it.
+- Be aware that creating a user-defined `.semgrepignore` file enables developers to edit it.
+- Include the `.semgrepignore` file in Git tracking to keep a log of changes and ensure it's applied consistently.
 - To **include** a file or folder for scanning, ensure that you have checked the following places:
   - Global path ignores
   - Project path ignores
--- a/docs/ignoring-files-folders-code.md.template
+++ b/docs/ignoring-files-folders-code.md.template
@@ -245,7 +245,7 @@ This section focuses on ignoring as excluding or skipping files, not as a triage
 
 Because Semgrep ignore logic is configured at the file, repository, and platform level, you may sometimes encounter unexpected behavior.
 
-- It is recommended to only create an custom, user-defined `.semgrepignore` file if you are **overriding** the Semgrep defaults, if possible. This means defining all other items to ignore through the global or project path ignores.
+- It is recommended to only create an custom, user-defined `.semgrepignore` file if you are **overriding** the Semgrep defaults. This means defining all other items to ignore through the global or project path ignores.
   - This method works well if your organization primarily scans using the `semgrep ci` command.
 - Be aware that creating a user-defined `.semgrepignore` file enables developers to edit it. Don't exclude the `.semgrepignore` file from Git tracking to keep a log of changes to it.
 - To **include** a file or folder for scanning, ensure that you have checked the following places:
--- a/docs/ignoring-files-folders-code.md.template
+++ b/docs/ignoring-files-folders-code.md.template
@@ -249,7 +249,7 @@ Because Semgrep ignore logic is configured at the file, repository, and platform
   - This method works well if your organization primarily scans using the `semgrep ci` command.
 - Be aware that creating a user-defined `.semgrepignore` file enables developers to edit it.
 - Include the `.semgrepignore` file in Git tracking to keep a log of changes and ensure it's applied consistently.
-- To **include** a file or folder for scanning, ensure that you have checked the following places:
+- To **include** a file or folder for scanning, ensure it's not in any of the following places:
   - Global path ignores
   - Project path ignores
   - User-defined `.semgrepignore`
--- a/docs/ignoring-files-folders-code.md.template
+++ b/docs/ignoring-files-folders-code.md.template
@@ -27,8 +27,8 @@ All Semgrep environments (CLI, CI, and Semgrep AppSec Platform) adhere to user-d
 | Method  | Usage    |
 |:--------|:---------|
 | To ignore blocks of code: Add a `nosemgrep` annotation | Create a comment, followed by a space (` `), followed by `nosemgrep` at the first line or preceding line of the pattern match. This generates a finding that is automatically ignored. <br /><br />For example:<br /><br />` // nosemgrep` &nbsp; &nbsp; &nbsp;<br />`// nosemgrep: rule-id` <br /> `# nosemgrep` |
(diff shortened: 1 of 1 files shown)


khorne3 (approved):

docs/ignoring-files-folders-code.md.template
```diff
@@ -23,32 +24,39 @@ All Semgrep environments (CLI, CI, and Semgrep AppSec Platform) adhere to user-d
...
+| For **teams or organizations** deploying Semgrep as part of a security program: <br /><br /><ul><li>Ignore files and folders through the use of Semgrep AppSec Platform's **Project or Global ignores**</li><li>Override the implicit ignorelist through the use of a `.semgrepignore` file.</li></ul> | Navigate to **Projects > <PL>PROJECT_NAME</PL> > Details > Settings > Path ignores**.<br /><br /> ![Set ignore paths for a project in Semgrep AppSec Platform.](/img/per-project-ignores.png) |
+| For **individuals** scanning repositories or local codebases, or Semgrep CE users:<br /><br />Ignore files and folders through a `.semgrepignore` file | Create a `.semgrepignore` file in your **repository's root directory** or your **project's working directory** and add patterns for files and folders there. Patterns follow `.gitignore` syntax with some caveats. See [Defining ignored files and folders in `.semgrepignore`](#define-ignored-files-and-folders-in-semgrepignore). | [`.semgrepignore` sample file](https://raw.githubusercontent.com/semgrep/semgrep/develop/cli/src/semgrep/templates/.semgrepignore) |
```
relsqui, line 31: I'm a little confused by the team/individual distinction here. Anyone with access to the platform can use the platform ignores, no? (Sorry if we discussed this already, it feels familiar but I can't remember where the original conversation was.) There's also some outdated info at the "defining ignored files and folders" link but I don't know if we want to deal with that now or make it a separate change.

docs/ignoring-files-folders-code.md.template
```diff
@@ -23,32 +24,39 @@ All Semgrep environments (CLI, CI, and Semgrep AppSec Platform) adhere to user-d
...
+ 
 
 ## Understand Semgrep defaults
 
 Without user customization, Semgrep refers to the following to define ignored files and folders:
 
 * Semgrep's default `.semgrepignore` file
 * Your repository's `.gitignore` file (if it exists)
+* For Semgrep AppSec Platform users: each project (repository) in Semgrep **includes** a list of ignored files and folders by default in the **Path ignores** list
```
relsqui, line 40: "Includes" is a word we use in several places to mean "definitely for sure scan this file" so emphasizing it for an ignore list might be misleading. Maybe something like "has a list of ignored files and folders in its project details page"? That way we're also giving a hint about where to find it.

docs/ignoring-files-folders-code.md.template
```diff
@@ -23,32 +24,39 @@ All Semgrep environments (CLI, CI, and Semgrep AppSec Platform) adhere to user-d
...
+## Override defaults
+
 The default `.semgrepignore` file causes Semgrep to skip these folders:
 
 * `/tests`, `/test`
 * `/vendors`
 
-To include these folders, create a `.semgrepignore` file without those paths.
-:::
+To include these folders or override the defaults:
```
relsqui, line 55: I don't actually remember whether you need to do both of these or either, which means we should probably specify. 😭😅

docs/ignoring-files-folders-code.md.template
```diff
@@ -223,9 +235,27 @@ Semgrep AppSec Platform users can disable rules and rulesets through the Policie
...
+
+:::tip
+This section focuses on ignoring as excluding or skipping files, not as a triage action. 
+:::
+
+Because Semgrep ignore logic is configured at the file, repository, and platform level, you may sometimes encounter unexpected behavior.
+
+- It is recommended to only create an custom, user-defined `.semgrepignore` file if you are **overriding** the Semgrep defaults. This means defining all other items to ignore through the global or project path ignores.
+  - This method works well if your organization primarily scans using the `semgrep ci` command.
+- Be aware that creating a user-defined `.semgrepignore` file enables developers to edit it. Don't exclude the `.semgrepignore` file from Git tracking to keep a log of changes to it.
```
relsqui, line 250:
```suggestion
- Be aware that creating a user-defined `.semgrepignore` file enables developers to edit it.
- Include the `.semgrepignore` file in Git tracking to keep a log of changes and ensure it's applied consistently.
```
Two negatives -> positive, and also they really should do this anyway. 😅 It's necessary for CI or SMS and even on local dev machines, you'd need to be the same for everyone to avoid messing up findings. Those felt like separate enough points to be their own bullets, but I don't feel strongly about that.

docs/ignoring-files-folders-code.md.template
```diff
@@ -223,9 +235,27 @@ Semgrep AppSec Platform users can disable rules and rulesets through the Policie
...
+- To **include** a file or folder for scanning, ensure that you have checked the following places:
```
relsqui, line 251:
```suggestion
- To **include** a file or folder for scanning, ensure it's not in any of the following places:
```
Just being more specific about what the "check" action is.

relsqui (changes requested):
Big content improvement! I like the added headings. Left a couple of suggestions and some questions, can chat more next week about my confusion if that'd help. 😅

docs/ignoring-files-folders-code.md.template
s-santillan: Yeah, I think the stronger distinction is probably just CE vs AppSec Platform. I'll remove the "individuals" bit.

s-santillan:
> There's also some outdated info at the "defining ignored files and folders" link but I don't know if we want to deal with that now or make it a separate change.

Hiya @relsqui! Which part specifically? They are additive, are they not? So removing an item from one place (AppSec Platform) doesn't override (override in this case means that it doesn't override the semgrepignore default) because the other (semgrepignore) exists.

docs/ignoring-files-folders-code.md.template
s-santillan: Perhaps it's not clear what this sentence is saying. "To include these folders (i.e. override the defaults)" is what I was going for.

s-santillan: Shortened it to prevent confusion!

docs/ignoring-files-folders-code.md.template
s-santillan: Yeap, good callout!

PR #2157, opened 2025-05-21
update ADO instructions re: diff-aware scans

khorne3:
### Please ensure

### DIFF
--- a/docs/kb/semgrep-ci/azure-self-hosted-ubuntu.md
+++ b/docs/kb/semgrep-ci/azure-self-hosted-ubuntu.md
@@ -52,14 +52,14 @@ steps:
   persistCredentials: true
 - script: |
     pipx install semgrep
-    if [ $(System.PullRequest.PullRequestId) -ge 0 ]; then
+    if [ $(Build.SourceBranchName) = "master" ]; then
+        echo "Semgrep full scan"
+        semgrep ci
+    elif [ $(System.PullRequest.PullRequestId) -ge 0 ]; then
         echo "Semgrep diff scan"
         git fetch origin master:origin/master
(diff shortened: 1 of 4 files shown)


docs/kb/semgrep-ci/azure-self-hosted-ubuntu.md
```diff
@@ -52,15 +52,15 @@ steps:
   persistCredentials: true
 - script: |
     pipx install semgrep
-    if [ $(Build.SourceBranchName) = "master" ]; then
-        echo "Semgrep full scan"
-        semgrep ci
-    elif [ $(System.PullRequest.PullRequestId) -ge 0 ]; then
+    if [ $(System.PullRequest.PullRequestId) -ge 0 ]; then
```
khorne3, line 55: @armchairlinguist I inverted this section to eliminate the need for hardcoding the default branch name. Is this too strange/not done? WDYT? cc: @stuartcmehrens

armchairlinguist: What you've done isn't equivalent. It may be something we want to do, but we should discuss the approach.

This config runs full scans on the default branch (unfortunately hardcoded; I'm not sure if we can detect it) and diff scans on pull requests. Your config runs full scans on any branch unless a pull request ID is provided. This could mean that we would run a full scan on a feature branch before a PR is opened, and therefore mess with PR comments. I wouldn't recommend making the change as-is for that reason.

khorne3: Got it; thanks for the sanity check! I'm undoing all of this and will revisit at a later date after we talk about the best way to do so.

s-santillan (approved):

PR #2164, opened 2025-05-26
Adjust CI/CD code snippets for OSS

Nirusu:
When reading the CI/CD snippets in the docs regarding the OSS version, I noticed two things:
- The Gitlab CI/CD snippet refers to `semgrep ci --config auto`, which does not work like this. Changed it to `semgrep scan --config auto` similar to the other examples
- The pull/merge request comments mention that they would offer 'diff-based' scanning. However, diff-based scanning would require `semgrep ci` (which is a non-OSS feature), so with `semgrep scan` they would scan the full repository. Hence, I remove the mentions of diff-based scanning from the OSS snippets.

### Please ensure


s-santillan:
@Nirusu Thank you for the catch!

PR #2165, opened 2025-05-27
Updated footer 

s-santillan:

### Please ensure

Updated links to match about 80% the content in the main site (where appropriate).
Match the colors used in the site.


s-santillan:
Note that the color discrepancy has been fixed. However, the "fix" was #fff, which is not an official color in the brand guidelines (or it would have been defined as a variable in the `custom.scss`).

PR #2210, opened 2025-06-24
update pattern-not to include references to additional options

khorne3:
### Please ensure

### DIFF
--- a/docs/writing-rules/rule-syntax.md
+++ b/docs/writing-rules/rule-syntax.md
@@ -856,6 +856,54 @@ db_query("SELECT * FROM ...")
 db_query("SELECT * FROM ...", verify=True, env="prod")
 ```
 
+Additionally, `pattern-not` accepts one `pattern` or `patterns` property, which negates the entire `pattern` or `patterns` construction:
+
+```yaml
+rules:
+  - id: gcp-backend-service-missing-hsts
+    languages:
+      - hcl
+    severity: ERROR
+    message: GCP backend service missing HSTS header in custom_response_headers
+    patterns:
+      - pattern: |-
+          resource "google_compute_backend_service" $_ {
+            ...
+          }
+      - pattern-not:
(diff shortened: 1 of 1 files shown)


brandonspark (approved):
great example!

docs/writing-rules/rule-syntax.md
```diff
@@ -856,6 +856,8 @@ db_query("SELECT * FROM ...")
 db_query("SELECT * FROM ...", verify=True, env="prod")
 ```
 
+Additionally, `pattern-not` accepts one `pattern` or `patterns` property, which negates the entire `pattern` or `patterns` construction.
```
relsqui, line 859: I don't love this example but I think it's clear enough in context!

I'm also not sure how to describe the unit of allowable constructions here that includes `pattern` and `patterns` and `pattern-either` (and maybe even `pattern-not` although that would be silly). Opted to suggest removing `pattern` partly for brevity but also because there's no point in using just one pattern this way (you could just use the basic `pattern-not` syntax instead).

Swapped "additionally" for "alternatively" because I think it's one or the other, not both.

Nota bene that the fix suggestion doesn't seem to be handling the closing backticks for the interior code block correctly (might need fixing if you take this suggestion).


```suggestion
Alternatively, `pattern-not` accepts a `patterns` or `pattern-either` property, and negates everything inside it.

```yaml
rules:
  - id: unverified-db-query
    patterns:
      - pattern: db_query(...)
      - pattern-not:
          pattern-either:
            - pattern: db_query(..., verify=True, ...)
            - pattern-inside: |
                with ensure_verified(db_query):
                  db_query(...)
    message: Found unverified db query
    severity: ERROR
    languages:
      - python
```
```

relsqui (changes requested):
Left you a rule example and a couple of minor tweaks.

relsqui (approved):

PR #2216, opened 2025-06-26
update roles and scopes information

khorne3:
This PR:
- updates GitLab docs to include roles and token scopes information
- updates the Network Broker doc to link out to tokens info (except for GitHub, since most use the app)

### Please ensure

### DIFF
--- a/docs/deployment/managed-scanning/gitlab.md
+++ b/docs/deployment/managed-scanning/gitlab.md
@@ -66,7 +66,8 @@ You can enabled managed scanning for additional repositories after onboarding us
 
 ### If the page doesn't display any repositories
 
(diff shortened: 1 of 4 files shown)


armchairlinguist (approved):
This pretty much looks good to me, so I'm going to approve. 

I do have one suggestion, though: since we're calling a bit more attention to the use of tokens, I think we should also have a comment that we recommend having the token in the SCM config in the platform unless there's a specific reason that it's better to provide it via the broker. 

As far as I understand this aligns with our usual recommendations (although @zyannes can correct me if I'm wrong, as well). Totally viable to consider that in a follow-up PR too since it can be done separately.

PR #2223, opened 2025-07-02
Explicitly use `semgrep` switch in `opam env` commands

chrisnovakovic:
The `opam switch create` command in the instructions for contributing creates an opam switch named `semgrep`, but later commands don't explicitly refer to a switch when setting up the environment in order to use it, which may cause unintended behaviour if more than one switch exists. Explicitly refer to the `semgrep` switch in invocations of `opam env`.

### Please ensure

### DIFF
--- a/docs/contributing/semgrep-core-contributing.md
+++ b/docs/contributing/semgrep-core-contributing.md
@@ -30,7 +30,7 @@ git submodule update --init --recursive
 brew install opam
 opam init
 opam switch create semgrep 5.3.0
-eval $(opam env)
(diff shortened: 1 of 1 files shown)


khorne3:
Hey @chrisnovakovic! Thanks for the contribution!

We believe that adding the suggested command introduces unnecessary complexity—when you create a new switch, opam automatically switches to it. Users managing multiple switches are likely already familiar with switching between them. That said, we can add a brief note to remind users to run `opam switch semgrep` followed by `eval $(opam env)` if they've switched away from their Semgrep switch.

PR #2224, opened 2025-07-02
Update docs with AWS Bedrock support info

4tulrawat:
Updated the document to include missing information about AWS Bedrock support.

### Please ensure


khorne3:
I agree with @armchairlinguist -- Bedrock is default, and the instructions are for customizing Assistant so you're not using the defaults. If we reach the point where we support BYO key for Bedrock, we should proceed with these changes at that time.

PR #2243, opened 2025-07-16
Update customize-semgrep-precommit.md

tembalii:

### Please ensure

### DIFF
--- a/docs/kb/integrations/customize-semgrep-precommit.md
+++ b/docs/kb/integrations/customize-semgrep-precommit.md
@@ -23,11 +23,11 @@ This example is based on the `semgrep-ci` hook, but a similar adjustment would a
 ```yaml
 repos:
 - repo: https://github.com/semgrep/pre-commit
-  rev: 'v1.101.0'
+  rev: 'v1.128.0'
   hooks:
-    - id: semgrep-verbose
+    - id: semgrep-ci
       entry: semgrep
-      args: ["ci", "--dry-run", "--baseline-commit", "HEAD" "2>/dev/null"]
+      args: ["ci", "--dry-run", "--baseline-commit", "HEAD","--quiet"]
       verbose: true
       pass_filenames: false
 ```
@@ -37,7 +37,7 @@ To print only a portion of the scan output, consider using shell-based text tool
 ```yaml
 repos:
 - repo: https://github.com/semgrep/pre-commit
-  rev: 'v1.126.0'
+  rev: 'v1.128.0'
   hooks:
       - id: semgrep-scan-summary-only
         entry: bash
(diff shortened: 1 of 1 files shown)


oscarhearsawho (approved):
LGTM!

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -23,11 +23,11 @@ This example is based on the `semgrep-ci` hook, but a similar adjustment would a
...
 repos:
 - repo: https://github.com/semgrep/pre-commit
-  rev: 'v1.101.0'
+  rev: 'v1.128.0'
   hooks:
-    - id: semgrep-verbose
+    - id: semgrep-ci
       entry: semgrep
-      args: ["ci", "--dry-run", "--baseline-commit", "HEAD" "2>/dev/null"]
+      args: ["ci", "--dry-run", "--baseline-commit", "HEAD","--quiet"]
```
armchairlinguist, line 30: Both options are mentioned here, so what's the advantage of changing which one is provided?

docs/kb/integrations/customize-semgrep-precommit.md
```diff
@@ -23,11 +23,11 @@ This example is based on the `semgrep-ci` hook, but a similar adjustment would a
 ```yaml
 repos:
 - repo: https://github.com/semgrep/pre-commit
-  rev: 'v1.101.0'
+  rev: 'v1.128.0'
   hooks:
-    - id: semgrep-verbose
+    - id: semgrep-ci
```
armchairlinguist, line 28: I don't think this is correct. `semgrep-ci` is a named hook that already exists, and this isn't it.

armchairlinguist (changes requested):

docs/kb/integrations/customize-semgrep-precommit.md
tembalii: image
@armchairlinguist when you run what is currently in the doc, you get the error in the screenshot. It does not complain about that error when you change it from semgrep-verbose to either semgrep or semgrep-ci. However a different error pops up . see below

image

it only works when I change to --quiet see below 
image


or when you run the pre-commit with 2>/dev/null - see below

image


Hence why I am making the updates

armchairlinguist: There are no errors in this last run. What is the issue you're talking about? The hook "failed" because there are blocking findings.

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: Yes good catch on the comma, but I don't think there is an issue with the redirect (see previous thread). Both are behaving correctly.

docs/kb/integrations/customize-semgrep-precommit.md
armchairlinguist: The first issue you shared is similar to one that Jon brought up. I still don't entirely understand what's causing that; I don't have that issue in my test repo. Let's discuss it in the internal thread so we don't clutter this one too much.

PR #2258, opened 2025-07-30
upgrade node

khorne3:
the first step in upgrading Docusaurus is to upgrade Node


khorne3:
@pabloest yeah, this upgrade should fix those warnings; it's one of those chicken-and-egg things

PR #2300, opened 2025-09-03
update number of findings that can be grouped together on one Jira ticket

khorne3:


lostorblady (approved):
Thanks! I sneaked this one in with the other changes but that was paused and this should still be updated.

PR #2598, opened 2026-04-29
[TEC-565] AI credits and usage limits update

abhijna:

### DIFF
--- a/docs/semgrep-code/ai-detection-concepts.md
+++ b/docs/semgrep-code/ai-detection-concepts.md
@@ -49,7 +49,7 @@ During beta, you can choose between OpenAI, Anthropic, and Bedrock AI providers.
 
 ### Credits required for AI actions
 
-Review the number of AI credits required to run AI-powered detection scans, as well as your monthly AI credit allocation, on the [Usage and billing](/usage-and-billing/overview#ai-credits) page.
+See [Usage and billing](/usage-and-billing/overview#ai-credits) for information about credits required for AI actions.
 
 ### Data privacy and finding severity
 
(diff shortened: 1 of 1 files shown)


docs/semgrep-code/ai-detection-concepts.md
```diff
@@ -47,9 +47,9 @@ AI-powered detection builds on Semgrep's existing integration framework, such as
 
 During beta, you can choose between OpenAI, Anthropic, and Bedrock AI providers.
 
-### Scan limits 
+### Credits required for AI actions
 
-Each full AI-powered scan counts as one scan. Paying customers can trigger **30 scans per month**, and free tier customers can trigger **10 scans per month**. Please contact your Semgrep account manager or [Semgrep support](/docs/support) to discuss increasing your quota.  
+Review the number of AI credits required to run AI-powered detection scans, as well as your monthly AI credit allocation, on the [Usage and billing](/usage-and-billing/overview#ai-credits) page.
```
khorne3: The link notes that "AI-powered detection scans use a variable number of credits per scan. Credit use depends on scan size and complexity, so larger or more complex scans may use more credits" so does it make sense to call out here that the credits required is variable and then point to the linked page? It feels like we're sending them to a place where they still won't get an answer

abhijna: From conversations with the team, it sounds like this info is evolving quickly. I thought it would be better to link to one source that we can update as needed?

khorne3: Oh gotcha. Then maybe just like "[Usage and billing](/usage-and-billing/overview#ai-credits) for information about credits required for AI actions"? I know it's kind of minor, but it feels less like we're sending people on a wild goose chase

abhijna:
```suggestion
See [Usage and billing](/usage-and-billing/overview#ai-credits) for information about credits required for AI actions.
```

khorne3 (approved):

PR #2644, opened 2026-06-01
docs(sca): document private registry access via network broker

salolivares:
## Context

Supply Chain recently added a **Use network broker for registry access**
checkbox when connecting a private registry, letting Semgrep reach registries
that are only accessible from a customer's private network. The docs didn't
cover this option, and the existing **Registry support** section incorrectly
stated private registries weren't supported.

## Solution

CleanShot 2026-06-01 at 14 55 54

CleanShot 2026-06-01 at 14 55 50

- Correct the **Registry support** section: public npm/PyPI and private PyPI
  registries are supported, with a link to the connect flow.
- Add an optional step to the **Connect a private registry** flow documenting
  the **Use network broker for registry access** checkbox, calling out that it
  requires a deployed Network Broker and that the broker `allowlist` must
  include the registry URL.
- Add a **Use Semgrep Network Broker for private registry access** section to
  the Network Broker page with the required `allowlist` config (`GET`/`HEAD`
  on the registry host). Cross-linked both directions.

## Testing plan

- Verified internal links resolve to real anchors (`#registry-support`,
  `#connect-a-private-registry-to-semgrep`,
  `#use-semgrep-network-broker-for-private-registry-access`).
- Visual check of the nested list and tooltip rendering in local preview.

### DIFF
--- a/docs/semgrep-ci/network-broker.mdx
+++ b/docs/semgrep-ci/network-broker.mdx
@@ -376,6 +376,23 @@ inbound:
 
 In Network Broker `v0.31.0` and later, this URL is part of the default allowlist.
 
+## Use Semgrep Network Broker for private registry access
+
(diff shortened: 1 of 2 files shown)


abhijna (approved):
Lgtm! This triage and remediation doc is getting long and could probably use some reorg and trimming. We can do this in a separate PR.

PR #2674, opened 2026-06-17
docs: default Guardian for Claude Code to the hosted remote server

milanwilliams:
## Summary

- Updates the Semgrep Guardian docs so Claude Code install defaults to Semgrep's remote server. The Claude Code local plugin is reframed as a legacy fallback, and points at [`semgrep/guardian-local`](https://github.com/semgrep/guardian-local).

- **Non-Claude tabs** (Codex, Cursor, GitHub Copilot, VS Code, Windsurf, Kiro, Other IDEs) link to a new shared **Install the Semgrep CLI** section instead of repeating the steps.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/guardian.mdx
+++ b/docs/guardian.mdx
@@ -254,7 +254,7 @@ See [Kiro docs](https://kiro.dev/docs/mcp/) for more info.
 
 ## Install the Semgrep CLI
 
-Every IDE except Claude Code runs Semgrep from a locally installed CLI. (Claude Code uses Semgrep's hosted remote server by default and doesn't need a local CLI; its legacy local plugin installs the CLI for you.)
+Every IDE except Claude Code runs Semgrep from a locally installed CLI. (Claude Code uses Semgrep's hosted remote server by default and doesn't need a local CLI.)
 
 <Steps>
 <Step>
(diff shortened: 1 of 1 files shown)


docs/guardian.mdx
```diff
@@ -79,22 +43,27 @@ semgrep login && semgrep install-semgrep-pro
...
 
-    <Accordion title="Alternative: Use Semgrep’s hosted remote server (beta)">
-      <Note>Beta.</Note>
-
-      If you can't install the Semgrep CLI locally, you can use Semgrep's hosted remote server instead.
+    <Accordion title="Legacy: Run Semgrep Guardian locally">
+      The remote server is the recommended default. If you need to run Semgrep locally instead, you can install the legacy local plugin from the [`semgrep/guardian-local`](https://github.com/semgrep/guardian-local) repo.
 
       <Steps>
         <Step>
```
liukatkat, line 72: We should add a install the semgrep cli step here too since the local plugin requires semgrep.

liukatkat (approved):
I left one comment on the legacy instruction that i think we should fix but other than that it looks good!

docs/guardian.mdx
milanwilliams: Fixed!

PR #2725, opened 2026-07-20
Changes in Atlassian service accounts

armchairlinguist:
I guess at the time this line was added, "service" accounts must have included interactive login. Now the thing that Atlassian calls "service accounts" does not allow interactive login, so they can't actually be used with our Jira integration at the moment; the best option is just an account that isn't associated with a specific person.

Please ensure:

### DIFF
--- a/docs/semgrep-appsec-platform/jira.mdx
+++ b/docs/semgrep-appsec-platform/jira.mdx
@@ -8,7 +8,7 @@ description: "The Semgrep Jira integration allows you to create Jira tickets bas
 
 - You must have a **Jira Cloud** plan. Jira Data Center (self-managed or on-premise) is not supported.
 - You must have at least one Jira project to set as the default location where tickets will be created.
-- Optional, but recommended: A [service account](https://support.atlassian.com/user-management/understand-service-accounts/) to set up and configure the Jira integration.
+- Optional, but recommended: An Atlassian account that is not associated with a specific human user, to set up and configure the Jira integration.
 
 ## Features
 


komal-dhull (approved):

docs/semgrep-appsec-platform/jira.mdx
```diff
@@ -8,7 +8,7 @@ description: "The Semgrep Jira integration allows you to create Jira tickets bas
 
 - You must have a **Jira Cloud** plan. Jira Data Center (self-managed or on-premise) is not supported.
 - You must have at least one Jira project to set as the default location where tickets will be created.
-- Optional, but recommended: A [service account](https://support.atlassian.com/user-management/understand-service-accounts/) to set up and configure the Jira integration.
+- Optional, but recommended: An Atlassian account that is not associated with a specific human user, to set up and configure the Jira integration.
```
khorne3, line 11:
```suggestion
- Optional, but recommended: An Atlassian account that is not associated with a specific user, to set up and configure the Jira integration.
```

khorne3 (approved):

docs/semgrep-appsec-platform/jira.mdx
armchairlinguist: @khorne3 curious about your suggestion here. It is necessarily the case in our current setup that the account is a user account, and thus associated with a user entity. It just doesn't have to be owned by one single real person. So to me the qualifier feels necessary.

khorne3: I think specific user implies a person, so a more general account would therefore not be person-specific. I don't have strong feelings about this, though, so I'd be ok with your original change!

PR #2741, opened 2026-07-27
Region support info for Bedrock

armchairlinguist:
Add region support info for Bedrock, update model identifiers.

Please ensure:

CleanShot 2026-07-27 at 15 34 46@2x

### DIFF
--- a/docs/semgrep-multimodal/customize.mdx
+++ b/docs/semgrep-multimodal/customize.mdx
@@ -251,7 +251,7 @@ When you use your own Amazon Bedrock account, you can select one of the followin
 All Bedrock calls are pinned to the region you select. Semgrep does not use AWS cross-region inference profiles for customer-managed Bedrock accounts, so requests stay within your configured region. 
 
 <Note>
-The models Multimodal requires must be available for on-demand inference in the selected region. Because of that requirement, regions outside the US are not currently supported. In those regions, AWS offers the required Anthropic Claude models only through cross-region inference profiles rather than as direct on-demand models.
+Since Semgrep does not use AWS cross-region inference profiles, the models Multimodal requires must be available for on-demand inference in the selected region. Because of that requirement, regions outside the US (where the models are offered only through cross-region inference profiles) are not currently supported.
 </Note>
 
 ### Azure OpenAI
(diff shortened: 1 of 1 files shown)


Sly1029 (approved):

docs/semgrep-multimodal/customize.mdx
```diff
@@ -237,7 +237,22 @@ If you want to keep all data within your AWS account, you can use your own Amazo
...
+
+- US East (N. Virginia) - `us-east-1`
+- US East (Ohio) - `us-east-2`
+- US West (N. California) - `us-west-1`
+- US West (Oregon) - `us-west-2`
+
+All Bedrock calls are pinned to the region you select. Semgrep does not use AWS cross-region inference profiles for customer-managed Bedrock accounts, so requests stay within your configured region. 
+
+<Note>
+The models Multimodal requires must be available for on-demand inference in the selected region. Because of that requirement, regions outside the US are not currently supported. In those regions, AWS offers the required Anthropic Claude models only through cross-region inference profiles rather than as direct on-demand models.
```
khorne3: I struggled a bit to parse this, so what do you think of something like this?

> In non-US regions, AWS offers the required Anthropic Claude models through cross-region inference profiles, not direct on-demand models. This, combined with the fact that the models required by Multimodal must be available for on-demand inference in the selected region, means non-US regions aren't currently supported.

khorne3 (approved):

docs/semgrep-multimodal/customize.mdx
armchairlinguist: I don't think that actually has quite as much information - it removes the detail about requests staying within the configured region. Can you say more about what you were having trouble parsing? This is a Claude-generated text originally, although I edited it down, so I won't be offended by criticism. 😆

khorne3: Oops, sorry for being unclear -- the suggestion I made was only for the `<note>`; the text immediately above the note is fine. Does that change things a bit?

armchairlinguist: Oh okay, sorry, misunderstood that. What about:
> Since Semgrep does not use AWS cross-region inference profiles, the models Multimodal requires must be available for on-demand inference in the selected region. Because of that requirement, regions outside the US (where the models are offered only through cross-region inference profiles) are not currently supported.

khorne3: LGTM!

armchairlinguist:
```suggestion
Since Semgrep does not use AWS cross-region inference profiles, the models Multimodal requires must be available for on-demand inference in the selected region. Because of that requirement, regions outside the US (where the models are offered only through cross-region inference profiles) are not currently supported.
```

armchairlinguist: Applied the edit!

PR #2745, opened 2026-07-28
Broker: tip to remind token users to always include the token

armchairlinguist:
Clarify that if the broker config specifies a token for a given SCM base URL, the right token has to appear in all inbound/allowlist entries or errors can occur.

Covers token use, but just a troubleshooting tip, so no direct security implications.

Please ensure:

### DIFF
--- a/docs/semgrep-ci/network-broker.mdx
+++ b/docs/semgrep-ci/network-broker.mdx
@@ -407,7 +407,7 @@ inbound:
     baseURL: https://GITHUB_BASE_URL/api/v3
 ```
 
-Subsequent entries for the same SCM require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
+Subsequent entries for the same SCM type (e.g. a second GitHub Enterprise Server instance) require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
 
 ```yaml
 inbound:
@@ -430,3 +430,5 @@ inbound:
         Authorization: "Bearer GITHUB_PAT"
     ...
 ```
+
+If you are setting the SCM token in the Network Broker config, rather than in Semgrep AppSec Platform, the token must be specified in all entries referring to the same base URL. Missing token entries may result in authentication errors.
--- a/docs/semgrep-ci/network-broker.mdx
+++ b/docs/semgrep-ci/network-broker.mdx
@@ -407,7 +407,7 @@ inbound:
     baseURL: https://GITHUB_BASE_URL/api/v3
 ```
 
-Subsequent entries for the same SCM type (e.g. a second GitHub Enterprise Server instance) require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
+Subsequent entries for the same SCM type, such as a second GitHub Enterprise Server instance, require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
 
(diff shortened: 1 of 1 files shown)


docs/semgrep-ci/network-broker.mdx
```diff
@@ -407,7 +407,7 @@ inbound:
     baseURL: https://GITHUB_BASE_URL/api/v3
 ```
 
-Subsequent entries for the same SCM require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
+Subsequent entries for the same SCM type (e.g. a second GitHub Enterprise Server instance) require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
```
khorne3:
```suggestion
Subsequent entries for the same SCM type, such as a second GitHub Enterprise Server instance, require you to modify `allowlist` and add specific information needed for the HTTP requests. The following is a sample allowlist for additional GitHub entries:
```

docs/semgrep-ci/network-broker.mdx
```diff
@@ -430,3 +430,5 @@ inbound:
         Authorization: "Bearer GITHUB_PAT"
     ...
 ```
+
+If you are setting the SCM token in the Network Broker config, rather than in Semgrep AppSec Platform, the token must be specified in all entries referring to the same base URL. Missing token entries may result in authentication errors.
```
khorne3:
```suggestion
If you are setting the SCM token in the Network Broker config, rather than in Semgrep AppSec Platform, the token must be specified using the same base URL in all entries. Missing token entries might result in authentication errors.
```

khorne3 (approved):

docs/semgrep-ci/network-broker.mdx
armchairlinguist: Hm, this changes the meaning in a way I'm not intending to do - it's suggesting that the token has to always be associated with the same base URL, but the focus here is that the base URL always needs to be associated with a valid token - specifically, that you can't omit the token in some entries.

Do you have thoughts about better ways to say that?

khorne3: Maybe something like this?

> If you set the SCM token in the Network Broker config, all entries must include a valid token for the base URL specified in those entires

I don't think I fully understand, so I apologize if my suggestions aren't quite on point!

armchairlinguist: So the broader context here is it's not too hard to end up with a lot of allowlist entries - multiple ones for the same base URL - if you're using this strategy. If you are specifying a token here, every allowlist entry has to have the token specified, and it's easy to miss one. So this is a reminder to check each entry for the token - it's not going to carry over from other entries.

khorne3: Does a more general statement like, "If you set the SCM token in the Network Broker config, ensure that each entry is complete and includes the required URL and token. Missing or incomplete entries might result in authentication errors" work?

armchairlinguist: What would you think about  turning it around, actually? Something like:

> If you are seeing authorization issues but are confident the token provided in your Network Broker configuration has the correct permissions, review all entries in the allowlist to be sure that they specify the correct token.

khorne3: That works!

armchairlinguist:
```suggestion
If you are seeing authorization issues but are confident the token provided in your Network Broker configuration has the correct permissions, review all entries in the allowlist to be sure that they specify the correct token.
```

PR #2755, opened 2026-07-29
Distinguish movement to provisionally ignored from TP/FP assessment

armchairlinguist:
Autotriage is the process of Multimodal assessing a finding as a true or false positive and applies to both rule-based and AID findings (confirmed with AI team, also how it's defined in the linked docs). 

Moving a finding to the provisionally ignored state is a separate action currently only implemented for rule-based findings, so we should keep that separation clear.

Please ensure:

- [ ] Check the **Mintlify** bot preview link on this PR (requires PR to `main`)


abhijna (approved):
Improved the definition of Provisionally ignored and called out the difference between code vs AI-powered detection. Also updated all instances of auto-triage to autotriage throughout the docs to match the UI.

jaksiks (approved):

PR #2768, opened 2026-08-05
Add note about restricted triage by comment in public repos

armchairlinguist:
You can't triage by comment in public repos on most SCMs; somehow we lost this info in a prior update.

Please ensure:

- [ ] Check the **Mintlify** bot preview link on this PR (requires PR to `main`)

This change does touch on security being a reason for the restriction, but it doesn't change security behavior at all.

### DIFF
--- a/docs/semgrep-code/triage-remediation.mdx
+++ b/docs/semgrep-code/triage-remediation.mdx
@@ -281,7 +281,7 @@ You can turn off a specific rule or ruleset to prevent Semgrep Code from using i
 You can triage your Semgrep AppSec Platform findings displayed as comments in PRs and MRs by replying with another comment.
 
 <Note>
-For security reasons, triage by comment is not permitted on public repos, except for GitLab.com, GitLab Self-Managed, and GitHub Enterprise Server.
+For security reasons, you can't triage through PR or MR comments in public repositories. This restriction does not apply to GitLab.com, GitLab Self-Managed, and GitHub Enterprise Server.
 </Note>
 
 Before proceeding, ensure that you have:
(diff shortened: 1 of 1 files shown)


docs/semgrep-code/triage-remediation.mdx
```diff
@@ -280,6 +280,10 @@ You can turn off a specific rule or ruleset to prevent Semgrep Code from using i
 
 You can triage your Semgrep AppSec Platform findings displayed as comments in PRs and MRs by replying with another comment.
 
+<Note>
+For security reasons, triage by comment is not permitted on public repos, except for GitLab.com, GitLab Self-Managed, and GitHub Enterprise Server.
```
abhijna:
```suggestion
For security reasons, you can't triage findings by replying to Semgrep's pull request comments in **public repositories**. This restriction does not apply to GitLab.com, GitLab Self-Managed, or GitHub Enterprise Server.
```

abhijna (approved):
Tried to make it a little more explicit

docs/semgrep-code/triage-remediation.mdx
armchairlinguist: I think you're right to call out my use of by here - I forgot we settled on through - but in the same way, we should use "through" to be consistent with the section title, I think. How does this sound?


```suggestion
For security reasons, you can't triage through PR or MR comments in public repositories. This restriction does not apply to GitLab.com, GitLab Self-Managed, and GitHub Enterprise Server.
```

zyannes (approved):

PR #2771, opened 2026-08-06
[TEC-614] Azure OpenAI endpoint format in main Customize Multimodal doc

abhijna:
[Preview](https://semgrep-ee9d73d8-abhijna-tec-614-give-azure-openai-endpoint.mintlify.site/semgrep-multimodal/customize#select-your-ai-provider)

- Added Azure OpenAI endpoint format in main Customize Multimodal page
- Converted the **Select your AI provider** section into tabs. 

Review:
- [ ] A technical writer reviews the PR

### DIFF
--- a/docs/semgrep-multimodal/customize.mdx
+++ b/docs/semgrep-multimodal/customize.mdx
@@ -266,9 +266,17 @@ To use Azure OpenAI with Semgrep Multimodal, you must retrieve the endpoint URL 
    1. Sign in to [Semgrep AppSec Platform](https://semgrep.dev/login) and navigate to [<Icon icon="gear" iconType="solid"/>**Settings > Global**](https://semgrep.dev/orgs/-/settings/general).
    2. Click the <Icon icon="gear" iconType="solid"/> **icon** next to **AI provider**.
    3. Select **Azure OpenAI**.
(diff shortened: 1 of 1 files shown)


armchairlinguist:
In general this looks good to me, both for the requested change and for the doc organization overall. Re-reading the doc did raise a couple of questions for me, which could potentially be addressed separately:

* What is a user supposed to do with this note?
> As of May 2025, the best model for noise filtering is gpt-5.4. The best model for other Semgrep Multimodal features is gpt-5.5. You cannot have multiple Azure OpenAI models active at a given time, but you can switch to a different one by repeating these configuration steps using the Target URI and API key for the new model.

This seems to be saying "you should pick one of those two models, but you can only pick one", but the language is complicated. Is that right? Can it be simplified?

* Do we also need to provide the pattern for xAI, or is it clear enough that people aren't making the mistake?

armchairlinguist (approved):

abhijna:
@armchairlinguist, maybe we should double-check if Semgrep still stands behind those model picks? Who's a good person to tag for this? Or, since the note is so old, we can just take it out.

armchairlinguist:
@abhijna if you want to check I would ask in #team-ai internally!

PR #2765, opened 2026-08-03
Supply Chain docs re-architecture: Malware

khorne3:
- Moved incident response guide from KB to docs and removed screenshots. Fixed nav-related things related to this move, and added redirects.
  - Made minor edits to text
- Updated and edited "Detect and remove malicious dependencies" doc
- Rearranged the sidebar nav and made titles parallel

Please ensure:

- [x] A technical writer reviews the PR
- [x] Check the **Mintlify** bot preview link on this PR (requires PR to `main`)

PR #2783, opened 2026-08-12
fix(docs.json): drop nav entry for relocated incident-response page

gaprl:
Removes one stale `docs.json` nav entry: `kb/semgrep-supply-chain/incident-response`, a page that #2765 relocated to `docs/semgrep-supply-chain/incident-response.mdx`.

Mintlify treats a nav entry with no backing file as a validation warning, and `mintlify validate` exits non-zero on warnings. That has been failing the validate step of the scheduled OpenAPI spec sync ever since — the Aug 10, 11 and 12 runs all failed, so no spec refresh PR has opened in three weekdays.

`npx mintlify validate` goes from exit 1 to exit 0 with this line removed.

The page itself was never broken: it renders at its new location, is listed in the Supply Chain nav group, and the old `/kb/` URL still redirects.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -693,7 +693,6 @@
                     "root": "kb/semgrep-supply-chain",
                     "pages": [
                       "kb/semgrep-supply-chain/exclude-rule",
-                      "kb/semgrep-supply-chain/incident-response",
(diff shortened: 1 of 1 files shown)


abhijna (approved):
Thanks for fixing this!

PR #2786, opened 2026-08-13
fix(api-reference): show beta and experimental badges

gaprl:
> Stacked on #2784 — merge that one first.

v2 marks 195 endpoints as beta or experimental and none of it rendered. This makes them show up: a sidebar label plus a callout on the endpoint page.

**Why they were invisible:** the specs use `x-badges`, a ReDoc extension Mintlify doesn't read, left over from when the API reference was served by ReDoc. Confirmed with a control — three sibling operations with identical `x-badges` render nothing, and only the one given `x-mint` shows a badge.

`scripts/mirror_openapi.py` maps each badge to `x-mint.metadata.tag` (sidebar) and `x-mint.content` (callout): `<Warning>` for experimental, `<Note>` for beta. The tag alone renders as accent text, weaker than the native `deprecated` pill beside it, and can't carry the red/blue split the `x-badges` colors encoded.

It also absorbs the inline perl `x-displayName` → `x-group` step so that finally has tests — which immediately caught a bug the perl never had: v1 emits `tags:` at indent 0 and v2 at indent 4, and the region detector read a dash at column 0 as the next top-level key, so v1 mirrored nothing.

Here's how it looks like:
image

**Verification:** 50 tests written test-first; stripped every `x-group` from both real specs and re-derived them, all 55 matching what perl produced; `mintlify validate` exit 0; badges confirmed rendering in a local preview.


abhijna (approved):
Lgtm!

PR #2796, opened 2026-08-18
feat(api-reference): auto-generated API changelog pages for v1 and v2

gaprl:
## What this does

Semgrep's API changes most weeks, and until now nothing told customers what changed. This adds a **changelog page for each API version**, generated from the OpenAPI specs' own git history — so it stays current without anyone writing it by hand.

- `/api-reference/v1/Changelog` and `/api-reference/v2/Changelog`
- One entry per day the API changed, newest first, backfilled to May 2026
- Changes grouped into Breaking / Potentially breaking / Other, each linking to its endpoint's reference page
- Subscribable by RSS, with instructions on the page

I've modified the feed such that updates to `oasdiff` or to the changelog generation script won't trigger RSS feed updates per se.

Screenshot 2026-09-03 at 11 31 06 AM

## How it works

A script walks each spec's git history and compares consecutive days with [oasdiff](https://github.com/oasdiff/oasdiff), which diffs specs *semantically* — reordered lines and description edits produce no entry, only real API changes do. The weekday cron that already syncs the specs now regenerates these pages in the same PR.

oasdiff's raw wording isn't customer-ready, so we rewrite it. Where it emits this:

```
filters/allOf[#/components/schemas/protos.projects.v1.RepoFilters]/latestScanStatus
```

we render the field a customer actually sends:

```
filters.latestScanStatus
```

## Worth a look

- **oasdiff is new in CI**, pinned to v1.28.0 with a checksum. It runs locally against our own public specs; nothing leaves the runner.
- **v1's July 14 entry is large.** That's the day docs switched to auto-generated specs, so the spec emitter changed, not the API. Easy to trim if you'd rather.

## Test plan

- [x] 132 unit tests pass; `mintlify validate` passes; every emitted endpoint link verified 200 against a local `mintlify dev`
- [ ] After merge: confirm the feed emits one item per day. `mintlify dev` serves no RSS, so this can't be checked locally.

Part of APPEX-969; implements the generator scoped in APPEX-960.


gaprl:
Closing after team discussion — we decided an auto-generated API changelog page isn't the right investment right now. The branch is kept for reference: the generator (scripts/generate_api_changelog.py), its test suite, and the verified Mintlify URL-slug rules are reusable if we revisit (see APPEX-960 / APPEX-969).

aaronmichaelacosta:
Reopening this, as we got feedback from stakeholders that the changelog + an RSS feed would be important for customers to stay up-to-date.

aaronmichaelacosta:
* **#2833** <a href="https://app.graphite.com/github/pr/semgrep/semgrep-docs/2833?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2832** <a href="https://app.graphite.com/github/pr/semgrep/semgrep-docs/2832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2796** <a href="https://app.graphite.com/github/pr/semgrep/semgrep-docs/2796?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/semgrep/semgrep-docs/2796?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

docs/api-reference/v2/Changelog.mdx
gaprl: nit: can we rename this to `changelog.mdx` and the same for v1? the title for the page comes from the frontmatter title property. a change to other files (such as our .py scripts) to use the lowercase version would be needed too.

gaprl (commented):
@aaronmichaelacosta do you think we could surface the stability badge on each changelog row? so the customer knows a "removed" was for an experimental endpoint, for example

gaprl:
I'd check the live RSS feed in the mintlify preview to verify if it's rendering all the changelog entries correctly.

aaronmichaelacosta (approved):
Approving on behalf of @gaprl (since he cannot approve a PR he opened, and I took it over).

abhijna (approved):
I reviewed the preview and it looks good. Ty!

PR #2803, opened 2026-08-21
ci: request tech-writers and appex review on OpenAPI sync PRs

gaprl:
The scheduled OpenAPI spec sync was opening PRs with nobody requested for review — there was no `reviewers` or `team-reviewers` config in the workflow at all. This requests `tech-writers` and `appex`. (There is no `docs` team in the org; `tech-writers` is the equivalent.)

Requesting *team* reviewers needs a token with organization `members: read`, which the default `GITHUB_TOKEN` cannot have — the workflow `permissions:` block has no org-level key at all. So this also switches the PR step to the `semgrep-docs-release` App and drops the `DOCS_BOT_TOKEN` fallback, which was never created as a secret, so that expression was silently resolving to `GITHUB_TOKEN`. IT has since granted the App `members: read` (confirmed via `gh api apps/semgrep-docs-release`). Bonus: App-authored PRs trigger Docs CI, so the close/reopen workaround noted in the old comment goes away.

Note for whoever merges: `tech-writers` has only read+triage on this repo, so it can approve these PRs but not merge them.

### DIFF
--- a/.github/workflows/update-openapi-specs.yml
+++ b/.github/workflows/update-openapi-specs.yml
@@ -24,7 +24,21 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 10
     steps:
+      # Runs as the semgrep-docs-release App rather than GITHUB_TOKEN for two
+      # reasons: App-authored PRs trigger Docs CI (GITHUB_TOKEN-authored ones do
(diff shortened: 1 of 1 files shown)


abhijna (approved):
@gaprl I see that I can now approve but not merge the automatically generated API docs update. Who would be incharge of merging and how are they notified?

PR #2831, opened 2026-09-03
Refine Semgrep proxy information

armchairlinguist:
I'm making some changes to proxy handling, and need to reflect that in the docs. 

The section title change also makes it clearer that a proxy is used/useful for any outside calls, not just the registry. The only link to this section is in past release notes (Oct 2022) so I haven't updated it but can.

Please ensure:

- [ ] A technical writer reviews the PR

### DIFF
--- a/docs/cli-reference.mdx
+++ b/docs/cli-reference.mdx
@@ -74,7 +74,7 @@ export HTTP_PROXY="http://10.10.1.10:3128"
 export HTTPS_PROXY="http://10.10.1.10:1080"
 ```
 
-If the proxy variables are set but empty, Semgrep will treat them as if they were not set, and will make direct HTTP requests. Semgrep will attempt to add the protocol to an otherwise-valid proxy without a protocol.
+If the proxy environment variables are set but empty, Semgrep treats them as unset and makes direct HTTP requests. If a valid proxy is missing a protocol, Semgrep attempts to add one.
 
 ## Exit codes
 
--- a/docs/cli-reference.mdx
+++ b/docs/cli-reference.mdx
@@ -56,9 +56,9 @@ The Semgrep command line tool supports a `.semgrepignore` file that follows `.gi
 
 In addition to `.semgrepignore` there are several methods to set up ignore patterns. See [Ignoring files, folders, or code](/ignoring-files-folders-code).
 
-## Connect to Semgrep Registry through a proxy
+## Connect to Semgrep Registry and other endpoints through a proxy
 
-Semgrep uses the Python3 `requests` library. Set the following environment variables to point to your proxy:
+Semgrep uses the Python3 `requests` library and OCaml's `cohttp`. Both support proxy settings using the following standard environment variables:
 
 ```bash
 export HTTP_PROXY="HTTP_PROXY_URL"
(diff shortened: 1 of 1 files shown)


docs/cli-reference.mdx
```diff
@@ -74,6 +74,8 @@ export HTTP_PROXY="http://10.10.1.10:3128"
 export HTTPS_PROXY="http://10.10.1.10:1080"
 ```
 
+If the proxy variables are set but empty, Semgrep will treat them as if they were not set, and will make direct HTTP requests. Semgrep will attempt to add the protocol to an otherwise-valid proxy without a protocol.
```
abhijna:
```suggestion
If the proxy environment variables are set but empty, Semgrep treats them as unset and makes direct HTTP requests. If a valid proxy is missing a protocol, Semgrep attempts to add one.
```

abhijna: What does "attempt" mean here? What would prevent it from adding a protocol?

docs/cli-reference.mdx
```diff
@@ -56,9 +56,9 @@ The Semgrep command line tool supports a `.semgrepignore` file that follows `.gi
 
 In addition to `.semgrepignore` there are several methods to set up ignore patterns. See [Ignoring files, folders, or code](/ignoring-files-folders-code).
 
-## Connect to Semgrep Registry through a proxy
+## Connect to Semgrep Registry and other endpoints through a proxy
```
abhijna, line 59:
```suggestion
## Connect to Semgrep through a proxy
```

abhijna: Wondering if perhaps dropping "Registry and other endpoints" achieves the same goal of making it clear that the proxy applies to all outbound requests.

armchairlinguist: I'm fine either way - the only reason I didn't delete it is that the registry is definitely the most common thing people see in terms of outbound requests.

docs/cli-reference.mdx
armchairlinguist: The protocol add only adds HTTP, so if for some reason someone had a proxy that didn't work with that scheme, it wouldn't fix the problem. That said, it would definitely add it, so I can be less vague. :) I would stick with otherwise valid though; a proxy without a scheme isn't fully valid in this context.

armchairlinguist: I'll just apply your suggestion and then restore that qualifier

docs/cli-reference.mdx
```diff
@@ -74,6 +74,8 @@ export HTTP_PROXY="http://10.10.1.10:3128"
 export HTTPS_PROXY="http://10.10.1.10:1080"
 ```
 
+If the proxy environment variables are set but empty, Semgrep treats them as unset and makes direct HTTP requests. If a valid proxy is missing a protocol, Semgrep attempts to add one.
```
armchairlinguist, line 77:
```suggestion
If the proxy environment variables are set but empty, Semgrep treats them as unset and makes direct HTTP requests. If an otherwise-valid proxy value is missing a protocol, Semgrep adds an appropriate protocol.
```
Waiting to accept this suggestion bc scheme vs protocol is potentially confusing, taking an informal poll :)

docs/cli-reference.mdx
abhijna: Sounds good. You can also edit my suggestion directly (I think)

